PR #22, opened 2021-12-13
Increase text width of comments

Caroline-Holmstrom:
Change-Id: Ic750d8e62e2a5e0eaae19a08d40f911bbf3e646c


bjornvolcker (approved):
LGTM, but make sure to remove Change-Id from commit message. Also, co-author address will probably show up wrong as well.

PR #27, opened 2021-12-14
Adds set_hash_list_size(...)

bjornvolcker:
This internal function is used to set the size of the |hash_list|.
This comes in handy for check tests to test what happens when the
limit is hit.

Further, by setting the size of the |hash_list| in the tests, the
test-settings meson option is removed since there is no longer a need
to pass in MAX_GOP_LENGTH as define at compile time.

The test_checks.sh script and READMEs have been updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,9 +78,9 @@ meson install -C build
 ```
 
 ## Configure, build and run unittests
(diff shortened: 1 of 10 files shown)


Caroline-Holmstrom (approved):
LGTM

lusikamalo (approved):
LGTM.

PR #31, opened 2021-12-22
Detect change of public key

Caroline-Holmstrom:
A video can only be signed by one Private key, hence only validated by
one Public key. If a second Public key is detected in the stream or set
by the user, any further validation should be indicated as invalid.

### DIFF
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -741,6 +741,7 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
   latest->number_of_expected_picture_nalus = -1;
   latest->number_of_received_picture_nalus = -1;
   latest->number_of_pending_picture_nalus = -1;
+  latest->public_key_has_changed = false;
 
   svi_rc status = SVI_UNKNOWN;
   SVI_TRY()
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -567,6 +567,7 @@ validate_authenticity(signed_video_t *self)
     num_expected_nalus = -1;
     num_received_nalus = -1;
   }
+  if (latest->public_key_has_changed) valid = SV_AUTH_RESULT_NOT_OK;
 
   // Update |latest_validation| with the validation result.
   latest->authenticity = valid;
--- a/lib/src/signed_video_tlv.c
+++ b/lib/src/signed_video_tlv.c
@@ -542,7 +542,7 @@ decode_public_key(signed_video_t *self, const uint8_t *data, size_t data_size)
     SVI_THROW(sv_rc_to_svi_rc(openssl_key_memory_allocated(
         &signature_info->public_key, &signature_info->public_key_size, pubkey_size)));
 
-    if(memcmp(data_ptr, signature_info->public_key, pubkey_size) != 0 && self->gop_state.has_public_key) {
+    if (memcmp(data_ptr, signature_info->public_key, pubkey_size) && self->gop_state.has_public_key) {
       self->latest_validation->public_key_has_changed = true;
     }
     memcpy(signature_info->public_key, data_ptr, pubkey_size);
--- a/lib/src/signed_video_tlv.c
+++ b/lib/src/signed_video_tlv.c
@@ -542,6 +542,9 @@ decode_public_key(signed_video_t *self, const uint8_t *data, size_t data_size)
     SVI_THROW(sv_rc_to_svi_rc(openssl_key_memory_allocated(
         &signature_info->public_key, &signature_info->public_key_size, pubkey_size)));
 
+    if(memcmp(data_ptr, signature_info->public_key, pubkey_size) != 0 && self->gop_state.has_public_key) {
+      self->latest_validation->public_key_has_changed = true;
+    }
     memcpy(signature_info->public_key, data_ptr, pubkey_size);
     self->gop_state.has_public_key = true;
     data_ptr += pubkey_size;
--- a/tests/check/check_h26xsigned_auth.c
+++ b/tests/check/check_h26xsigned_auth.c
@@ -24,18 +24,19 @@
 #include <string.h>  // strcmp
 
 #include "lib/src/includes/signed_video_auth.h"  // signed_video_authenticity_t
-#include "lib/src/includes/signed_video_sign.h"  // signed_video_set_authenticity_level()
 #include "lib/src/includes/signed_video_common.h"  // signed_video_t
-
+#include "lib/src/includes/signed_video_sign.h"  // signed_video_set_authenticity_level()
 #include "nalu_list.h"  // nalu_list_create()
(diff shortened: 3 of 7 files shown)


lib/src/signed_video_tlv.c
```diff
@@ -542,6 +542,9 @@ decode_public_key(signed_video_t *self, const uint8_t *data, size_t data_size)
     SVI_THROW(sv_rc_to_svi_rc(openssl_key_memory_allocated(
         &signature_info->public_key, &signature_info->public_key_size, pubkey_size)));
 
+    if(memcmp(data_ptr, signature_info->public_key, pubkey_size) != 0 && self->gop_state.has_public_key) {
```
bjornvolcker, line 545: white space between 'if' and '('.
You can also go with the shorter
"!memcmp(...)" instead of "memcmp(...) != 0"

lib/src/signed_video_h26x_auth.c
```diff
@@ -741,6 +741,7 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
   latest->number_of_expected_picture_nalus = -1;
   latest->number_of_received_picture_nalus = -1;
   latest->number_of_pending_picture_nalus = -1;
+  latest->public_key_has_changed = false;
```
bjornvolcker, line 744: If |public_key_has_changed| switch to true when decoding the SEI validation should fail. I suggest adding a check on line 570. Something like this:

if (latest->public_key_has_changed) valid = SV_AUTH_RESULT_NOT_OK;

tests/check/check_h26xsigned_auth.c
```diff
@@ -859,7 +863,36 @@ START_TEST(camera_reset_on_signing_side)
   // will be marked as invalid and compute 3 more NALUs than expected. In G it is communicated
   // there is only 2 NALUs present (GI). So missed NALUs equals -3 (IPP).
   const struct validation_stats expected = {
-      .valid_gops = 2, .invalid_gops = 2, .missed_nalus = -3, .pending_nalus = 4
+      .valid_gops = 2, .invalid_gops = 2, .missed_nalus = -3, .pending_nalus = 4, .public_key_has_changed = true
```
bjornvolcker, line 866: Add a TODO above this declaration that public_key_has_changed is expected to be true now when we have changed the behavior in generate private key.

tests/check/check_h26xsigned_auth.c
```diff
@@ -859,7 +863,36 @@ START_TEST(camera_reset_on_signing_side)
...
+START_TEST(detect_change_of_public_key)
+{
+  // This test runs in a loop with loop index _i, corresponding to struct sv_setting _i in
+  // |settings|; See signed_video_helpers.h.
+
+  // Generate 2 GOPs
+  nalu_list_t *list = create_signed_nalus("IPPIPP", settings[_i]);
+  nalu_list_check_str(list, "GIPPGIPP");
+
+  // Generate another GOP from scratch
```
bjornvolcker, line 884: Mention that this will get a new Public key

tests/check/check_h26xsigned_auth.c
```diff
@@ -859,7 +863,36 @@ START_TEST(camera_reset_on_signing_side)
...
+  nalu_list_t *second_list = create_signed_nalus("IPPPI", settings[_i]);
```
bjornvolcker, line 885: Suggestion: second_list -> list_with_new_public_key

tests/check/check_h26xsigned_auth.c
```diff
@@ -859,7 +863,36 @@ START_TEST(camera_reset_on_signing_side)
...
+  nalu_list_check_str(second_list, "GIPPPGI");
+
+  nalu_list_append_and_free(list, second_list);
+  nalu_list_check_str(list, "GIPPGIPPGIPPPGI");
+
+  // One pending NALU per GOP. Note that the mid GOP (IPPGI) includes the reset on the camera. It
```
bjornvolcker, line 891: Update comments. "Reset on camera" is not of interest here.

tests/check/check_h26xsigned_auth.c
```diff
@@ -1245,6 +1278,7 @@ signed_video_suite(void)
   tcase_add_loop_test(tc, no_signature, s, e);
   tcase_add_loop_test(tc, multislice_no_signature, s, e);
   tcase_add_loop_test(tc, recurrence, s, e);
+  tcase_add_loop_test(tc, detect_change_of_public_key, s, e);
```
bjornvolcker, line 1281: Put this in the same order as it appears in the code above, that is, right after camera_reset_on_signing_side

bjornvolcker:
I cannot find the change in generate private key, that overwrites existing keys. The check tests fails otherwise.

tests/check/check_h26xsigned_auth.c
```diff
@@ -882,18 +872,20 @@ START_TEST(detect_change_of_public_key)
...
   nalu_list_check_str(list, "GIPPGIPPGIPPPGI");
 
-  // One pending NALU per GOP. Note that the mid GOP (IPPGI) includes the reset on the camera. It
-  // will be marked as invalid and compute 3 more NALUs than expected. In G it is communicated
-  // there is only 2 NALUs present (GI). So missed NALUs equals -3 (IPP).
-  const struct validation_stats expected = {
-      .valid_gops = 2, .invalid_gops = 2, .missed_nalus = -3, .pending_nalus = 4, .public_key_has_changed = true
-  };
+  // One pending NALU per GOP. It will be marked as invalid and compute 3 more NALUs than expected.
+  // In G it is communicated there is only 2 NALUs present (GI). So missed NALUs equals -3 (IPP).
```
bjornvolcker, line 883: May I suggest something like this
```
// The list will be validated successfully up to the third SEI (G) which has the new Public key.
//
//   GI      -> .P (valid, 1 pending, public_key_has_changed = false)
//   IPPGI   -> ....P (valid, 1 pending, public_key_has_changed = false)
//   IPPG*I  -> NNN.P (invalid, 1 pending, public_key_has_changed = true, -3 missing)
//   IPPPG*I -> N....P (invalid, 1 pending, public_key_has_changed = false)
// where G* has the new Public key. Note that we get -3 missing since we receive 3 more than what is expected according to G*.
```

tests/check/check_h26xsigned_auth.c
```diff
@@ -879,8 +879,14 @@ START_TEST(detect_change_of_public_key)
   nalu_list_append_and_free(list, list_with_new_public_key);
   nalu_list_check_str(list, "GIPPGIPPGIPPPGI");
 
-  // One pending NALU per GOP. It will be marked as invalid and compute 3 more NALUs than expected.
-  // In G it is communicated there is only 2 NALUs present (GI). So missed NALUs equals -3 (IPP).
+  // The list will be validated successfully up to the third SEI (G) which has the new Public key.
+  //
+  //   GI      -> .P (valid, 1 pending, public_key_has_changed = false)
```
bjornvolcker, line 884: Align the (

PR #32, opened 2021-12-23
Late public key arrival handling

lusikamalo:


lib/src/signed_video_h26x_auth.c
```diff
@@ -706,6 +706,11 @@ prepare_for_validation(signed_video_t *self)
       // TODO: Is it possible to avoid a memcpy by using a pointer strategy?
       memcpy(signature_info->hash, self->gop_info->gop_hash, HASH_DIGEST_SIZE);
     }
+
+    if (signature_info->public_key) {
+      gop_state->has_public_key = true;
```
bjornvolcker, line 711: This is already (and should be) set when decoding the SEI.

tests/check/check_h26xsigned_auth.c
```diff
@@ -1200,6 +1200,107 @@ START_TEST(recurrence)
...
+  nalu_list_item_t *i_1 = nalu_list_remove_item(list, 1);
+  nalu_list_item_check_str(i_1, "I");
+  nalu_list_item_t *p_1 = nalu_list_remove_item(list, 1);
+  nalu_list_item_check_str(p_1, "P");
+  nalu_list_item_t *p_2 = nalu_list_remove_item(list, 1);
+  nalu_list_item_check_str(p_2, "P");
+  nalu_list_check_str(list, "GIPPGIPPGIPPGIPPGIPPGI");
+  /* First public key now exist in item 9 */
+
+  // GIPPGIPPGI     valid & 1 pending & 2 unsigned
```
bjornvolcker, line 1233: Why do we get 2 unsigned GOPs?

lib/src/includes/signed_video_auth.h
```diff
@@ -45,6 +45,8 @@ typedef enum {
   // validation.
   SV_AUTH_RESULT_OK = 4,
   // Successfully verified all NALUs that could be verified, and all expected NALUs are present.
+  SV_AUTH_RESULT_NO_PUBLIC_KEY = 5,
```
bjornvolcker, line 48: I do not think this is needed. Better to use SV_AUTH_RESULT_SIGNATURE_PRESENT. See comment further down.

tests/check/check_h26xsigned_sign.c
```diff
@@ -411,11 +413,11 @@ START_TEST(fallback_to_gop_level)
     nalu_list_item_check_str(sei_2, "G");
     nalu_list_item_t *sei_1 = nalu_list_remove_item(list, 1);
     nalu_list_item_check_str(sei_1, "G");
-
     // Verify SEI sizes.
-    ck_assert_uint_lt(sei_1->data_size, sei_2->data_size);
-    ck_assert_uint_lt(sei_3->data_size, sei_1->data_size);
-
+    if (settings[_i].recurrence == SV_RECURRENCE_DEFAULT) {
```
bjornvolcker, line 417: How would you change these so it applies when recurrence is present?

tests/check/signed_video_helpers.h
```diff
@@ -32,13 +32,16 @@
 #define MANUFACT "manufacturer"
 #define ADDR "address"
 
+typedef enum { SV_RECURRENCE_DEFAULT = 1, SV_RECURRENCE_THREE = 3} SignedVideoRecurrence;
```
bjornvolcker, line 35: Maybe we should have three variants; SV_NO_RECURRENCE, SV_RECURRENCE_THREE, SV_RECURRENCE_THREE_WITH_OFFSET

lib/src/signed_video_h26x_auth.c
```diff
@@ -736,6 +737,50 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
+      if (item->nalu && item->nalu->is_gop_sei && item->validation_status == 'P') {
+        const uint8_t *tlv_data = item->nalu->tlv_data;
+        size_t tlv_size = item->nalu->tlv_size;
+        // TODO: tlv_find_tag -> tlv_find_tag_and_decode?
+        const uint8_t *public_key_tag_ptr =
+            tlv_find_tag(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false);
+
+        if (public_key_tag_ptr) {
+          size_t length = ((size_t)public_key_tag_ptr[1] << 8 | public_key_tag_ptr[2]);
+          decode_public_key(self, public_key_tag_ptr + 3, length);
```
bjornvolcker, line 767: I think it is better to have an API in tlv; tlv_find_tag_and_decode().

lib/src/signed_video_tlv.c
```diff
@@ -900,6 +902,40 @@ tlv_find_tag(signed_video_t *self,
...
+
+  if (!self || !tlv_data || tlv_data_size == 0) return SVI_INVALID_PARAMETER;
+
+  const uint8_t *tag_ptr = tlv_find_tag(self, tlv_data, tlv_data_size, tag, with_ep);
+
+  if (!tag_ptr) {
+    return SVI_INVALID_PARAMETER;
+  }
+
+  // Read the length
```
bjornvolcker, line 922: Suggestion: Instead of reading the tag length and then use tlv_decode() with that length (to tell it we only want to read the first tag), I think it is more clear to simply decode_tlv_header(), get_decoder() and then decode()

Something like this, with proper error handling?
```
size_t tlv_header_size = 0;
size_t length = 0;
status = decode_tlv_header(tlv_data, &tlv_header_size, &tag, &length);
sv_tlv_decoder_t decoder = get_decoder(tag);
status = decoder(self, tlv_data + tlv_header_size, length);
```

lib/src/signed_video_tlv.h
```diff
@@ -150,11 +150,25 @@ uint8_t
 read_byte(uint16_t *last_two_bytes, const uint8_t **payload, bool do_emulation_prevention);
 
 /**
- * @brief Decodes the PUBLIC_KEY_TAG from data
+ * @brief Scans the TLV part of a SEI payload and decodes the found tag when it's detected.
```
bjornvolcker, line 153: nit: it's -> it is

lib/src/signed_video_tlv.h
```diff
@@ -150,11 +150,25 @@ uint8_t
...
  *
- * @returns SVI_OK if decoding was successful, otherwise an error code.
+ * The data is assumed to have been written in a TLV format. tlv_find_tag parses data as long as
```
bjornvolcker, line 155: Shouldn't we say what tlv_find_tag_and_decode() does instead of what tlv_find_tag() does not?

lib/src/signed_video_h26x_auth.c
```diff
@@ -758,25 +759,19 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
-        if (public_key_tag_ptr) {
-          size_t length = ((size_t)public_key_tag_ptr[1] << 8 | public_key_tag_ptr[2]);
-          decode_public_key(self, public_key_tag_ptr + 3, length);
-
+        status = tlv_find_tag_and_decode(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false);
+        if (status == SVI_OK) {
           public_key_found = true;
         } else {
-          DEBUG_LOG("Public key missing");
+          DEBUG_LOG("Public key missing (error %d)", status);
```
bjornvolcker, line 766: Hmm, is it really an error if the Public key does not exists?

lib/src/signed_video_h26x_auth.c
```diff
@@ -758,25 +759,19 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
         size_t tlv_size = item->nalu->tlv_size;
-        // TODO: tlv_find_tag -> tlv_find_tag_and_decode?
-        const uint8_t *public_key_tag_ptr =
-            tlv_find_tag(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false);
-
-        if (public_key_tag_ptr) {
-          size_t length = ((size_t)public_key_tag_ptr[1] << 8 | public_key_tag_ptr[2]);
-          decode_public_key(self, public_key_tag_ptr + 3, length);
-
+        status = tlv_find_tag_and_decode(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false);
```
bjornvolcker, line 762: There is a minor security/bug risk here when we combine both `status` and `public_key_found`. We should aim at using `status` for the SVI_TRY() scope below. Here we can put the call in the if explicitly.

lib/src/signed_video_tlv.c
```diff
@@ -909,30 +910,31 @@ tlv_find_tag_and_decode(signed_video_t *self,
...
-  status = tlv_decode(self, tag_ptr, length);
+  sv_tlv_decoder_t decoder = get_decoder(tag);
+  status = decoder(self, tag_ptr + tlv_header_size, length);
   if (status != SVI_OK) {
-    DEBUG_LOG("TLV decode failed (error %d)", status);
-    status = SVI_DECODING_ERROR;
+    DEBUG_LOG("Could not decode data (error %d)", status);
   }
 
+catch_error:
```
bjornvolcker, line 937: Consider using SVI_TRY/CATCH instead.

lib/src/signed_video_tlv.h
```diff
@@ -150,10 +150,10 @@ uint8_t
...
 
 /**
- * @brief Scans the TLV part of a SEI payload and decodes the found tag when it's detected.
+ * @brief Scans the TLV part of a SEI payload and decodes the found tag when it is detected.
  *
- * The data is assumed to have been written in a TLV format. tlv_find_tag parses data as long as
- * there are new tags, but never decodes it. The function can handle data both with and without
+ * The data is assumed to have been written in a TLV format. tlv_find_tag_and_decode parses data
+ * and finds correct tag and decodes it. The function can handle data both with and without
  * emulation prevention bytes.
```
bjornvolcker, line 157: You should probably tell that SVI_INVALID_PARAMETER is returned if the tag is not found.

lib/src/signed_video_h26x_auth.c
```diff
@@ -759,11 +758,13 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
-        status = tlv_find_tag_and_decode(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false);
-        if (status == SVI_OK) {
+        svi_rc ret = SVI_UNKNOWN;
+        ret = tlv_find_tag_and_decode(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false);
+        if (ret == SVI_OK) {
           public_key_found = true;
+          break;
         } else {
-          DEBUG_LOG("Public key missing (error %d)", status);
+          DEBUG_LOG("Public key missing");
```
bjornvolcker, line 767: I think this should be under the if at line 772 instead. Which then can help simplify the code above. For example
```
public_key_found = tlv_find_tag_and_decode(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false) == SVI_OK;
```
med ändring i `while()`
```
while (item && !public_key_found) {
```

lib/src/signed_video_h26x_auth.c
```diff
@@ -759,11 +758,13 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
       if (item->nalu && item->nalu->is_gop_sei && item->validation_status == 'P') {
         const uint8_t *tlv_data = item->nalu->tlv_data;
         size_t tlv_size = item->nalu->tlv_size;
-        status = tlv_find_tag_and_decode(self, tlv_data, tlv_size, PUBLIC_KEY_TAG, false);
-        if (status == SVI_OK) {
+        svi_rc ret = SVI_UNKNOWN;
```
bjornvolcker, line 761: The name `ret` indicates that it is a variable that is returned. I suggest removing `ret` entirely; See comment below.

lib/src/signed_video_tlv.c
```diff
@@ -20,6 +20,7 @@
  */
 #include "signed_video_tlv.h"
 
+#include <assert.h>  // assert
```
lusikamalo, line 23: Remove line. No assert used in this file.

lib/src/signed_video_h26x_auth.c
```diff
@@ -736,6 +737,44 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
+        sizeof(gop_info_detected_t));
+    nalu_list->gop_idx++;
+  } else {
+    DEBUG_LOG("Warning: Number of pending gops exeeds limit > %d", NR_OF_PENDING_GOPS);
+    return SVI_MEMORY;
+  }
+
+  // Check if public_key has been received
+  if (!self->has_public_key && gop_state->signing_present) {
+    bool public_key_decoded = false;
```
bjornvolcker, line 755: public_key_decoded -> recurrent_data_decoded

lib/src/signed_video_h26x_auth.c
```diff
@@ -736,6 +737,44 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
   // We cannot end up in AUTH_STATE_VALIDATE if the NALU is not hashable.
   assert(nalu->is_hashable);
 
+  // Copy gop_info_detected and gop_state to struct h26x_nalu_list_t. This is needed if the public
+  // key arrives late. When public key eventually arrives correct gop_info_detected and gop_state
+  // can be used for that specific gop.
```
bjornvolcker, line 742: I would put this in a separate function.

lib/src/signed_video_h26x_auth.c
```diff
@@ -744,30 +783,43 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
+      memcpy(
+          gop_info_detected, &nalu_list->gop_info_detected_pending[i], sizeof(gop_info_detected_t));
 
-    if (!gop_state->signing_present) {
-      verify_hashes_without_sei(self);
-      latest->authenticity = SV_AUTH_RESULT_NOT_SIGNED;
-    } else {
-      validate_authenticity(self);
-    }
+      if (!first_validation_check) gop_state->is_first_validation = false;
```
bjornvolcker, line 794: Would the entire usage of |first_validation_check| be equivalent with
```
if (i > 0) gop_state->is_first_validation = false;
```

lib/src/signed_video_h26x_auth.c
```diff
@@ -744,30 +783,43 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
+
+      // The flag |is_first_validation| is used to ignore the first validation if we start the
+      // validation in the middle of a stream. Now it is time to reset it.
+      gop_state->is_first_validation = false;
+      first_validation_check = false;
+
+      // Reset the gop_state_t and gop_info_detected_t.
+      gop_state_reset(gop_state, gop_info_detected);
+      // If we find a pending SEI and it is the latest NALU the state should be
+      // AUTH_STATE_WAIT_FOR_NEXT_NALU.
```
bjornvolcker, line 813: This special operation is not handled when storing gop_states on line 744-745. Is this case tested?
Try removing the first GOP in test lost_all_nalus_between_two_seis instead

lib/src/signed_video_h26x_internal.h
```diff
@@ -29,6 +29,8 @@
 typedef struct _h26x_nalu_list_item_t h26x_nalu_list_item_t;
 typedef struct _h26x_nalu_t h26x_nalu_t;
 
+#define NR_OF_PENDING_GOPS 120
```
bjornvolcker, line 32: MAX_PENDING_GOPS
otherwise when reading code it could be interpreted as how many pending GOPs we have right now.

lib/src/signed_video_h26x_internal.h
```diff
@@ -62,6 +64,10 @@ struct _h26x_nalu_list_t {
   h26x_nalu_list_item_t *last_item;  // Points to the last item in the linked list, that is, the
   // latest NALU added for validation.
   int num_items;  // The number of items linked together in the list.
+
+  gop_state_t gop_state_pending[NR_OF_PENDING_GOPS];
+  gop_info_detected_t gop_info_detected_pending[NR_OF_PENDING_GOPS];
+  int gop_idx;
```
bjornvolcker, line 70: Add comment on what these are.

lib/src/signed_video_h26x_auth.c
```diff
@@ -744,30 +783,43 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
 
   svi_rc status = SVI_UNKNOWN;
   SVI_TRY()
-    SVI_THROW(prepare_for_validation(self));
+    // |first_validation_check| keeps track of first validation when looping through pending gops
+    bool first_validation_check = true;
+    // Loop through possible pending gops and validate them
+    for (int i = 0; i < nalu_list->gop_idx; i++) {
```
bjornvolcker, line 789: With a fixed list we can blow the limit if the user sets a very long recurrence, which you have a check for on line 743. Also, if we lose the SEIs with the public_key number of pending GOPs will/can grow past the limit.

We should consider using a linked list instead.
Unfortunately the auth-side does not know what recurrence was used in camera. Could we benefit from transmitting the recurrence interval in the general tag?

lib/src/signed_video_h26x_internal.h
```diff
@@ -205,4 +211,24 @@ parse_nalu_info(const uint8_t *nalu_data,
 SignedVideoReturnCode
 signed_video_set_recurrence_interval(signed_video_t *self, int recurrence);
 
+#ifdef UNIT_TEST
+/**
+ * @brief Sets the recurrence offset for the signed video session
+ *
+ * Without an offset the recurrent tags are included in the first SEI. But with an offset the
+ * recurrent tags are included later on dependent on offset.
```
bjornvolcker, line 219: "on dependent on offset" weird phrasing

lib/src/signed_video_h26x_internal.h
```diff
@@ -205,4 +211,24 @@ parse_nalu_info(const uint8_t *nalu_data,
...
+ *
+ * This API to set the recurrence offset is only possible to use when executing unit tests.
```
bjornvolcker, line 221: Remove "to set the recurrence offset"

lib/src/signed_video_h26x_sign.c
```diff
@@ -597,3 +597,19 @@ signed_video_set_recurrence_interval(signed_video_t *self, int recurrence)
...
+
+#ifdef UNIT_TEST
+SignedVideoReturnCode
+signed_video_set_recurrence_offset(signed_video_t *self, int offset)
+{
+  if (!self) return SV_INVALID_PARAMETER;
+
+  svi_rc status;
+  SVI_TRY()
+    SVI_THROW_IF_WITH_MSG(offset > self->recurrence, SVI_NOT_SUPPORTED, "Invalid offset");
```
bjornvolcker, line 609: Shouldn't it even be >=

lib/src/signed_video_h26x_sign.c
```diff
@@ -597,3 +597,19 @@ signed_video_set_recurrence_interval(signed_video_t *self, int recurrence)
...
+    self->recurrence_offset = offset;
+  SVI_CATCH()
+  SVI_DONE(status)
+  return svi_rc_to_signed_video_rc(status);
```
bjornvolcker, line 613: Add empty line before return

lib/src/signed_video_internal.h
```diff
@@ -170,6 +171,8 @@ struct _signed_video_t {
   uint8_t *arbitrary_data;  // Enables the user to transmit user specific data and is automatically
   // sent through the ARBITRARY_DATA_TAG.
   size_t arbitrary_data_size;  // Size of |arbitrary_data|.
+
+  bool has_public_key;  // State to indicate if public key is received.
```
bjornvolcker, line 175: "received" -> "received/added"

lib/src/signed_video_tlv.h
```diff
@@ -149,4 +149,24 @@ write_byte(uint16_t *last_two_bytes,
...
+ *
+ * @param signed_video Pointer to the signed_video_t session.
+ * @param tlv_data Pointer to the TLV data to scan.
+ * @param tlv_data_size Size of the TLV data.
+ * @param with_ep Flag to indicate if emulation prevention bytes is on.
+ *
+ * @returns SVI_OK if find and decoding tag was successful, otherwise an error code.
+ */
+svi_rc
+tlv_find_tag_and_decode_recurrent_tags(signed_video_t *self,
```
bjornvolcker, line 167: Change to `tlv_find_and_decode_recurrent_tags`

lib/src/signed_video_tlv.c
```diff
@@ -902,6 +904,56 @@ tlv_find_tag(signed_video_t *self,
   return NULL;
 }
 
+svi_rc
+tlv_find_tag_and_decode_recurrent_tags(signed_video_t *self,
+    const uint8_t *tlv_data,
+    size_t tlv_data_size,
+    bool with_ep)
```
bjornvolcker, line 911: emulation prevention bytes have already been removed if we use this on the auth side

lib/src/signed_video_tlv.c
```diff
@@ -902,6 +904,56 @@ tlv_find_tag(signed_video_t *self,
...
+  while (tlv_data_ptr < tlv_data + tlv_data_size) {
+    latest_tag_location = tlv_data_ptr;
+    // Read the tag
+    sv_tlv_tag_t this_tag = read_byte(&last_two_bytes, &tlv_data_ptr, with_ep);
+    // If it is a recurrent tag then decode it
+    if (!tlv_tuples[this_tag].is_always_present) {
+      size_t tlv_header_size = 0;
+      size_t length = 0;
+      sv_tlv_tag_t tag_check = 0;
+      status = decode_tlv_header(latest_tag_location, &tlv_header_size, &tag_check, &length);
```
bjornvolcker, line 929: Can't we use decode_tlv_header already above when extracting the tag?

```
while (tlv_data_ptr < tlv_data + tlv_data_size) {
  size_t tlv_header_size = 0;
  size_t length = 0;
  sv_tlv_tag_t this_tag = UNDEFINED_TAG;
  status = decode_tlv_header(tlv_data_ptr, &tlv_header_size, &this_tag, &length);
  tlv_data_ptr += tlv_header_size;
  if (!tlv_tuples[this_tag].is_always_present) {
    sv_tlv_decoder_t decoder = get_decoder(this_tag);
    status = decoder(self, tlv_data_ptr, length);
  }
  tlv_data_ptr += length;
}
```
and handle return values.

meson.build
```diff
@@ -15,6 +15,8 @@ if get_option('test-settings')
   endif
   # Set MAX_GOP_LENGTH to 10 in tests to avoid long streams
   add_global_arguments('-DMAX_GOP_LENGTH=10', language : 'c')
+  # Option for code related to unit tests
+  add_global_arguments('-DUNIT_TEST', language : 'c')
```
bjornvolcker, line 19: meson has changed so We'll fix this later.
Maybe change define name to SV_UNIT_TEST, in case someone else out there has the same define.

lib/src/signed_video_h26x_auth.c
```diff
@@ -706,8 +706,9 @@ prepare_for_validation(signed_video_t *self)
       // TODO: Is it possible to avoid a memcpy by using a pointer strategy?
       memcpy(signature_info->hash, self->gop_info->gop_hash, HASH_DIGEST_SIZE);
     }
-    SVI_THROW_IF_WITH_MSG(gop_state->signing_present && !gop_state->has_public_key, SVI_UNKNOWN,
-        "No public key found");
+
+    SVI_THROW_IF_WITH_MSG(
+        gop_state->signing_present && !self->has_public_key, SVI_UNKNOWN, "No public key found");
```
bjornvolcker, line 711: "No public key present" since we do not search for it here. That's done elsewhere.

lib/src/signed_video_h26x_auth.c
```diff
@@ -720,6 +721,41 @@ prepare_for_validation(signed_video_t *self)
   return status;
 }
 
+// If public_key is not received then try to decode all recurrent tags.
+static bool
+is_recurrent_data_decoded(signed_video_t *self)
```
bjornvolcker, line 726: Maybe declare this function at top.

lib/src/signed_video_h26x_auth.c
```diff
@@ -720,6 +721,41 @@ prepare_for_validation(signed_video_t *self)
...
+{
+  gop_state_t *gop_state = &(self->gop_state);
+  gop_info_detected_t *gop_info_detected = &(self->gop_info_detected);
+  signed_video_latest_validation_t *latest = self->latest_validation;
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
+
+  if (!self->has_public_key && gop_state->signing_present) {
```
bjornvolcker, line 733: Since the whole function is sort of wrapped in within this if-statement I suggest turning it into a sanity check.

lib/src/signed_video_h26x_auth.c
```diff
@@ -720,6 +721,41 @@ prepare_for_validation(signed_video_t *self)
...
+    bool recurrent_data_decoded = false;
```
bjornvolcker, line 734: You can use |recurrent_data_decoded| as the return value, which also makes it clearer what is returned.

lib/src/signed_video_h26x_auth.c
```diff
@@ -720,6 +721,41 @@ prepare_for_validation(signed_video_t *self)
...
+    h26x_nalu_list_item_t *item = nalu_list->first_item;
+
+    while (item && !recurrent_data_decoded) {
+      if (item->nalu && item->nalu->is_gop_sei && item->validation_status == 'P') {
+        const uint8_t *tlv_data = item->nalu->tlv_data;
+        size_t tlv_size = item->nalu->tlv_size;
+        recurrent_data_decoded =
+            tlv_find_and_decode_recurrent_tags(self, tlv_data, tlv_size) == SVI_OK;
+        if (!recurrent_data_decoded) {
+          DEBUG_LOG("Failed to find and decode recurrent tags");
```
bjornvolcker, line 744: I find it redundant to have this DEBUG_LOG when we have another one below. This will be triggered until we hit one including the key and displaying "Failed" will indicate something went wrong.

lib/src/signed_video_h26x_auth.c
```diff
@@ -720,6 +721,41 @@ prepare_for_validation(signed_video_t *self)
...
+        }
+      }
+      item = item->next;
+    }
+    if (!recurrent_data_decoded) {
+      DEBUG_LOG("Public key missing");
+      gop_state_reset(gop_state, gop_info_detected);
```
bjornvolcker, line 751: I think we should add a TODO to investigate if we need to take any special actions when moving to the next GOP. Recall the test we did the other day.

lib/src/signed_video_h26x_auth.c
```diff
@@ -736,6 +772,26 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
+        sizeof(gop_info_detected_t));
+    nalu_list->gop_idx++;
+  } else {
+    DEBUG_LOG("Warning: Number of pending gops exeeds limit > %d", MAX_PENDING_GOPS);
+    return SVI_MEMORY;
+  }
+
+  bool recurrent_data_decoded = false;
+  recurrent_data_decoded = is_recurrent_data_decoded(self);
+  if (!recurrent_data_decoded) {
```
bjornvolcker, line 790: You can call the function in the if directly

lib/src/signed_video_h26x_auth.c
```diff
@@ -736,6 +772,26 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
+    DEBUG_LOG("Recurrent data not decoded");
```
bjornvolcker, line 791: "No recurrent data received yet"

lib/src/signed_video_internal.h
```diff
@@ -170,6 +171,8 @@ struct _signed_video_t {
   uint8_t *arbitrary_data;  // Enables the user to transmit user specific data and is automatically
   // sent through the ARBITRARY_DATA_TAG.
   size_t arbitrary_data_size;  // Size of |arbitrary_data|.
+
+  bool has_public_key;  // State to indicate if public key is received/added
```
bjornvolcker, line 175: Can we remove the one in gop_state now?

lib/src/signed_video_internal.h
```diff
@@ -53,6 +53,7 @@ typedef struct _h26x_nalu_list_t h26x_nalu_list_t;
 
 #define DEFAULT_MAX_GOP_LENGTH 300
 #define RECURRENCE_MIN 1
```
bjornvolcker, line 55: I think we should rename "RECURRENCE_MIN" to "RECURRENCE_ALWAYS" or something to indicate that it means it is always present.

tests/check/check_h26xsigned_sign.c
bjornvolcker: I have an upcoming change for this

tests/check/signed_video_helpers.h
```diff
@@ -32,13 +32,21 @@
 #define MANUFACT "manufacturer"
 #define ADDR "address"
 
+typedef enum { SV_RECURRENCE_DEFAULT = 1, SV_RECURRENCE_THREE = 3 } SignedVideoRecurrence;
+typedef enum {
+  SV_RECURRENCE_OFFSET_DEFAULT = 0,
```
bjornvolcker, line 37: Is it possible to use RECURRENCE_OFFSET_DEFAULT instead of 0 to avoid bugs in case of change

lib/src/signed_video_h26x_auth.c
```diff
@@ -720,6 +723,40 @@ prepare_for_validation(signed_video_t *self)
...
+          tlv_find_and_decode_recurrent_tags(self, tlv_data, tlv_size) == SVI_OK;
+    }
+    item = item->next;
+  }
+  if (!recurrent_data_decoded) {
+    DEBUG_LOG("Public key missing");
+    // TODO: Investigate if it's needed to take any special actions when moving to the next GOP
+    gop_state_reset(gop_state, gop_info_detected);
+    latest->authenticity = SV_AUTH_RESULT_SIGNATURE_PRESENT;
+    return recurrent_data_decoded;
```
bjornvolcker, line 754: You can remove this return

lib/src/signed_video_h26x_sign.c
```diff
@@ -590,10 +590,30 @@ signed_video_set_recurrence_interval(signed_video_t *self, int recurrence)
 
   svi_rc status;
   SVI_TRY()
+    SVI_THROW_IF_WITH_MSG(recurrence < RECURRENCE_ALWAYS, SVI_NOT_SUPPORTED,
+        "Recurrence interval < RECURRENCE_ALWAYS");
     SVI_THROW_IF_WITH_MSG(
-        recurrence < RECURRENCE_MIN, SVI_NOT_SUPPORTED, "Invalid recurrence interval");
+        recurrence > MAX_GOP_LENGTH, SVI_NOT_SUPPORTED, "Recurrence interval > MAX_GOP_LENGTH");
```
bjornvolcker, line 596: This is wrong. MAX_GOP_LENGTH is number of frames per GOP. Recurrence counts in GOPs, not in frames.
Further, I don't think there should be an upper limit. There could be applications where it is super long, and still would work on the validation side, because all streams get the first SEI.

lib/src/signed_video_h26x_sign.c
```diff
@@ -590,10 +590,30 @@ signed_video_set_recurrence_interval(signed_video_t *self, int recurrence)
...
+signed_video_set_recurrence_offset(signed_video_t *self, int offset)
+{
+  if (!self) return SV_INVALID_PARAMETER;
+
+  svi_rc status;
+  SVI_TRY()
+    SVI_THROW_IF_WITH_MSG(offset >= self->recurrence, SVI_NOT_SUPPORTED, "Invalid offset");
+    self->recurrence_offset = offset;
+  SVI_CATCH()
+  SVI_DONE(status)
```
bjornvolcker, line 615: Maybe no need for TRY/CATCH here since it is only one try. The check is a sanity check.

lib/src/signed_video_h26x_auth.c
```diff
@@ -737,6 +773,24 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
   // We cannot end up in AUTH_STATE_VALIDATE if the NALU is not hashable.
   assert(nalu->is_hashable);
 
+  // Copy gop_info_detected and gop_state to struct h26x_nalu_list_t. This is needed if the public
+  // key arrives late. When public key eventually arrives correct gop_info_detected and gop_state
```
bjornvolcker, line 777: "," after "arrives" and put || around variables.

lib/src/signed_video_internal.h
```diff
@@ -147,6 +147,7 @@ struct _signed_video_t {
   gop_state_t gop_state;
   gop_info_detected_t gop_info_detected;
   int recurrence;
+  int recurrence_offset;
```
bjornvolcker, line 150: Maybe we should use `unsigned` on both `recurrence` and `recurrence_offset`

lib/src/signed_video_tlv.h
```diff
@@ -144,4 +144,21 @@ write_byte(uint16_t *last_two_bytes,
 uint8_t
 read_byte(uint16_t *last_two_bytes, const uint8_t **payload, bool do_emulation_prevention);
 
+/**
+ * @brief Scans the TLV part of a SEI payload and decodes all tags dependent on recurrency.
+ *
+ * The data is assumed to have been written in a TLV format. tlv_find_and_decode_recurrent_tags
+ * parses data and finds all tags dependent on recurrency and decodes them.
+ *
+ * @param signed_video Pointer to the signed_video_t session.
```
bjornvolcker, line 153: signed_video -> self

meson.build
```diff
@@ -9,6 +9,12 @@ cc = meson.get_compiler('c')
 
 check_dep = dependency('check', required : false)
 
+if check_dep.found()
+  # Option for code related to unit tests
+  # Can't be used to change defines/settings in the lib
```
bjornvolcker, line 14: "Do not use to change defines and settings in the lib"

lib/src/signed_video_tlv.c
```diff
@@ -902,6 +903,43 @@ tlv_find_tag(const uint8_t *tlv_data, size_t tlv_data_size, sv_tlv_tag_t tag, bo
   return NULL;
 }
 
+svi_rc
```
bjornvolcker, line 906: This function is used to as a boolean, by checking if it returns SVI_OK. How about returning a bool directly?

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -806,6 +805,7 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
       memcpy(
           gop_info_detected, &nalu_list->gop_info_detected_pending[i], sizeof(gop_info_detected_t));
 
+      // Need to force setting of |is_first validation| since it is overwritten in the memcpy above
```
bjornvolcker, line 808: remove "of"

lib/src/signed_video_tlv.c
```diff
@@ -902,6 +902,42 @@ tlv_find_tag(const uint8_t *tlv_data, size_t tlv_data_size, sv_tlv_tag_t tag, bo
...
+    size_t length = 0;
+    sv_tlv_tag_t this_tag = UNDEFINED_TAG;
+    status = decode_tlv_header(tlv_data_ptr, &tlv_header_size, &this_tag, &length);
+    if (status != SVI_OK) {
+      DEBUG_LOG("Could not decode tlv header");
+      break;
+    }
+    tlv_data_ptr += tlv_header_size;
+    if (!tlv_tuples[this_tag].is_always_present) {
+      recurrent_tags_found = true;
```
bjornvolcker, line 927: Setting `recurrent_tags_found` should be done after the status check, that is, after line 933. Otherwise we will mark as `true` even if we failed decoding it.

lib/src/signed_video_tlv.c
```diff
@@ -902,6 +902,42 @@ tlv_find_tag(const uint8_t *tlv_data, size_t tlv_data_size, sv_tlv_tag_t tag, bo
...
+tlv_find_and_decode_recurrent_tags(signed_video_t *self,
+    const uint8_t *tlv_data,
+    size_t tlv_data_size)
+{
+  const uint8_t *tlv_data_ptr = tlv_data;
+
+  if (!self || !tlv_data || tlv_data_size == 0) return false;
+
+  svi_rc status = SVI_UNKNOWN;
+  bool recurrent_tags_found = false;
```
bjornvolcker, line 915: Maybe change name to `recurrent_tags_decoded`

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #37, opened 2022-01-26
Always generate new private keys

bjornvolcker:
create_rs_private_key() and create_ecdsa_private_key() both always generate new keys. That is, if a key already exists in the specified location it will be overwritten. Previously, no new key was created if one already existed.


Caroline-Holmstrom (approved):
LGTM

PR #38, opened 2022-01-31
Detects changed public key

Caroline-Holmstrom:
Adds a test for detecting a change in public key.
Detection is done when decoding the TLV tag.


bjornvolcker (approved):
LGTM

lusikamalo (approved):
OK.

PR #40, opened 2022-02-03
Splits validation status 'U' into 'U' and '_'

bjornvolcker:
'U' now marks an unknown authenticity, for example when a NALU header
is broken.
'_' marks an ignored NALU, which is not hashed and therefore not part
of the signature.

A new SignedVideoFrameType (NALU_TYPE_OTHER) for 'uninteresting' NALUs.


lusikamalo (approved):
LGTM.

Caroline-Holmstrom (approved):
LGTM

PR #42, opened 2022-02-03
Makes it possible to sign multiple streams

bjornvolcker:
The previous implementation of the unthreaded plugin did not work
if used to sign multiple streams. To scale for other plugins, the
interfaces in signed_video_interfaces.h have changed. Now a handle
should be returned at setup which is used when signing and getting
the signature.

In addition, the sv_interface_get_signature() API has been extended
with a signature_size output since the size is may not be fixed.


Caroline-Holmstrom (dismissed):
LGTM

lusikamalo (approved):
LGTM.

Caroline-Holmstrom (approved):

lusikamalo (approved):
LGTM.

PR #47, opened 2022-02-14
Only generate private_key if neccesary

lusikamalo:
It takes long time to generate private keys during unit tests. Therefore two
copies of private keys are saved. One with algo RSA and one with algo
ECDSA.

### DIFF
--- a/tests/check/check_h26xsigned_auth.c
+++ b/tests/check/check_h26xsigned_auth.c
@@ -264,7 +264,7 @@ START_TEST(intact_stream)
   // |settings|; See signed_video_helpers.h.
 
   // Create a list of NALUs given the input string.
-  nalu_list_t *list = create_signed_nalus("IPPIPPI", settings[_i], false);
+  nalu_list_t *list = create_signed_nalus("IPPIPPI", settings[_i]);
   nalu_list_check_str(list, "GIPPGIPPGI");
 
   // One pending NALU per GOP.
@@ -280,7 +280,7 @@ START_TEST(intact_multislice_stream)
   // This test runs in a loop with loop index _i, corresponding to struct sv_setting _i in
   // |settings|; See signed_video_helpers.h.
 
-  nalu_list_t *list = create_signed_nalus("IiPpPpIiPpPpIi", settings[_i], false);
+  nalu_list_t *list = create_signed_nalus("IiPpPpIiPpPpIi", settings[_i]);
   nalu_list_check_str(list, "GIiPpPpGIiPpPpGIi");
 
   // One pending NALU per GOP.
@@ -299,7 +299,7 @@ START_TEST(intact_stream_with_pps_nalu_stream)
   // This test runs in a loop with loop index _i, corresponding to struct sv_setting _i in
   // |settings|; See signed_video_helpers.h.
 
-  nalu_list_t *list = create_signed_nalus("VIPPIPPI", settings[_i], false);
+  nalu_list_t *list = create_signed_nalus("VIPPIPPI", settings[_i]);
   nalu_list_check_str(list, "VGIPPGIPPGI");
 
   // One pending NALU per GOP.
@@ -315,7 +315,7 @@ START_TEST(intact_stream_with_pps_bytestream)
   // This test runs in a loop with loop index _i, corresponding to struct sv_setting _i in
   // |settings|; See signed_video_helpers.h.
 
-  nalu_list_t *list = create_signed_nalus("VIPPIPPI", settings[_i], false);
+  nalu_list_t *list = create_signed_nalus("VIPPIPPI", settings[_i]);
   nalu_list_check_str(list, "VGIPPGIPPGI");
 
   // Pop the PPS NALU and inject it before the I-NALU.
@@ -338,7 +338,7 @@ START_TEST(intact_ms_stream_with_pps_nalu_stream)
   // This test runs in a loop with loop index _i, corresponding to struct sv_setting _i in
   // |settings|; See signed_video_helpers.h.
 
-  nalu_list_t *list = create_signed_nalus("VIiPpPpIiPpPpIi", settings[_i], false);
+  nalu_list_t *list = create_signed_nalus("VIiPpPpIiPpPpIi", settings[_i]);
   nalu_list_check_str(list, "VGIiPpPpGIiPpPpGIi");
 
   // One pending NALU per GOP.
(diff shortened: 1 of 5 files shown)


tests/check/check_h26xsigned_auth.c
```diff
@@ -264,7 +264,7 @@ START_TEST(intact_stream)
   // |settings|; See signed_video_helpers.h.
 
   // Create a list of NALUs given the input string.
-  nalu_list_t *list = create_signed_nalus("IPPIPPI", settings[_i]);
+  nalu_list_t *list = create_signed_nalus("IPPIPPI", settings[_i], false);
```
bjornvolcker, line 267: Since you use `create_signed_nalus()` almost exclusively with setting `false`, may I suggest to keep `create_signed_nalus()` as before, which implicitly calls `create_signed_nalus_int()` with `new_priv_key = false`. For more info see comment in `signed_video_helpers.h`

tests/check/signed_video_helpers.h
```diff
@@ -32,6 +32,9 @@
 #define MANUFACT "manufacturer"
 #define ADDR "address"
 
+#define RSA_PRIVATE_KEY_ALLOC_BYTES 2000
+#define ECDSA_PRIVATE_KEY_ALLOC_BYTES 1000
```
bjornvolcker, line 36: Can't these be moved to the c-file?

tests/check/signed_video_helpers.c
```diff
@@ -122,10 +129,10 @@ create_signed_nalus_with_sv(signed_video_t *sv, const char *str)
  * generated NALUs are then passed through the signing process and corresponding generated
  * sei-nalus are added to the stream. */
 nalu_list_t *
-create_signed_nalus(const char *str, struct sv_setting settings)
+create_signed_nalus(const char *str, struct sv_setting settings, bool new_priv_key)
```
bjornvolcker, line 132: How about rename this function to `create_signed_nalus_int()` and add back
```
nalu_list_t *
create_signed_nalu(const char *str, struct sv_setting settings)
{
  return create_signed_nalu_int(st, settings, false);
}
```

bjornvolcker (changes requested):

tests/check/signed_video_helpers.c
```diff
@@ -129,7 +139,7 @@ create_signed_nalus_with_sv(signed_video_t *sv, const char *str)
  * generated NALUs are then passed through the signing process and corresponding generated
  * sei-nalus are added to the stream. */
```
bjornvolcker, line 140: Add information on `new_priv_key`

tests/check/signed_video_helpers.h
```diff
@@ -74,7 +75,7 @@ get_initialized_signed_video(SignedVideoCodec codec, sign_algo_t algo, bool new_
  * new_priv_key = Generate a new private key or not.
  */
 nalu_list_t *
-create_signed_nalus(const char *str, struct sv_setting settings, bool new_priv_key);
+create_signed_nalus_int(const char *str, struct sv_setting settings, bool new_priv_key);
```
bjornvolcker, line 78: If it fits on line rename to `new_private_key`

tests/check/signed_video_helpers.h
```diff
@@ -53,6 +50,10 @@ extern const struct sv_setting settings[NUM_SETTINGS];
 signed_video_t *
 get_initialized_signed_video(SignedVideoCodec codec, sign_algo_t algo, bool new_priv_key);
```
bjornvolcker, line 51: If it fits on line rename to `new_private_key`

tests/check/signed_video_helpers.c
```diff
@@ -129,7 +139,7 @@ create_signed_nalus_with_sv(signed_video_t *sv, const char *str)
...
 nalu_list_t *
-create_signed_nalus(const char *str, struct sv_setting settings, bool new_priv_key)
+create_signed_nalus_int(const char *str, struct sv_setting settings, bool new_priv_key)
```
bjornvolcker, line 142: If it fits on line rename to `new_private_key`

bjornvolcker (changes requested):

tests/check/check_h26xsigned_auth.c
```diff
@@ -1211,7 +1211,7 @@ START_TEST(recurrence)
   int recurrence = 3;
 
   nalu_list_t *list =
-      create_signed_nalus_recurrence("IPPIPPIPPIPPIPPIPPI", settings[_i], recurrence);
+      create_signed_nalus_recurrence("IPPIPPIPPIPPIPPIPPI", settings[_i], recurrence, false);
```
bjornvolcker, line 1214: In all places where `create_signed_nalus_recurrence()` is called `new_private_key = false`. I think we can set that implicitly and keep the API as before.

bjornvolcker (changes requested):

tests/check/signed_video_helpers.c
```diff
@@ -20,15 +20,26 @@
...
 #include "lib/src/includes/signed_video_common.h"
 #include "lib/src/includes/signed_video_openssl.h"
 #include "lib/src/includes/signed_video_sign.h"
 #include "lib/src/signed_video_h26x_internal.h"  // signed_video_set_recurrence_interval()
 #include "lib/src/signed_video_internal.h"  // _signed_video_t
 
+#define RSA_PRIVATE_KEY_ALLOC_BYTES 2000
+#define ECDSA_PRIVATE_KEY_ALLOC_BYTES 1000
+
+char global_private_key_rsa[RSA_PRIVATE_KEY_ALLOC_BYTES];
```
bjornvolcker, line 36: One potential naming improvement.
`global_private_key_rsa` -> `private_key_rsa`
and same for ecdsa.

No need for `global` I think.

tests/check/signed_video_helpers.c
```diff
@@ -20,15 +20,26 @@
...
+size_t global_private_key_size_rsa;
+sign_algo_t global_algo_rsa;
```
bjornvolcker, line 38: You do not need `global_algo_rsa`, see comment where it is used.

tests/check/signed_video_helpers.c
```diff
@@ -162,17 +178,45 @@ create_signed_nalus_recurrence(const char *str,
...
+      rc = signed_video_generate_private_key(algo, "./", &private_key, &private_key_size);
+      ck_assert_int_eq(rc, SV_OK);
+
+      assert(private_key_size < RSA_PRIVATE_KEY_ALLOC_BYTES);
+      memcpy(global_private_key_rsa, private_key, private_key_size);
+      global_private_key_size_rsa = private_key_size;
+      global_algo_rsa = algo;
+    }
+    rc = signed_video_set_private_key(
+        sv, global_algo_rsa, global_private_key_rsa, global_private_key_size_rsa);
```
bjornvolcker, line 202: Since we are inside the if statement, algo is already known. You can use `algo` directly here. Same goes for ECDSA.

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #48, opened 2022-02-14
Proper action in unthreaded plugin

bjornvolcker:
The generated signature is now stored and kept until the user asks
for it through sv_interface_get_signature().


lusikamalo (approved):
LGTM.

PR #49, opened 2022-02-15
Adds a payload_buffer to store ongoing SEIs

bjornvolcker:
When a SEI is about to be created everything up to the signature is
written. It is then completed when the signature exists. This means
that SEIs can pile up and storing pointers to these memory slots is
necessary for other signing plugins.

The members payload_ptr and signature_payload_ptr have been removed
since the are not used anymore.

### DIFF
--- a/lib/src/signed_video_h26x_sign.c
+++ b/lib/src/signed_video_h26x_sign.c
@@ -34,13 +34,21 @@
 static void
 h26x_set_nal_uuid_type(signed_video_t *self, uint8_t **payload, SignedVideoUUIDType uuid_type);
 static size_t
-get_sign_and_complete_sei_nalu(signed_video_t *self, uint8_t **payload);
+get_sign_and_complete_sei_nalu(signed_video_t *self,
+    uint8_t **payload,
+    uint8_t *payload_signature_ptr);
+
+/* Functions for payload_buffer. */
+static void
+add_payload_to_buffer(signed_video_t *self, uint8_t *payload_ptr, uint8_t *payload_signature_ptr);
+static svi_rc
+complete_sei_nalu_and_add_to_prepend(signed_video_t *self);
 
 /* Functions related to the list of NALUs to prepend. */
 static void
 reset_nalu_to_prepend(signed_video_nalu_to_prepend_t *nalu_to_prepend);
 static svi_rc
-generate_sei_nalu(signed_video_t *self, uint8_t **payload);
+generate_sei_nalu(signed_video_t *self, uint8_t **payload, uint8_t **payload_signature_ptr);
 static svi_rc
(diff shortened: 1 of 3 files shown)


Caroline-Holmstrom (dismissed):
LGTM

lib/src/signed_video_h26x_sign.c
```diff
@@ -65,6 +73,92 @@ h26x_set_nal_uuid_type(signed_video_t *self, uint8_t **payload, SignedVideoUUIDT
...
+  SignedVideoPrependInstruction prepend_instruction = SIGNED_VIDEO_PREPEND_NOTHING;
+  size_t data_size = 0;
+  svi_rc status = SVI_UNKNOWN;
+
+  // If the signature could not be generated |signature_size| equals zero. Free the started SEI and
+  // move on. This is a valid operation. What will happen is that the video will have an unsigned
+  // GOP.
+  if (self->signature_info->signature_size == 0) {
+    signed_video_nalu_data_free(payload);
+    payload = NULL;
```
lusikamalo, line 125: Should we set "payload_signature_ptr" to NULL as well?

lib/src/signed_video_h26x_sign.c
```diff
@@ -65,6 +73,92 @@ h26x_set_nal_uuid_type(signed_video_t *self, uint8_t **payload, SignedVideoUUIDT
...
+  } else if (!payload) {
+    // No more pending payloads. Already freed due to too many unsigned SEIs.
+    status = SVI_OK;
+    goto done;
+  }
+
+  SVI_TRY()
+    // Add the signature to the SEI payload.
+    data_size = get_sign_and_complete_sei_nalu(self, &payload, payload_signature_ptr);
+    SVI_THROW_IF(!data_size, SVI_UNKNOWN);
```
lusikamalo, line 137: Maybe have "get_sign_and_complete_sei_nalu" inside SVI_THROW_IF?

lusikamalo (changes requested):
Some minor comments.

lib/src/signed_video_h26x_sign.c
bjornvolcker: Unfortunately, `data_size` is needed on line 143

lib/src/signed_video_h26x_sign.c
bjornvolcker: definitely a good idea.

lusikamalo: Thanx.

lib/src/signed_video_h26x_sign.c
lusikamalo: Good.

lusikamalo (approved):
LGTM.

PR #50, opened 2022-02-16
Adds a threaded plugin

bjornvolcker:
The new plugin adds a dependency on threads and glib-2.0. It sets up
a worker thread from which the OpenSSL signing is called.

A new meson-option (signingplugin) to select between the two plugins is
added. There is an opation to use the threaded plugin unless we build with
libcheck (threaded_unless_check_dep). Note that the check tests only works
as intended with the unthreaded plugin.

### DIFF
--- a/lib/plugins/README.md
+++ b/lib/plugins/README.md
@@ -2,16 +2,41 @@
 
 # Signed Video Framework plugins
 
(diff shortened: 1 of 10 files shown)


lusikamalo (dismissed):
LGTM.

PR #51, opened 2022-02-16
Upgrades version to v1.1.0

bjornvolcker:
Note that the initial version character has changed from 'R' to 'v'.
This is more in line with other repos.

Main features included:
 - threaded signing plugin
 - less frequent metadata can now be transmitted according to a
   recurrence interval defined in number of GOPs


lusikamalo (approved):
OK.

PR #53, opened 2022-02-18
Add interface to set recurrence interval in frames

lusikamalo:


lib/src/includes/signed_video_sign.h
```diff
@@ -302,4 +302,22 @@ SignedVideoReturnCode
 signed_video_set_authenticity_level(signed_video_t *self,
     SignedVideoAuthenticityLevel authenticity_level);
 
+/**
+ * @brief Sets the average recurrence interval for the signed video session in frames
+ *
+ * This API will set the recurrence interval which decides how often metadata is sent. Some
+ * metadata is only needed to send once, such as public key and product info. Metadata that can
+ * be sent less often is the public key, product info and attestation report.
```
bjornvolcker, line 310: Metadata that is only needed once when validating the authenticity can be transmitted with a different recurrence interval than the signatures. This API sets that recurrence, counted in frames. Note that this type of metadata is still bundled together in the same SEI as the signature, hence the true recurrence will be correct on the average.

Example of metadata that is only needed once are the public key and product info.

tests/check/check_h26xsigned_sign.c
```diff
@@ -129,6 +129,12 @@ START_TEST(api_inputs)
   ck_assert_int_eq(sv_rc, SV_INVALID_PARAMETER);
   sv_rc = signed_video_set_recurrence_interval(sv, 0);
   ck_assert_int_eq(sv_rc, SV_NOT_SUPPORTED);
+  sv_rc = signed_video_set_recurrence_interval_frames(sv, 1);
+  ck_assert_int_eq(sv_rc, SV_OK);
```
bjornvolcker, line 133: Put the correct call after the incorrect ones. If we do not allow a parameter to change, the order is important.

lib/src/includes/signed_video_sign.h
```diff
@@ -302,4 +302,24 @@ SignedVideoReturnCode
 signed_video_set_authenticity_level(signed_video_t *self,
     SignedVideoAuthenticityLevel authenticity_level);
 
+/**
+ * @brief Sets the average recurrence interval for the signed video session in frames
+ *
+ * Metadata that is only needed once when validating the authenticity can be transmitted with a
+ * different recurrence interval than the signatures. This API sets that recurrence, counted in
+ * frames. Note that this type of metadata is still bundled together in the same SEI as the
+ * signature, hence the true recurrence will be correct on the average.
```
bjornvolcker, line 311: Do we need to say something about frames vs NALUs?

lib/src/signed_video_h26x_sign.c
```diff
@@ -497,6 +501,13 @@ signed_video_add_nalu_for_signing(signed_video_t *self,
 
     SVI_THROW_IF(nalu.is_valid < 0, SVI_INVALID_PARAMETER);
```
bjornvolcker, line 503:
```
// Note that |recurrence| is counted in frames and not in NALUs, hence we only increment the counter for primary slices.
```

lib/src/signed_video_h26x_sign.c
```diff
@@ -497,6 +501,13 @@ signed_video_add_nalu_for_signing(signed_video_t *self,
...
 
+    if (nalu.is_primary_slice) {
+      if (((self->frame_count + self->recurrence_offset) % self->recurrence) == 0) {
+        self->nr_of_frames_for_recurrence_passed = true;
+      }
+      self->frame_count++;
```
bjornvolcker, line 508: Add a comment on that wraparound is OK

lib/src/signed_video_internal.h
```diff
@@ -152,6 +152,10 @@ struct _signed_video_t {
   unsigned recurrence;
   unsigned recurrence_offset;
 
+  // Frame counter and flag to handle recurrence in frames functionality
+  bool nr_of_frames_for_recurrence_passed;
```
bjornvolcker, line 156: Maybe `add_recurrent_data`

tests/check/check_h26xsigned_sign.c
```diff
@@ -563,17 +562,24 @@ START_TEST(recurrence)
   nalu_list_check_str(list, "GIPPGIPPGIPPGIPPGIPPGIPPGI");
 
   nalu_list_item_t *item;
-  int gop_counter = 0;
-  for (int i = 1; i <= (list->num_items); i++) {
+  int last_sei_size = 0;
```
bjornvolcker, line 565: You have made a mistake when resolving this section.

bjornvolcker (changes requested):

tests/check/check_h26xsigned_sign.c
lusikamalo: Good catch. Also the description for this test needs to be changed. It's still about GOPs instead of frames.

lib/src/includes/signed_video_sign.h
```diff
@@ -307,8 +307,8 @@ signed_video_set_authenticity_level(signed_video_t *self,
  *
  * Metadata that is only needed once when validating the authenticity can be transmitted with a
  * different recurrence interval than the signatures. This API sets that recurrence, counted in
- * frames. Note that this type of metadata is still bundled together in the same SEI as the
- * signature, hence the true recurrence will be correct on the average.
+ * frames(not NALUs). Note that this type of metadata is still bundled together in the same SEI
```
bjornvolcker, line 310: white space before (

lib/src/signed_video_internal.h
```diff
@@ -153,7 +153,7 @@ struct _signed_video_t {
   unsigned recurrence_offset;
 
   // Frame counter and flag to handle recurrence in frames functionality
```
bjornvolcker, line 155: Remove "in frames functionality"

tests/check/check_h26xsigned_sign.c
```diff
@@ -580,6 +577,7 @@ START_TEST(recurrence)
           ck_assert_int_gt(item->data_size, last_sei_size);
         }
       }
+      last_sei_size = item->data_size;
```
bjornvolcker, line 580: You have still removed the entire change from master. The test on master is more robust. Please add it back.

lib/src/signed_video_internal.h
```diff
@@ -153,7 +153,7 @@ struct _signed_video_t {
...
-  bool nr_of_frames_for_recurrence_passed;
+  bool add_recurrent_data;
```
bjornvolcker, line 156: Changed my mind. `recurrent_data_ready` or simply `has_recurrent_data`

When you ready `self->has_recurrent_data` it is self desriptive.

bjornvolcker (changes requested):

tests/check/check_h26xsigned_sign.c
```diff
@@ -544,10 +543,7 @@ START_TEST(undefined_nalu_in_sequence)
...
 
 /* Test description
- * Verify that metadata is sent with recurrence interval equal to 3.
- * Recurrence equal to 1 means that all metadata is sent in every GOP.
- * Recurrence equal to 3 means that all metadata is sent in every second GOP.
- * Metadata that can be sent more rarily are public key and product info.
+ * Verify that metadata is sent with correct recurrence interval in frames on the average.
  * The operation is as follows:
  * 1. Generate a nalu_list with a sequence of signed GOPs.
  * 2. Check the sequence of NALUs.
```
bjornvolcker, line 549: Change text of 3. below. We no longer check the size. We check if the SEI has PUBLIC_KEY_TAG, which is recurrent data.

tests/check/check_h26xsigned_sign.c
```diff
@@ -564,14 +560,33 @@ START_TEST(recurrence)
...
+
   for (int i = 1; i <= (list->num_items); i++) {
     item = nalu_list_get_item(list, i);
     if (strncmp(item->str_code, "G", 1) == 0) {
-      // if ((gop_counter % recurrence) == 0) {
-      if (((gop_counter + settings[_i].recurrence_offset) % settings[_i].recurrence) == 0) {
-        ck_assert(tag_is_present(item, settings[_i].codec, PUBLIC_KEY_TAG));
-      } else {
-        ck_assert(!tag_is_present(item, settings[_i].codec, PUBLIC_KEY_TAG));
+      if (settings[_i].recurrence == SV_RECURRENCE_ONE &&
```
bjornvolcker, line 567: I think this whole chunk can be simplified to
```
if (((gop_counter * 4 + settings[_i].recurrence_offset) / 4 % settings[_i].recurrence) == 0) {
  ck_assert(tag_is_present(item, settings[_i].codec, PUBLIC_KEY_TAG));
} else {
  ck_assert(!tag_is_present(item, settings[_i].codec, PUBLIC_KEY_TAG));
}
```

bjornvolcker (changes requested):

bjornvolcker (approved):

Caroline-Holmstrom (approved):
LGTM

PR #54, opened 2022-02-18
Adds vendor axis communications

bjornvolcker:
Axis Communications specific code for setting and transmitting an Attestation Report.


lusikamalo (approved):
OK.

PR #59, opened 2022-03-18
Fixes a bug when the signing side falls back to light version

bjornvolcker:
The light version here means the SV_AUTHENTICITY_LEVEL_GOP, which kicks in
if the hash_list becomes too long; currently at 300 NALUs.

A test in check_h26xsigned_auth.c has been added and the description of
the corresponding test in check_h26xsigned_sign.c has been updated.

### DIFF
--- a/tests/check/check_h26xsigned_auth.c
+++ b/tests/check/check_h26xsigned_auth.c
@@ -30,6 +30,10 @@
 #include "lib/src/includes/signed_video_openssl.h"  // signed_video_generate_private_key()
 #include "lib/src/includes/sv_vendor_axis_communications.h"
 #endif
+#ifdef SV_UNIT_TEST
+#include "lib/src/signed_video_h26x_internal.h"  // signed_video_set_recurrence_offset()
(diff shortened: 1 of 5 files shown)


tests/check/check_h26xsigned_auth.c
```diff
@@ -1624,6 +1671,7 @@ signed_video_suite(void)
   tcase_add_loop_test(tc, no_signature, s, e);
   tcase_add_loop_test(tc, multislice_no_signature, s, e);
   tcase_add_loop_test(tc, late_public_key_and_no_sei_before_key_arrives, s, e);
+  tcase_add_loop_test(tc, fallback_to_gop_level, 0, NUM_SETTINGS);
```
lusikamalo, line 1674: 0, NUM_SETTINGS -> s, e ?

bjornvolcker: Fixed in 94ae17c

lusikamalo (approved):
LGTM.

PR #60, opened 2022-03-22
Fixes unwanted behavior when signing fails in threaded plugin

bjornvolcker:
If signing fails the error is propagated from the plugin to the library
when sv_interface_get_signature() is called, and further to the user.

NOTE: This causes a change in the public API sv_interface_get_signature().

Version is bumped to 1.1.6.


lusikamalo (approved):
OK.

PR #63, opened 2022-04-22
Adds new state in SignedVideoAuthenticityResult

bjornvolcker:
which indicates to the user that there is a bad mismatch between
versions on signing and validation sides.


lusikamalo (approved):
OK.

Caroline-Holmstrom (approved):
LGTM

PR #64, opened 2022-04-22
Changes behavior of test sei_increase_with_gop_length

bjornvolcker:
It has shown to be flaky with a one byte mismatch in SEI sizes which can
happen if an emulation prevention byte is added. Therefore, the test
allows a slack of one byte.


lusikamalo (approved):
LGTM.

Caroline-Holmstrom (approved):
LGTM

PR #67, opened 2022-04-25
Adds new vendor API set_axis_communications_public_key(...)

bjornvolcker:
which verifies type and size of the public key and keeps a reference
for later use.

It is called when decoding the public key tag if the product info
identifies Axis Communications as manufacturer.


lusikamalo (approved):
OK.

PR #68, opened 2022-04-26
Changes behavior of set_axis_communications_public_key()

bjornvolcker:
now updates public_key_validation if the key is not EC instead of returning
an error.

Also adds checking public_key_validation in Axis test.


lusikamalo (approved):
LGTM.

PR #69, opened 2022-04-28
Properly initializes supplemental_authenticity

bjornvolcker:
in sv_vendor_axis_communications_t.

### DIFF
--- a/lib/vendors/axis-communications/sv_vendor_axis_communications.c
+++ b/lib/vendors/axis-communications/sv_vendor_axis_communications.c
@@ -74,7 +74,15 @@ typedef struct _sv_vendor_axis_communications_t {
 void *
 sv_vendor_axis_communications_setup(void)
 {
-  return calloc(1, sizeof(sv_vendor_axis_communications_t));
+  sv_vendor_axis_communications_t *self = calloc(1, sizeof(sv_vendor_axis_communications_t));
+
+  if (!self) return NULL;
+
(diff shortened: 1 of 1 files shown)


lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -74,7 +74,15 @@ typedef struct _sv_vendor_axis_communications_t {
...
 sv_vendor_axis_communications_setup(void)
 {
-  return calloc(1, sizeof(sv_vendor_axis_communications_t));
+  sv_vendor_axis_communications_t *self = calloc(1, sizeof(sv_vendor_axis_communications_t));
+
+  if (!self) return NULL;
+
+  // Initialize |public_key_validation| to unknown/error.
+  self->supplemental_authenticity.public_key_validation = -1;
+  strcpy(self->supplemental_authenticity.serial_number, "Unknown");
```
lusikamalo, line 83: No problem in this case but keep in mind that strcpy doesn't know about length. |serial_number| has room for 20 bytes if I remember correctly. "Unknown\0" is only 8 bytes so it's fine in this case.

But to be really sure nothing bad happens maybe length should be checked? I have no strong opinion about this but please comment.

If more than 20 bytes is copied I guess that coverity will scream.

lusikamalo (changes requested):
Just one comment for you to look at.

lusikamalo (approved):
OK.

PR #70, opened 2022-04-29
New vendor API get_axis_communications_supplemental_authenticity

bjornvolcker:
to fetch status of public key used to validate the video.

The API is called before validating the authenticity and is triggered
if Axis is the manufacturer.


lusikamalo (approved):
LGTM.

PR #71, opened 2022-04-29
Adds new Axis helper function verify_certificate_chain

bjornvolcker:
that takes a trusted CA and a stack of untrusted certificates as input.
Then verifies the first certificate in the chain and return SVI_VENDOR
if the verification was not successful.

### DIFF
--- a/lib/vendors/axis-communications/sv_vendor_axis_communications.c
+++ b/lib/vendors/axis-communications/sv_vendor_axis_communications.c
@@ -20,9 +20,11 @@
  */
 #include "includes/sv_vendor_axis_communications.h"
 
+#include <assert.h>
 #include <openssl/bio.h>  // BIO_*
 #include <openssl/evp.h>  // EVP_*
 #include <openssl/pem.h>  // PEM_*
+#include <openssl/x509.h>  // X509_*
 #include <stdbool.h>
 #include <stdlib.h>  // malloc, memcpy, calloc, free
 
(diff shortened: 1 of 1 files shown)


lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -69,6 +71,54 @@ typedef struct _sv_vendor_axis_communications_t {
...
+static svi_rc
+verify_certificate_chain(X509 *trusted_ca, STACK_OF(X509) * untrusted_certificates);
+
+// Definitions of static functions.
+
+/* Verifies the untrusted certificate chain.
+ *
+ * Uses the |trusted_ca| to verify the first certificate in the chain. The last certificate verified
+ * is the |attestation_certificate| which will be used to verify the transmitted public key. */
+static ATTR_UNUSED svi_rc
```
lusikamalo, line 84: Remove "ATTR_UNUSED" ?

lusikamalo (changes requested):
Please check one minor comment.

lib/vendors/axis-communications/sv_vendor_axis_communications.c
bjornvolcker: Since the function is not used yet we will get a compilation warning (treated as error) otherwise. It will be removed when function is used in a later commit.

lusikamalo: Oki doki.

lusikamalo (approved):
LGTM.

PR #72, opened 2022-05-02
New vendor API verify_and_parse_certificate_chain

bjornvolcker:
with an incomplete skeleton implementation. Only verifies the certificate
chain.
Further, the trusted root CA has change name to kTrustedAxisRootCA and is
converted to X.509 form at setup and stored in a new member |trusted_ca|.


lusikamalo (dismissed):
OK.

lusikamalo (approved):
OK.

PR #74, opened 2022-05-02
Extracts serial number from the Axis attestation certificate

bjornvolcker:
Also adds a define for string 'Unknown' used in several places.


lusikamalo (approved):
LGTM.

PR #75, opened 2022-05-03
Gets pubkey from Axis attestation certificate

bjornvolcker:
and stores it in a message digest context for use when validating the public
key, used in video authenticity validation.


lusikamalo (approved):
LGTM.

PR #76, opened 2022-05-03
New Axis vendor function deserialize_attestation

bjornvolcker:
which turns the serialized attestation data, part of the SEI, into an
attestation report.

### DIFF
--- a/lib/vendors/axis-communications/sv_vendor_axis_communications.c
+++ b/lib/vendors/axis-communications/sv_vendor_axis_communications.c
@@ -289,12 +289,12 @@ deserialize_attestation(sv_vendor_axis_communications_t *self)
 {
   assert(self);
 
+  if (!self->attestation) return SVI_VENDOR;
+  // The |attestation_size| has to be at least 23 bytes to be deserializable.
+  if (self->attestation_size < 24) return SVI_VENDOR;
+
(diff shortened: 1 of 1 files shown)


lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -250,6 +274,64 @@ verify_and_parse_certificate_chain(sv_vendor_axis_communications_t *self)
...
+ *     - signature (|signature_size| bytes) */
+static svi_rc
+deserialize_attestation(sv_vendor_axis_communications_t *self)
+{
+  assert(self);
+
+  uint8_t *attestation_ptr = (uint8_t *)self->attestation;
+  size_t signature_size = 0;
+
+  if (!self->attestation) return SVI_VENDOR;
```
lusikamalo, line 295: Maybe we should have the check for (!self->attestation) above row 292 where it's used the first time.

bjornvolcker: Great catch!

lusikamalo (changes requested):
No other findings than the single comment I wrote. Waiting for your next patch and I will approve.

lusikamalo (approved):
OK.

PR #78, opened 2022-05-03
Converts Axis public key to uncompressed Weierstrass form

bjornvolcker:
which is the form used in the signed data.


lusikamalo (approved):
OK.

PR #81, opened 2022-05-04
Fixes a bug in set_axis_communications_public_key

bjornvolcker:
Now truly skips setting the public key if previous validation failed.
Further, the check if pointer to public_key is the same is no guarantee
the keys differ, and comparing the content is not enough either since
the pointers may differ. To be completely safe, we always check the
size and type of the public_key.


lusikamalo (approved):
OK.

PR #89, opened 2022-05-18
Fixes a bug in signed_video_reset(...)

bjornvolcker:
where global_gop_counter_is_synced was not reset. This caused incorrect
behavior when seeking in a file. Tests were not implemented correctly,
missing this flaw, but have now been updated.

Version is bumped to v1.1.8.


lusikamalo (approved):
LGTM.

PR #92, opened 2022-07-13
Puts the signing requests in a queue

Caroline-Holmstrom:
Puts the signing requests in a queue one by one and gets the oldest
signing request from the queue.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_sign.c
+++ b/lib/src/signed_video_h26x_sign.c
@@ -40,7 +40,10 @@ get_sign_and_complete_sei_nalu(signed_video_t *self,
 
 /* Functions for payload_buffer. */
 static void
-add_payload_to_buffer(signed_video_t *self, uint8_t *payload_ptr, uint8_t *payload_signature_ptr, uint16_t last_two_bytes);
+add_payload_to_buffer(signed_video_t *self,
+    uint8_t *payload_ptr,
+    uint8_t *payload_signature_ptr,
+    uint16_t last_two_bytes);
 static svi_rc
 complete_sei_nalu_and_add_to_prepend(signed_video_t *self);
 
@@ -88,7 +91,10 @@ free_payload_buffer(uint8_t *payload_buffer[])
 
 /* Adds the |payload| to the next available slot in |payload_buffer|. */
 static void
-add_payload_to_buffer(signed_video_t *self, uint8_t *payload, uint8_t *payload_signature_ptr, uint16_t last_two_bytes)
+add_payload_to_buffer(signed_video_t *self,
+    uint8_t *payload,
+    uint8_t *payload_signature_ptr,
+    uint16_t last_two_bytes)
 {
   assert(self);
 
@@ -100,7 +106,7 @@ add_payload_to_buffer(signed_video_t *self, uint8_t *payload, uint8_t *payload_s
 
   self->payload_buffer[self->payload_buffer_idx] = payload;
   self->payload_buffer[self->payload_buffer_idx + 1] = payload_signature_ptr;
-  if (self->payload_buffer_idx != 0){
+  if (self->payload_buffer_idx != 0) {
     self->last_two_bytes[self->payload_buffer_idx / 2] = last_two_bytes;
   } else {
     self->last_two_bytes[self->payload_buffer_idx] = last_two_bytes;
@@ -165,9 +171,11 @@ done:
   // failure.
   for (int j = 0; j < buffer_end - 2; j++) {
     self->payload_buffer[j] = self->payload_buffer[j + 2];
+    self->last_two_bytes[j] = self->last_two_bytes[j + 1];
   }
   self->payload_buffer[buffer_end - 1] = NULL;
   self->payload_buffer[buffer_end - 2] = NULL;
+  self->last_two_bytes[(buffer_end / 2) - 1] = LAST_TWO_BYTES_INIT_VALUE;
   self->payload_buffer_idx -= 2;
 
   return status;
--- a/lib/src/signed_video_h26x_sign.c
+++ b/lib/src/signed_video_h26x_sign.c
@@ -40,10 +40,7 @@ get_sign_and_complete_sei_nalu(signed_video_t *self,
 
 /* Functions for payload_buffer. */
 static void
-add_payload_to_buffer(signed_video_t *self,
-    uint8_t *payload_ptr,
-    uint8_t *payload_signature_ptr,
-    uint16_t last_two_bytes);
+add_payload_to_buffer(signed_video_t *self, uint8_t *payload_ptr, uint8_t *payload_signature_ptr);
 static svi_rc
 complete_sei_nalu_and_add_to_prepend(signed_video_t *self);
 
@@ -89,12 +86,10 @@ free_payload_buffer(uint8_t *payload_buffer[])
   }
 }
 
-/* Adds the |payload| to the next available slot in |payload_buffer|. */
+/* Adds the payload to the next available slot in |payload_buffer| and |last_two_bytes| to the
+ * next available slot in |last_two_bytes_buffer|. */
 static void
-add_payload_to_buffer(signed_video_t *self,
-    uint8_t *payload,
-    uint8_t *payload_signature_ptr,
-    uint16_t last_two_bytes)
+add_payload_to_buffer(signed_video_t *self, uint8_t *payload, uint8_t *payload_signature_ptr)
 {
   assert(self);
 
@@ -106,7 +101,7 @@ add_payload_to_buffer(signed_video_t *self,
 
   self->payload_buffer[self->payload_buffer_idx] = payload;
   self->payload_buffer[self->payload_buffer_idx + 1] = payload_signature_ptr;
-  self->last_two_bytes_buffer[self->payload_buffer_idx / 2] = last_two_bytes;
+  self->last_two_bytes_buffer[self->payload_buffer_idx / 2] = self->last_two_bytes;
   self->payload_buffer_idx += 2;
 }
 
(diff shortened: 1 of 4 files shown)


lib/src/signed_video_internal.h
```diff
@@ -72,6 +72,8 @@ typedef struct _h26x_nalu_list_t h26x_nalu_list_t;
 
 #define HASH_LIST_SIZE (HASH_DIGEST_SIZE * MAX_GOP_LENGTH)
 
+#define SIZE_OF_SIGNING_REQUEST_QUEUE 3
```
Caroline-Holmstrom, line 75: Remove

lib/src/signed_video_h26x_sign.c
```diff
@@ -119,8 +119,9 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
   // Transfer oldest pointer in |payload_buffer| to local |payload|
   uint8_t *payload = self->payload_buffer[0];
   uint8_t *payload_signature_ptr = self->payload_buffer[1];
-  self->payload_buffer[0] = NULL;  // Set to NULL since pointer has been transferred.
-  self->payload_buffer[1] = NULL;  // Set to NULL since pointer has been transferred.
+  for (uint8_t j = 0; j < buffer_end - 1; j++) {
+    self->payload_buffer[self->payload_buffer_idx] = self->payload_buffer[self->payload_buffer_idx + 2];
+  }
```
bjornvolcker, line 124:
```
for (int j = 0; j < buffer_end - 1; j++) {
  self->payload_buffer[j] = self->payload_buffer[j + 2];
  self->payload_buffer[j + 1] = self->payload_buffer[j + 3];
}
self->payload_buffer[buffer_end * 2] = NULL;
self->payload_buffer[buffer_end * 2 + 1] = NULL;
self->payload_buffer_idx--;
```

lib/src/signed_video_h26x_sign.c
```diff
@@ -119,8 +119,9 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
   // Transfer oldest pointer in |payload_buffer| to local |payload|
```
bjornvolcker, line 119: Since we touch `self->payload_buffer_idx` there should be a sanity check on it.
```
if (self->payload_buffer_idx < 1) return SVI_NOT_SUPPORTED;
```

bjornvolcker (changes requested):

lib/src/signed_video_h26x_sign.c
```diff
@@ -98,9 +98,9 @@ add_payload_to_buffer(signed_video_t *self, uint8_t *payload, uint8_t *payload_s
     return;
   }
 
-  self->payload_buffer[2 * self->payload_buffer_idx] = payload;
-  self->payload_buffer[2 * self->payload_buffer_idx + 1] = payload_signature_ptr;
-  self->payload_buffer_idx += 1;
+  self->payload_buffer[self->payload_buffer_idx] = payload;
+  self->payload_buffer[self->payload_buffer_idx + 1] = payload_signature_ptr;
+  self->payload_buffer_idx += 2;
```
bjornvolcker, line 103: Now when the behavior of `payload_buffer_idx` has changed you need to adjust the sanity check on line 95 above

lib/src/signed_video_h26x_sign.c
```diff
@@ -109,6 +109,7 @@ static svi_rc
 complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
 {
   assert(self);
+  if (self->payload_buffer_idx < 1) return SVI_NOT_SUPPORTED;
```
bjornvolcker, line 112: Note that also `payload_buffer_idx == 1` is invalid, hence this should be `payload_buffer_idx < 2`

lib/src/signed_video_h26x_sign.c
```diff
@@ -160,15 +158,12 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
...
   // failure.
-  if (buffer_end > 0) {
-    for (int i = 2; i < buffer_end; i++) {
-      self->payload_buffer[2 * (i - 1)] = self->payload_buffer[2 * i];
-      self->payload_buffer[2 * (i - 1) + 1] = self->payload_buffer[2 * i + 1];
-    }
-    self->payload_buffer[2 * (buffer_end - 1)] = NULL;
-    self->payload_buffer[2 * (buffer_end - 1) + 1] = NULL;
-    self->payload_buffer_idx -= 1;
+  for (uint8_t j = 0; j < buffer_end - 2; j++) {
```
bjornvolcker, line 161: `buffer_end` is an int, so use int also for `j`.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_sign.c
```diff
@@ -607,11 +616,12 @@ signed_video_set_end_of_stream(signed_video_t *self)
 
   uint8_t *payload = NULL;
   uint8_t *payload_signature_ptr = NULL;
+  uint16_t last_two_bytes = 0;
```
Caroline-Holmstrom, line 619: last_two_bytes should be initialized to LAST_TWO_BYTES_INIT_VALUE, not zero.

lib/src/signed_video_h26x_sign.c
```diff
@@ -158,11 +169,13 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
 done:
   // Done with the SEI payload. Move |payload_buffer|. This should be done even if we caught a
   // failure.
-  for (uint8_t j = 0; j < buffer_end - 2; j++) {
+  for (int j = 0; j < buffer_end - 2; j++) {
     self->payload_buffer[j] = self->payload_buffer[j + 2];
+    self->last_two_bytes[j] = self->last_two_bytes[j + 1];
```
bjornvolcker, line 174: You will read and write outside memory here.

lib/src/signed_video_h26x_sign.c
```diff
@@ -525,7 +538,7 @@ signed_video_add_nalu_for_signing(signed_video_t *self,
 
       SVI_THROW(generate_sei_nalu(self, &payload, &payload_signature_ptr));
       // Add |payload| to buffer. Will be picked up again when the signature has been generated.
-      add_payload_to_buffer(self, payload, payload_signature_ptr);
+      add_payload_to_buffer(self, payload, payload_signature_ptr, self->last_two_bytes[0]);
```
bjornvolcker, line 541: This becomes circular since inside the function you use the first instance to set where the index is

lib/src/signed_video_h26x_sign.c
```diff
@@ -287,7 +300,7 @@ generate_sei_nalu(signed_video_t *self, uint8_t **payload, uint8_t **payload_sig
     uint8_t *payload_ptr = *payload;
 
     // Start writing bytes.
-    uint16_t *last_two_bytes = &self->last_two_bytes;
+    uint16_t *last_two_bytes = &self->last_two_bytes[0];
```
bjornvolcker, line 303: You should use `last_two_bytes[self->payload_buffer_idx/2]`

bjornvolcker (changes requested):

lib/src/signed_video_h26x_sign.c
```diff
@@ -607,11 +615,12 @@ signed_video_set_end_of_stream(signed_video_t *self)
 
   uint8_t *payload = NULL;
   uint8_t *payload_signature_ptr = NULL;
+  uint16_t last_two_bytes = LAST_TWO_BYTES_INIT_VALUE;
```
bjornvolcker, line 618: Remove

bjornvolcker (changes requested):

lib/src/signed_video_h26x_sign.c
```diff
@@ -88,19 +91,23 @@ free_payload_buffer(uint8_t *payload_buffer[])
 
 /* Adds the |payload| to the next available slot in |payload_buffer|. */
```
bjornvolcker, line 92: Update description

lib/src/signed_video_internal.h
```diff
@@ -137,6 +137,7 @@ struct _signed_video_t {
   // of the payload and the second location to where the signature is about to be added.
   uint8_t *payload_buffer[MAX_NALUS_TO_PREPEND * 2];
   int payload_buffer_idx;  // Pointer to the current free location of the buffer.
+  uint16_t last_two_bytes_buffer[MAX_NALUS_TO_PREPEND];
```
bjornvolcker, line 140: Add description similar to

Buffer of last_two_bytes. Writing of the SEI is split in time and it is necessary to pick up this value when we continue writing.

bjornvolcker (dismissed):

lurvas777:
*Testing permissions*

lurvas777 (changes requested):
Since there's now another buffer that should be synchronized with the payload_buffer_idx I kinda dislike that the payload_buffer contains two items, whilst the last_two_bytes_buffer contains only one. Or that the last_two_bytes part isn't incorporated as a tuple of three items in the payload_buffer.
It makes them spread out and yet they use the same index but not in the same way since they're of different lengths. Keeping buffers of non-equal lengths is a potential for problems.

However, it looks to check out here so I'll approve except for one minor comment. But we should make it an issue to move there to separate buffers in the future.

lurvas777 (approved):

lib/src/signed_video_h26x_sign.c
```diff
@@ -294,8 +294,6 @@ generate_sei_nalu(signed_video_t *self, uint8_t **payload, uint8_t **payload_sig
     uint8_t *payload_ptr = *payload;
 
     // Start writing bytes.
-    // Reset last_two_bytes before writing bytes
-    self->last_two_bytes = LAST_TWO_BYTES_INIT_VALUE;
```
bjornvolcker: You still need to initialize this value. It is not taken from the buffer.
Either you initialize the temporary variable `self->last_two_bytes` like
```
self->last_two_bytes = LAST_TWO_BYTES_INIT_VALUE;
```
or you get it from the buffer (for which it is initialized already), like
```
self->last_two_bytes = self->last_two_bytes_buffer[self->payload_buffer_idx / 2];
```

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #93, opened 2022-07-18
Adds accumulated validation to authenticity report

### LINKED ISSUE #21: Add accumulated validation to signed_video_authenticity_t
A validation application needs to handle every `signed_video_authenticity_t` report that `signed_video_add_nalu_and_authenticate(...)` outputs. If the `signed_video_latest_validation_t` is not collected for a GOP it is lost. In many use cases an entire file is screened and a final result is desired.

By adding a `signed_video_accumulated_validation_t` to `signed_video_authenticity_t` it is easy for a user to validate a file in one chunk and then fetch a report using `signed_video_get_authenticity_report(...)`.

bjornvolcker:
### Describe your changes

Adds the new struct signed_video_accumulated_validation_t which keeps track of
the overall authenticity. The statistics is updated upon every validation and when
an authenticity report is fetched through either the API
signed_video_get_authenticity_report(...)
or through the call signed_video_add_nalu_and_authenticate(...).

Tests have been updated with these stats.

- Fixes #21


lib/src/signed_video_authenticity.c
```diff
@@ -260,9 +206,66 @@ authenticity_report_init(signed_video_authenticity_t *authenticity_report)
   authenticity_report->this_version = calloc(1, SV_VERSION_MAX_STRLEN);
 
   latest_validation_init(&authenticity_report->latest_validation);
-#ifdef ACCUMULATED_VALIDATION
   accumulated_validation_init(&authenticity_report->accumulated_validation);
-#endif
+}
+
+static void
+update_accumulated_validation(const signed_video_latest_validation_t *latest,
```
lurvas777, line 213: Const here is pointless. Parameters are passed by value, therefore you can still change what latest points to (which should violate the const imho), but not the pointer itself in this local context which also isn't meaningful for the latest pointer as it doesn't point to an array.

lib/src/signed_video_authenticity.c
```diff
@@ -289,25 +290,20 @@ signed_video_get_authenticity_report(signed_video_t *self)
...
     SVI_THROW_IF(!authenticity_report, SVI_MEMORY);
-    validation_str = h26x_nalu_list_get_validation_str(self->nalu_list);
-    SVI_THROW(
-        allocate_memory_and_copy_string(&self->latest_validation->validation_str, validation_str));
+    // Update |number_of_pending_nalus| since that may have changed since |latest_validation|.
+    signed_video_accumulated_validation_t *accumulated = self->accumulated_validation;
+    if (accumulated->authenticity == SV_AUTH_RESULT_NOT_SIGNED) {
+      accumulated->number_of_pending_nalus = accumulated->number_of_received_nalus;
+    } else {
+      accumulated->number_of_pending_nalus = self->nalu_list->num_items;
```
lurvas777, line 303: num_items comment in the struct says "The number of items linked together in the list", this doesn't accurately represent the number of pending items? Shouldn't the function h26x_nalu_list_num_pending_items be used? Haven't looked into the logic though..

lib/src/signed_video_h26x_nalu_list.c
```diff
@@ -520,17 +520,23 @@ h26x_nalu_list_get_validation_str(const h26x_nalu_list_t *list)
...
 h26x_nalu_list_clean_up(h26x_nalu_list_t *list)
 {
-  if (!list) return;
+  if (!list) return 0;
 
   // Remove validated items.
+  unsigned int removed_items = 0;
   h26x_nalu_list_item_t *item = list->first_item;
   while (item && item->validation_status != 'P' && !item->need_second_verification) {
+    if (item->validation_status != 'M') {
```
lurvas777, line 532: What about the statuses "U", "E" and "_", are these all treated as "validated", but might have failed validation?

lurvas777 (commented):
Haven't worked much with the validation, so I don't know all the logics and how these structs are operated on. But put some comments mostly to trigger some thought and hopefully some explanations :)

lib/src/signed_video_authenticity.c
```diff
@@ -29,57 +29,21 @@
 #include "signed_video_h26x_nalu_list.h"  // h26x_nalu_list_get_validation_str()
 
 // Adding accumulated authenticity results, valuable for screening a file, is work in progress.
```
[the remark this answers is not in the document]
bjornvolcker: Nice catch

lib/src/signed_video_authenticity.c
```diff
@@ -122,31 +86,25 @@ allocate_memory_and_copy_string(char **dst_str, const char *src_str)
...
-static svi_rc
-transfer_accumulated_validation(signed_video_accumulated_validation_t *dst,
-    const signed_video_accumulated_validation_t *src)
+svi_rc
+transfer_product_info(signed_video_product_info_t *dst, const signed_video_product_info_t *src)
 {
-  assert(dst && src);
+  // For simplicity we allow nullptrs for both |dst| and |src|. If so, we take no action and return
+  // SVI_OK.
+  if (!src || !dst) return SVI_OK;
```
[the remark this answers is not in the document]
bjornvolcker: The diff has messed up. This is actually old code. Back in the days when this was added, this function was (and may still be) a helper in other places and there was no guarantee that inputs where valid. To avoid breaking things, NULL ptr were allowed. We should revisit this and replace with an assert or follow your suggestion. But that should be in a separate PR.

lib/src/signed_video_authenticity.c
bjornvolcker: I do not fully agree. Here are my thoughts.
The pointer to `latest`is passed by value hence there is no point to mark that as const. The values in the memory `latest`points to is not and here we tell the compiler that these are not expected to change when calling this function. If so, something is wrong.
Yes, you can always cast const to non const if you want to change a value, but that means you have to do it explicitly. By adding const like above we can more easily catch bugs where values in `latest`change by accident.

lib/src/signed_video_authenticity.c
bjornvolcker: Very valid comment, and I should make this more clear in the code.
After discussions with a file player using this accumulated validation `number_of_pending_nalus` should mark the dangling end, that is, number of NALUs present after the last fully validated NALU.
This may include validated NALUs like a late arrived SEI and "ignoored" NALUs like AUD.

Maybe we should rename `number_of_pending_nalus` to `number_of_remaining_nalus` to avoid mismatch between the internal end public interpretation of 'pending'. What do you think?

lib/src/signed_video_h26x_nalu_list.c
bjornvolcker: Yes. A NALU has been validated when we have made a decision. That decision could be OK, Not OK, Unknown, Error, or Ignored.
Think of this like a user that blindly adds NALUs one-by-one and wants to know how many of these can I now pass forward for decoding/rendering (if they are authentic).

Caroline-Holmstrom (approved):
LGTM

lusikamalo (approved):
OK.

Caroline-Holmstrom (approved):

PR #103, opened 2022-10-10
Adds a queue of SEIs to sign

Caroline-Holmstrom:
Puts hashes and signing requests in buffers. A hash to sign can only be
processed if the previous request is completed and has been fetched. If
not, SE is blocked and GOPs will be unsigned. By putting signing
requests and hashes in buffers, delays can be handled without loss of
SEIs.

### Describe your changes

- Fixes #(issue)


lib/plugins/threaded-signing/plugin.c
```diff
@@ -111,16 +118,26 @@ local_signature_info_create(const signature_info_t *signature_info)
...
-  free(self->hash_to_sign);
-  self->hash_to_sign = NULL;
+
+  const int output_buffer_end = self->output_buffer_idx;
+  for (int i = 0; i < output_buffer_end; i++) {
+    free(self->output_buffer[i].signature);
+    self->output_buffer[i].signature = NULL;
+    self->output_buffer[i].size = 0;
+  }
+
```
lusikamalo, line 134: Also set "self->output_buffer_idx = 0" after the for loop?

lib/plugins/threaded-signing/plugin.c
```diff
@@ -111,16 +118,26 @@ local_signature_info_create(const signature_info_t *signature_info)
...
+ * |mutex|, |cond| and |is_running|. */
 static void
 sv_threaded_plugin_reset(sv_threaded_plugin_t *self)
 {
   local_signature_info_free(self->signature_info);
   self->signature_info = NULL;
-  free(self->hash_to_sign);
-  self->hash_to_sign = NULL;
+
+  const int output_buffer_end = self->output_buffer_idx;
```
lusikamalo, line 128: - How about to remove line 128.
- Use "self->output_buffer_idx" on row 129 instead of "output_buffer_end".
- And then finally add "self->output_buffer_idx = 0" after the for loop.

Please check if this suggestion works.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -111,16 +118,26 @@ local_signature_info_create(const signature_info_t *signature_info)
...
+  for (int i = 0; i < output_buffer_end; i++) {
+    free(self->output_buffer[i].signature);
+    self->output_buffer[i].signature = NULL;
+    self->output_buffer[i].size = 0;
+  }
+
+  const int input_buffer_end = self->input_buffer_idx;
```
lusikamalo, line 135: - How about to remove line 135.
- Use "self->input_buffer_idx" on row 136 instead of "input_buffer_end".
- And then finally add "self->input_buffer_idx = 0" after the for loop.

Please check if this suggestion works.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +154,51 @@ signing_worker_thread(void *user_data)
...
+      // Copy the hash to |signature_info| and start signing. In principle, it is now possible to
+      // prepare for a new hash.
       assert(self->hash_size == self->signature_info->hash_size);
-      memcpy(self->signature_info->hash, self->hash_to_sign, self->hash_size);
-      self->plugin_state = THREADED_SIGNING_SIGNS_HASH;
+      if (self->signature_info->hash) {
+      }
+      memcpy(self->signature_info->hash, self->input_buffer[0], self->hash_size);
+      if (self->signature_info->hash) {
+      }
```
lusikamalo, line 168: Looks strange here. No code inside if statement.

lib/plugins/threaded-signing/plugin.c
Caroline-Holmstrom: I think the code is easier to follow if we use output_buffer_end, but I'll change it if more people want me to.

lib/plugins/threaded-signing/plugin.c
Caroline-Holmstrom: See comment above.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -30,20 +30,25 @@
...
+typedef struct _threaded_signature_info_t threaded_signature_info_t;
+
 typedef enum {
-  THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN,
-  THREADED_SIGNING_HAS_HASH_TO_SIGN,
-  THREADED_SIGNING_SIGNS_HASH,
-  THREADED_SIGNING_HAS_SIGNATURE,
-  THREADED_SIGNING_ERROR,
+  THREADED_SIGNING_ERROR = 1,
 } threaded_signing_plugin_state;
```
bjornvolcker, line 43: I think you should try to get rid of this enum. It has no meaning now when there is only one value.

lib/plugins/threaded-signing/plugin.c
bjornvolcker: The name 'output_buffer_end' is actually wrong. It points to the first available element in the buffer and not the last one.
I think you should consider freeing all elements in the buffer instead of only the ones up to the "end". This will also make it possible for you to reuse allocated hashes. All these calloc and free are inefficient use of memory.
The input hashes are always the same length and the signatures are always at most 'max_signature_size'

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +281,34 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
+        self->output_buffer[j].size = self->output_buffer[j + 1].size;
+      }
+      self->output_buffer[output_buffer_end - 1].signature = NULL;
+      self->output_buffer[output_buffer_end - 1].size = 0;
+      self->output_buffer_idx -= 1;
     }
     // Change state and mark as copied.
-    self->plugin_state = THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN;
     has_copied_signature = true;
   } else if (self->plugin_state == THREADED_SIGNING_ERROR) {
```
bjornvolcker, line 306: This is the only place where 'plugin_state' is used. Right now it does not work for mainly two reasons. 1) it may be set for a different buffer element and, 2) you still copy to the output buffer even if the signing fails.

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -30,32 +30,36 @@
...
 #include "includes/signed_video_openssl.h"
 
-typedef enum {
-  THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN,
-  THREADED_SIGNING_HAS_HASH_TO_SIGN,
-  THREADED_SIGNING_SIGNS_HASH,
-  THREADED_SIGNING_HAS_SIGNATURE,
-  THREADED_SIGNING_ERROR,
-} threaded_signing_plugin_state;
+#define MAX_HASHES 60
```
bjornvolcker, line 38: How about renaming to `MAX_BUFFER_LENGTH` since it is used both for input (where it indeed is hashes) and output (where it includes signature and more)

lib/plugins/threaded-signing/plugin.c
```diff
@@ -30,32 +30,36 @@
...
+typedef struct _threaded_signature_info_t threaded_signature_info_t;
```
bjornvolcker, line 39: This is a local variable. No need to mark it as "threaded". How about renaming it so something like `output_data_t` or similar

lib/plugins/threaded-signing/plugin.c
```diff
@@ -30,32 +30,36 @@
...
 
-/* Threaded plugin handle maintaining the thread and locks. Further, stores the hash to sign and the
- * written signature (part of |signature_info|): */
+struct _threaded_signature_info_t {
+  uint8_t *signature;
+  size_t size;
+  bool threaded_signing_error;
```
bjornvolcker, line 44: Remove "threaded_" from this member

lib/plugins/threaded-signing/plugin.c
```diff
@@ -111,18 +115,28 @@ local_signature_info_create(const signature_info_t *signature_info)
...
 {
   local_signature_info_free(self->signature_info);
   self->signature_info = NULL;
-  free(self->hash_to_sign);
-  self->hash_to_sign = NULL;
+
+  for (int i = 0; i < MAX_HASHES; i++) {
+    free(self->output_buffer[i].signature);
+    self->output_buffer[i].signature = NULL;
+    self->output_buffer[i].size = 0;
```
bjornvolcker, line 128: To be fully correct, and minimize potential bugs, you should probably also set `threaded_signing_error = false;`

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +151,57 @@ signing_worker_thread(void *user_data)
...
+      } else {
+        sleep(5);
+      }
       g_mutex_lock(&self->mutex);
-      // When successfully done with signing, move |plugin_state| to THREADED_SIGNING_HAS_SIGNATURE,
-      // otherwise move to THREADED_SIGNING_ERROR to report the error when getting the signature.
-      if (status == SV_OK) {
-        self->plugin_state = THREADED_SIGNING_HAS_SIGNATURE;
+
+      self->output_buffer[self->output_buffer_idx].threaded_signing_error = false;
```
bjornvolcker, line 185:
```
self->output_buffer[self->output_buffer_idx].threaded_signing_error = (status != SV_OK);
```
and remove the if below

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +151,57 @@ signing_worker_thread(void *user_data)
...
       }
+
+      // Copy the |signature| to the output buffer
+      memcpy(self->output_buffer[self->output_buffer_idx].signature, self->signature_info->signature,
+          self->output_buffer[self->output_buffer_idx].size);
+      self->output_buffer_idx++;
+      }
+    } else {
+    // Wait for a signal, triggered when it is time to sign a hash.
+    g_cond_wait(&self->cond, &self->mutex);
```
bjornvolcker, line 204: Wrong indentation

bjornvolcker (changes requested):
Some more comments

lib/plugins/threaded-signing/plugin.c
```diff
@@ -172,13 +214,9 @@ signing_worker_thread(void *user_data)
...
  *
- * The |hash| is copied to |hash_to_sign| and the |plugin_state| is moved to
- * THREADED_SIGNING_HAS_HASH_TO_SIGN. If this is the first time of signing, memory for
- * |self->signature_info| and |hash_to_sign| is allocated and the |private_key| is copied from
- * |signature_info|.
- *
- * Further, if there is no worker thread running or if the latest signing has not yet finished an
- * SV_NOT_SUPPORTED is returned. */
+ * The |hash| is copied to |input_buffer|. If this is the first time of signing, memory for
+ * |output_buffer| and |input_buffer| is allocated and the |private_key| is copied from
```
bjornvolcker, line 218: You only allocate memory for the `input_buffer` here

lib/plugins/threaded-signing/plugin.c
```diff
@@ -188,25 +226,17 @@ threaded_openssl_sign_hash(sv_threaded_plugin_t *self, const signature_info_t *s
...
   if (!self->signature_info) {
     self->signature_info = local_signature_info_create(signature_info);
     if (!self->signature_info) goto catch_error;
   }
 
-  // Allocate memory for the |hash_to_sign| if necessary.
-  if (!self->hash_to_sign) {
-    self->hash_to_sign = calloc(1, signature_info->hash_size);
-    if (!self->hash_to_sign) goto catch_error;
+  // Allocate memory for the |input_buffer| if necessary.
```
bjornvolcker, line 236: To avoid all these malloc + free after consumption there are (at least) two other approaches to solve it.

1. Still allocate memory for a hash when there is no available slot, but instead of freeing it on line 164 you can move the pointer to the first non-NULL element in `input_buffer`
2. Use one big memory to stack all the hashes and use `memmove()` on the buffer after fetching a hash for signing

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +279,34 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
       *written_signature_size = 0;
     } else {
-      memcpy(signature, self->signature_info->signature, self->signature_info->signature_size);
-      *written_signature_size = self->signature_info->signature_size;
+      // Get the oldest signature
+      memcpy(signature, self->output_buffer[0].signature, self->output_buffer[0].size);
+      *written_signature_size = self->output_buffer[0].size;
+      // Free |signature| and move buffer
+      const int output_buffer_end = self->output_buffer_idx;
+      free(self->output_buffer[0].signature);
```
bjornvolcker, line 292: Instead of freeing the memory you should reuse the same slot if you allocate `max_signature_size`.

bjornvolcker (changes requested):

tests/check/check_h26xsigned_sign.c
```diff
@@ -714,7 +714,7 @@ signed_video_suite(void)
 #endif
   // tcase_add_loop_test(tc, correct_nalu_sequence_with_eos, s, e);
   // tcase_add_loop_test(tc, correct_multislice_sequence_with_eos, s, e);
-  tcase_add_loop_test(tc, correct_nalu_sequence_without_eos, s, e);
+  tcase_add_loop_test(tc, correct_nalu_sequence_without_eos, s, 1);
```
bjornvolcker, line 717: You need to temporarily change the meson as well. Otherwise the threaded plugin will not be executed in the tests and memory leaks cannot be detected.

bjornvolcker (changes requested):

Caroline-Holmstrom (commented):
Something is wrong with this commit. Lots of things are not updated correctly.

tests/meson.build
```diff
@@ -2,6 +2,7 @@ if check_dep.found()
   if (signing_plugin == 'unthreaded')
     subdir('check')
   else
+    subdir('check')
```
bjornvolcker, line 5: You need to change the default setting to `threaded` in `mseon_options.txt` as well to make sure we run the tests (temporarily) with the code in the plugin.

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -163,10 +163,11 @@ signing_worker_thread(void *user_data)
       assert(self->signature_info->hash);
       memcpy(self->signature_info->hash, self->input_buffer[0], self->hash_size);
 
-      free(self->input_buffer[0]);
       self->input_buffer[0]= NULL;
```
bjornvolcker, line 166: This will leak memory, since you do not free. You should reuse the memory for later by moving the pointer to the first available slot in input_buffer.

bjornvolcker (changes requested):

tests/check/check_h26xsigned_auth.c
```diff
@@ -1957,7 +1957,7 @@ signed_video_suite(void)
   //   for (int _i = s; _i < e; _i++) {}
 
   int s = 0;
-  int e = NUM_SETTINGS;
+  int e = 0;
```
lusikamalo, line 1960: Change back to NUM_SETTINGS.

tests/check/check_h26xsigned_sign.c
```diff
@@ -704,7 +704,7 @@ signed_video_suite(void)
   //   for (int _i = s; _i < e; _i++) {}
 
   int s = 0;
-  int e = NUM_SETTINGS;
+  int e = 0;
```
lusikamalo, line 707: Change back to NUM_SETTINGS.

bjornvolcker:
> 

That should not be necessary now when we've temporarily changed the default value.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -123,17 +123,22 @@ sv_threaded_plugin_reset(sv_threaded_plugin_t *self)
   local_signature_info_free(self->signature_info);
   self->signature_info = NULL;
 
-  for (int i = 0; i < MAX_BUFFER_LENGTH; i++) {
-    free(self->output_buffer[i].signature);
-    self->output_buffer[i].signature = NULL;
-    self->output_buffer[i].size = 0;
-    self->output_buffer[i].signing_error = false;
+  int j = 0;
+  while (self->output_buffer[j].signature != NULL) {
```
bjornvolcker, line 127: I think you should keep the for loop and free _all_ elements. If there by accident has become a gap in the buffer, that is, on NULL pointer has sneaked in in the middle of true pointers, you will not free all of the memory.
There is no need to minimize operations in the reset call, since it is only done when something has gone wrong or when tearing down the session.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -155,17 +161,19 @@ signing_worker_thread(void *user_data)
...
       int j = 0;
-      while (self->input_buffer[j + 1] != NULL) {
-        self->input_buffer[j]= self->input_buffer[j + 1];
+      while (self->input_buffer[j] != NULL) {
+        self->input_buffer[j] = self->input_buffer[j + 1];
+        if (self->input_buffer[j + 1] != NULL) {
+        }
         j++;
       }
       self->input_buffer[j + 1] = tmp;
```
bjornvolcker, line 179: This will create a gap in the buffer.
Consider the buffer
```
[0xabc, 0x0, 0x0, 0x0]
```
then ```tmp = 0xabc``` and the after the while loop you have
```
[0x0, 0x0, 0x0, 0x0]
j = 1
```
so after ```self->input_buffer[j+1] = self->input_buffer[2] = tmp = 0xabc```
hence
```
[0x0, 0x0, 0xabc, 0x0]
```

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -111,23 +116,34 @@ local_signature_info_create(const signature_info_t *signature_info)
...
-  self->plugin_state = THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN;
 }
 
+static int a = 0;
 /* The worker thread waits for a condition signal, triggered when there is a hash to sign. */
 static void *
 signing_worker_thread(void *user_data)
 {
   sv_threaded_plugin_t *self = (sv_threaded_plugin_t *)user_data;
+  self->output_buffer_idx = 0;
```
bjornvolcker, line 146: Remove. This is not needed. And it is not done under a lock either.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +153,58 @@ signing_worker_thread(void *user_data)
...
   while (self->is_running) {
-    // Wait for a signal, triggered when it is time to sign a hash.
-    g_cond_wait(&self->cond, &self->mutex);
-
-    if (self->plugin_state == THREADED_SIGNING_HAS_HASH_TO_SIGN) {
-      // Copy the |hash_to_sign| to |signature_info| and start signing. In principle, it is now
-      // possible to prepare for a new |hash_to_sign|.
+    if (self->input_buffer_idx > 0) {
+      // Get the oldest hash from the input buffer
+      assert(self->input_buffer_idx <= MAX_BUFFER_LENGTH);
```
bjornvolcker, line 158: I think you can remove this assert

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +153,58 @@ signing_worker_thread(void *user_data)
...
       assert(self->hash_size == self->signature_info->hash_size);
-      memcpy(self->signature_info->hash, self->hash_to_sign, self->hash_size);
-      self->plugin_state = THREADED_SIGNING_SIGNS_HASH;
+      assert(self->signature_info->hash);
+      memcpy(self->signature_info->hash, self->input_buffer[0], self->hash_size);
+
+      uint8_t *tmp = self->input_buffer[0];
+      int j = 0;
+      while (self->input_buffer[j] != NULL) {
+        self->input_buffer[j] = self->input_buffer[j + 1];
```
bjornvolcker, line 168: You will read (and write) outside memory if all ```MAX_BUFFER_LENGTH``` elements points to allocated memories.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +153,58 @@ signing_worker_thread(void *user_data)
...
+      if (a > 3) {
+        sleep(0);
+      } else {
+        sleep(5);
+      }
       g_mutex_lock(&self->mutex);
-      // When successfully done with signing, move |plugin_state| to THREADED_SIGNING_HAS_SIGNATURE,
-      // otherwise move to THREADED_SIGNING_ERROR to report the error when getting the signature.
+
+      self->output_buffer[self->output_buffer_idx].signing_error = false;
```
bjornvolcker, line 187: Remove this line. It is set on line 190

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +153,58 @@ signing_worker_thread(void *user_data)
...
+
       if (status == SV_OK) {
-        self->plugin_state = THREADED_SIGNING_HAS_SIGNATURE;
-      } else {
-        self->plugin_state = THREADED_SIGNING_ERROR;
+        // Allocate memory for the |signature| if necessary.
+        if (!self->output_buffer[self->output_buffer_idx].signature) {
+          self->output_buffer[self->output_buffer_idx].signature =
+              calloc(1, self->signature_info->max_signature_size);
+          self->output_buffer[self->output_buffer_idx].size = self->signature_info->signature_size;
```
bjornvolcker, line 197: Do not set the size until after you have successfully copied the signature.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +153,58 @@ signing_worker_thread(void *user_data)
...
+      self->output_buffer[self->output_buffer_idx].signing_error = (status != SV_OK);
+
       if (status == SV_OK) {
-        self->plugin_state = THREADED_SIGNING_HAS_SIGNATURE;
-      } else {
-        self->plugin_state = THREADED_SIGNING_ERROR;
+        // Allocate memory for the |signature| if necessary.
+        if (!self->output_buffer[self->output_buffer_idx].signature) {
+          self->output_buffer[self->output_buffer_idx].signature =
+              calloc(1, self->signature_info->max_signature_size);
```
bjornvolcker, line 196: What happens if ```calloc()``` fails and returns NULL?

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +281,34 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
+      self->output_buffer[i - 1] = tmp;
+      if (self->output_buffer_idx > 0) {
+        self->output_buffer_idx -= 1;
+      }
     }
     // Change state and mark as copied.
-    self->plugin_state = THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN;
     has_copied_signature = true;
-  } else if (self->plugin_state == THREADED_SIGNING_ERROR) {
+  } else if (self->output_buffer[0].signing_error) {
```
bjornvolcker, line 306: I think you should check ```signing_error``` first.
You can only end up here if ```output_buffer_idx = 0```, then it does not make sense to check the content in position 0 right?

Also, even if there is an error you need to move the buffer.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +281,34 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
   SignedVideoReturnCode status = SV_OK;
 
   g_mutex_lock(&self->mutex);
-  if (self->plugin_state == THREADED_SIGNING_HAS_SIGNATURE && self->signature_info) {
-    if (self->signature_info->signature_size > max_signature_size) {
+  if (self->output_buffer_idx > 0) {
+    if (self->output_buffer[0].size > max_signature_size) {
       // If there is no room to copy the signature, report zero size.
       *written_signature_size = 0;
```
bjornvolcker, line 287: If you end up here you still need to move the buffer.

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -21,41 +21,46 @@
 
 /**
  * This signing plugin sets up a worker thread and calls openssl_sign_hash(), from the worker
- * thread, when there is a new hash to sign. The plugin can only handle one signature at a time,
- * that is, there is no queue of hashes to sign. If the latest signature is not completed by the
- * time of a new request, that new hash is not signed.
+ * thread, when there is a new hash to sign. The plugin can only handle one signature at a time.
```
bjornvolcker, line 24: "The plugin can only handle one signature at a time"
This is not really true anymore, right?

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +152,61 @@ signing_worker_thread(void *user_data)
...
   while (self->is_running) {
-    // Wait for a signal, triggered when it is time to sign a hash.
-    g_cond_wait(&self->cond, &self->mutex);
-
-    if (self->plugin_state == THREADED_SIGNING_HAS_HASH_TO_SIGN) {
-      // Copy the |hash_to_sign| to |signature_info| and start signing. In principle, it is now
-      // possible to prepare for a new |hash_to_sign|.
+    if (self->input_buffer_idx > 0) {
+      // Get the oldest hash from the input buffer
+      // Copy the hash to |signature_info| and start signing. In principle, it is now possible to
```
bjornvolcker, line 157: Remove "In principle, it is now possible to
      // prepare for a new hash."

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +152,61 @@ signing_worker_thread(void *user_data)
...
+      // prepare for a new hash.
       assert(self->hash_size == self->signature_info->hash_size);
-      memcpy(self->signature_info->hash, self->hash_to_sign, self->hash_size);
-      self->plugin_state = THREADED_SIGNING_SIGNS_HASH;
+      assert(self->signature_info->hash);
+      memcpy(self->signature_info->hash, self->input_buffer[0], self->hash_size);
+
+      uint8_t *tmp = self->input_buffer[0];
+      int j = 0;
+      while (self->input_buffer[j] != NULL && j < MAX_BUFFER_LENGTH) {
```
bjornvolcker, line 165: This will still make you read outside memory if the entire buffer is filled with pointers to allocated memories.
You need to stop one element earlier.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +152,61 @@ signing_worker_thread(void *user_data)
...
+        // Allocate memory for the |signature| if necessary.
+        if (!self->output_buffer[self->output_buffer_idx].signature) {
+          self->output_buffer[self->output_buffer_idx].signature =
+              calloc(1, self->signature_info->max_signature_size);
+        }
+        if (!self->output_buffer[self->output_buffer_idx].signature) {
+          // Failed in memory allocation. Free all memory and set status to SV_MEMORY.
+          sv_threaded_plugin_reset(self);
+          status = SV_MEMORY;
+        }
```
bjornvolcker, line 199: The first thing would be to put this inside the if above.
Secondly, this will not fly since you move on after this if. Sp on line 202 below you'll get a crash since you copy to a NULL pointer.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +152,61 @@ signing_worker_thread(void *user_data)
...
+
+        // Copy the |signature| to the output buffer
+        memcpy(self->output_buffer[self->output_buffer_idx].signature,
+            self->signature_info->signature, self->output_buffer[self->output_buffer_idx].size);
```
bjornvolcker, line 203: You cannot use ```self->output_buffer[self->output_buffer_idx].size``` since it has not been set yet. Use ```self->signature_info->signature_size```

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +283,32 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
   SignedVideoReturnCode status = SV_OK;
 
   g_mutex_lock(&self->mutex);
-  if (self->plugin_state == THREADED_SIGNING_HAS_SIGNATURE && self->signature_info) {
-    if (self->signature_info->signature_size > max_signature_size) {
+  if (self->output_buffer[0].signing_error) {
```
bjornvolcker, line 286: Won't this crash for you? Element 0 is a NULL pointer until the first signature is ready.
You need to check if there are any signatures in the buffer first.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +283,32 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
-    // There are unsigned hashes in the pipe. Report them with zero size, since no signature exists.
-    *written_signature_size = 0;
-    self->nbr_of_unsigned_hashes--;
-    has_copied_signature = true;
+  }
+  if (self->output_buffer_idx > 0) {
+    // Move buffer
+    output_data_t tmp = self->output_buffer[0];
+    int i = 0;
+    while (self->output_buffer[i].signature != NULL) {
```
bjornvolcker, line 306: Like for the ```input_buffer```, you will read and write outside memory if the buffer is full of allocated pointers.

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -154,19 +153,22 @@ signing_worker_thread(void *user_data)
...
 
       uint8_t *tmp = self->input_buffer[0];
       int j = 0;
-      while (self->input_buffer[j] != NULL && j < MAX_BUFFER_LENGTH) {
+      while (self->input_buffer[j] != NULL && j < MAX_BUFFER_LENGTH - 1) {
         self->input_buffer[j] = self->input_buffer[j + 1];
         j++;
       }
-      self->input_buffer[j - 1] = tmp;
+      if (j < MAX_BUFFER_LENGTH - 1) {
```
bjornvolcker, line 167: You can get rid of this if-else and replace it with ```self->input_buffer[j] = tmp;``` if you in the wile loop use ```self->input_buffer[j+1] != NULL```

lib/plugins/threaded-signing/plugin.c
```diff
@@ -298,16 +302,18 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
     // Move buffer
     output_data_t tmp = self->output_buffer[0];
     int i = 0;
-    while (self->output_buffer[i].signature != NULL) {
+    while (self->output_buffer[i].signature != NULL && i < MAX_BUFFER_LENGTH - 1) {
       self->output_buffer[i] = self->output_buffer[i + 1];
       i++;
     }
-    self->output_buffer[i - 1] = tmp;
+    if (i < MAX_BUFFER_LENGTH - 1) {
```
bjornvolcker, line 312: See my comment on the ```input_buffer``` above

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +151,66 @@ signing_worker_thread(void *user_data)
...
+        }
+
+        // Copy the |signature| to the output buffer
+        memcpy(self->output_buffer[self->output_buffer_idx].signature,
+            self->signature_info->signature, self->signature_info->signature_size);
+        self->output_buffer[self->output_buffer_idx].size = self->signature_info->signature_size;
+        self->output_buffer_idx++;
       }
+    catch_error:
+      if (status == SV_UNKNOWN_FAILURE) {
```
bjornvolcker, line 206: This will not work.
If you fail on memory allocation on line 196, ```status == SV_OK```, because otherwise you would not have entered the if-statement.

The other thing is the error handling itself. Is this really what we want? You reset all buffers and continue signing. If memory allocation fails we are in big trouble and should stop the thread.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +287,34 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
   SignedVideoReturnCode status = SV_OK;
 
   g_mutex_lock(&self->mutex);
-  if (self->plugin_state == THREADED_SIGNING_HAS_SIGNATURE && self->signature_info) {
-    if (self->signature_info->signature_size > max_signature_size) {
+  if (self->output_buffer_idx > 0) {
```
bjornvolcker, line 290: I would recommend changing to check if ```output_buffer_idx == 0``` and if so jump straight to line 319 (unlock).
That will avoid nested if:s which makes the code harder to read.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +287,34 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
+    if (self->output_buffer[0].signing_error) {
+      *written_signature_size = 0;
+      // Propagate SV_EXTERNAL_ERROR when signing failed.
+      status = SV_EXTERNAL_ERROR;
+    } else if (self->output_buffer[0].size > max_signature_size) {
```
bjornvolcker, line 295: You should set ```status = SV_INVALID_PARAMETER;```

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,30 +287,34 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
   SignedVideoReturnCode status = SV_OK;
 
   g_mutex_lock(&self->mutex);
-  if (self->plugin_state == THREADED_SIGNING_HAS_SIGNATURE && self->signature_info) {
-    if (self->signature_info->signature_size > max_signature_size) {
+  if (self->output_buffer_idx > 0) {
+    if (self->output_buffer[0].signing_error) {
+      *written_signature_size = 0;
```
bjornvolcker, line 292: You can see this as an initialization and do it before the if.

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -21,41 +21,45 @@
...
+// seconds if the GOP length is 1 second
+#define MAX_BUFFER_LENGTH 60
 
-/* Threaded plugin handle maintaining the thread and locks. Further, stores the hash to sign and the
- * written signature (part of |signature_info|): */
+typedef struct _output_data_t {
+  uint8_t *signature;
+  size_t size;
+  bool signing_error;
+} output_data_t;
```
lurvas777, line 45: output_data_t is very generic. I'd change it to something like signature_output_data_t (goes for the name starting with underscore as well)

lib/plugins/threaded-signing/plugin.c
```diff
@@ -21,41 +21,45 @@
...
- * time of a new request, that new hash is not signed.
+ * thread, when there is a new hash to sign. To handle several signatures at the same time, the
+ * plugin has two buffers. One for incomming hashes and another for outgoing signatures.
  */
 
 #include <assert.h>
 #include <glib.h>
 #include <stdlib.h>  // calloc, malloc, free
 #include <string.h>  // memcpy
+#include <unistd.h>
```
bjornvolcker, line 32: What is this include used for?

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +151,59 @@ signing_worker_thread(void *user_data)
...
+    if (self->input_buffer_idx > 0) {
+      // Get the oldest hash from the input buffer
+      // Copy the hash to |signature_info| and start signing.
       assert(self->hash_size == self->signature_info->hash_size);
-      memcpy(self->signature_info->hash, self->hash_to_sign, self->hash_size);
-      self->plugin_state = THREADED_SIGNING_SIGNS_HASH;
+      assert(self->signature_info->hash);
+      memcpy(self->signature_info->hash, self->input_buffer[0], self->hash_size);
+
+      uint8_t *tmp = self->input_buffer[0];
```
bjornvolcker, line 161: Add a comment on what and why you do this while thing.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,29 +151,59 @@ signing_worker_thread(void *user_data)
...
+
+        // Copy the |signature| to the output buffer
+        memcpy(self->output_buffer[self->output_buffer_idx].signature,
+            self->signature_info->signature, self->signature_info->signature_size);
+        self->output_buffer[self->output_buffer_idx].size = self->signature_info->signature_size;
+        self->output_buffer_idx++;
       }
+    catch_error:
+      // Failed in memory allocation. Set status to SV_MEMORY.
+      status = SV_MEMORY;
```
bjornvolcker, line 203: setting status will not do anything. In fact you will always set status to SV_MEMORY since we always end up here. Even if there is no failure.
I think you should consider stopping the thread if memory allocation fails instead of doing nothing.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -172,13 +216,8 @@ signing_worker_thread(void *user_data)
...
  *
- * The |hash| is copied to |hash_to_sign| and the |plugin_state| is moved to
- * THREADED_SIGNING_HAS_HASH_TO_SIGN. If this is the first time of signing, memory for
- * |self->signature_info| and |hash_to_sign| is allocated and the |private_key| is copied from
- * |signature_info|.
- *
- * Further, if there is no worker thread running or if the latest signing has not yet finished an
- * SV_NOT_SUPPORTED is returned. */
+ * The |hash| is copied to |input_buffer|. If this is the first time of signing, memory for
+ * |input_buffer| is allocated and the |private_key| is copied from |signature_info|. */
```
bjornvolcker, line 220: I guess you still mean ```self->signature_info``` and not ```input_buffer```. The ```input_buffer``` itself was allocated when ```self``` was created. What should be added to the description though is something like "If memory for the hash has not been allocated it will be allocated."
You also need to describe what happens if we run out of hash slots in ```input_buffer```.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -188,25 +227,16 @@ threaded_openssl_sign_hash(sv_threaded_plugin_t *self, const signature_info_t *s
...
   if (!self->signature_info) {
     self->signature_info = local_signature_info_create(signature_info);
     if (!self->signature_info) goto catch_error;
   }
 
-  // Allocate memory for the |hash_to_sign| if necessary.
-  if (!self->hash_to_sign) {
-    self->hash_to_sign = calloc(1, signature_info->hash_size);
-    if (!self->hash_to_sign) goto catch_error;
+  if (!self->input_buffer[self->input_buffer_idx]) {
```
bjornvolcker, line 237: What if ```self->input_buffer_idx >= MAX_BUFFER_LENGTH```?

lib/plugins/threaded-signing/plugin.c
```diff
@@ -228,17 +259,12 @@ threaded_openssl_sign_hash(sv_threaded_plugin_t *self, const signature_info_t *s
...
   return status;
 }
 
-/* If |plugin_state| = THREADED_SIGNING_HAS_SIGNATURE a new |signature| has been written to
- * |signature_info|. The new |signature| is then copied to the output and the |plugin_state|
- * is set to THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN.
- *
- * Returns true if a new |signature| has been copied to output, otherwise false.
+/* Returns true if a new |signature| has been copied to output, otherwise false.
  * If the hash could not be signed due to a blocked openssl_sign_hash(), |signature_size| is set to
```
bjornvolcker, line 268: Is this description really correct now?

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,31 +280,33 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
-    // Change state and mark as copied.
-    self->plugin_state = THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN;
-    has_copied_signature = true;
-  } else if (self->plugin_state == THREADED_SIGNING_ERROR) {
-    *written_signature_size = 0;
-    // Change state and mark as copied.
-    self->plugin_state = THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN;
-    has_copied_signature = true;
+  if (self->output_buffer_idx == 0) goto done;
+  *written_signature_size = 0;
```
bjornvolcker, line 284: Add empty line before this one to avoid associating it with the if-statement above

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,31 +280,33 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
+  if (self->output_buffer[0].signing_error) {
     // Propagate SV_EXTERNAL_ERROR when signing failed.
     status = SV_EXTERNAL_ERROR;
-  } else if (self->plugin_state == THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN &&
-      self->nbr_of_unsigned_hashes > 0) {
-    // There are unsigned hashes in the pipe. Report them with zero size, since no signature exists.
+  } else if (self->output_buffer[0].size > max_signature_size) {
+    // If there is no room to copy the signature, report zero size and set status to invalid
+    // parameter.
     *written_signature_size = 0;
```
bjornvolcker, line 291: Remove this line. It has already been set to 0.

bjornvolcker (changes requested):

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,48 +149,86 @@ signing_worker_thread(void *user_data)
...
+      // If not successfully done with signing, set |signing_error| to true to
+      // report the error when getting the signature.
+      self->output_buffer[self->output_buffer_idx].signing_error = (status != SV_OK);
+
       if (status == SV_OK) {
-        self->plugin_state = THREADED_SIGNING_HAS_SIGNATURE;
-      } else {
-        self->plugin_state = THREADED_SIGNING_ERROR;
+        // Allocate memory for the |signature| if necessary.
+        if (!self->output_buffer[self->output_buffer_idx].signature) {
```
bjornvolcker, line 189: Here you may read outside memory, which probably will cause a crash or unknown behavior later on. You need a strategy if the ```output_buffer``` is full.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,48 +149,86 @@ signing_worker_thread(void *user_data)
...
+          self->output_buffer[self->output_buffer_idx].signature =
+              calloc(1, self->signature_info->max_signature_size);
+          // Failed in memory allocation. Stop the thread, free all memory and set status to
+          // SV_MEMORY.
+          if (!self->output_buffer[self->output_buffer_idx].signature) {
+            self->is_running = false;
+            status = SV_MEMORY;
```
bjornvolcker, line 196: Where will ```status``` be used?
There is no point in assigning it a value that will not be used.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,48 +149,86 @@ signing_worker_thread(void *user_data)
...
+            sv_threaded_plugin_reset(self);
+            g_mutex_unlock(&self->mutex);
+            goto catch_error;
```
bjornvolcker, line 199: The only difference jumping to ```catch_error``` compared to ```done``` is the unlock operation which you do here as well. That means you can replace these two lines with ```goto done;```.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,48 +149,86 @@ signing_worker_thread(void *user_data)
...
+      // Wait for a signal, triggered when it is time to sign a hash.
+      g_cond_wait(&self->cond, &self->mutex);
     }
   };
 
 done:
+
   g_mutex_unlock(&self->mutex);
 
+catch_error:
```
bjornvolcker, line 219: Not needed; See comment above. Remove.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -215,32 +262,27 @@ threaded_openssl_sign_hash(sv_threaded_plugin_t *self, const signature_info_t *s
...
+  self->input_buffer_idx++;
 
   status = SV_OK;
-  self->plugin_state = THREADED_SIGNING_HAS_HASH_TO_SIGN;
 
 catch_error:
-  if (status == SV_UNKNOWN_FAILURE) {
+  if (status != SV_OK) {
     // Failed in memory allocation. Free all memory and set status to SV_MEMORY.
+    if (status == SV_UNKNOWN_FAILURE) status = SV_MEMORY;
```
bjornvolcker, line 274: I think you should remove this line and instead make sure ```status``` is set right before each ```goto catch_error;```

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -193,10 +200,8 @@ signing_worker_thread(void *user_data)
           // SV_MEMORY.
```
bjornvolcker, line 200: The comment states that you set status to SV_MEMORY, but that is not true. Rewrite.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -184,6 +184,13 @@ signing_worker_thread(void *user_data)
       // report the error when getting the signature.
       self->output_buffer[self->output_buffer_idx].signing_error = (status != SV_OK);
 
+      // |output_buffer| is full. Buffers this long are not supported.
+      if (self->output_buffer_idx >= MAX_BUFFER_LENGTH) {
+        self->is_running = false;
+        sv_threaded_plugin_reset(self);
+        goto done;
```
bjornvolcker, line 191: One could argue that only a failing memory allocation should stop the thread. What do you propose?

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
Caroline-Holmstrom: I propose that we stop the thread if the buffer is full. Buffer size is large enough to handle delays, hence something bad has happened if the buffer is full and we should return some kind of error code, stop the thread and free memory.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -21,41 +21,45 @@
 
 /**
  * This signing plugin sets up a worker thread and calls openssl_sign_hash(), from the worker
- * thread, when there is a new hash to sign. The plugin can only handle one signature at a time,
- * that is, there is no queue of hashes to sign. If the latest signature is not completed by the
- * time of a new request, that new hash is not signed.
+ * thread, when there is a new hash to sign. To handle several signatures at the same time, the
+ * plugin has two buffers. One for incomming hashes and another for outgoing signatures.
```
bjornvolcker, line 25: I think it would be good to add, in the description, under what conditions the thread is stopped.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -111,18 +115,26 @@ local_signature_info_create(const signature_info_t *signature_info)
...
+  for (int i = 0; i < MAX_BUFFER_LENGTH; i++) {
+    free(self->output_buffer[i].signature);
+    self->output_buffer[i].signature = NULL;
+    self->output_buffer[i].size = 0;
+    self->output_buffer[i].signing_error = false;
+
+    free(self->input_buffer[i]);
+    self->input_buffer[i] = NULL;
+  }
   self->hash_size = 0;
```
bjornvolcker, line 134: Since you reset the buffers, you also need to reset the ```input_buffer_idx``` and ```output_buffer_idx```.

lib/plugins/threaded-signing/plugin.c
bjornvolcker: Ack

lib/plugins/threaded-signing/plugin.c
```diff
@@ -137,33 +149,75 @@ signing_worker_thread(void *user_data)
...
+      } else {
+        sleep(5);
+      }
       g_mutex_lock(&self->mutex);
-      // When successfully done with signing, move |plugin_state| to THREADED_SIGNING_HAS_SIGNATURE,
-      // otherwise move to THREADED_SIGNING_ERROR to report the error when getting the signature.
+
+      // |output_buffer| is full. Buffers this long are not supported.
+      if (self->output_buffer_idx >= MAX_BUFFER_LENGTH) {
+        self->is_running = false;
```
bjornvolcker, line 185: Stopping the thread will just stop the thread. Calls to plugin-APIs from the library thread will still keep running. The output buffer will never be filled further, but the input buffer will and the ```threaded_openssl_sign_hash()``` API will keep returning SV_OK until we reach ```MAX_BUFFER_LENGTH```.

The library will also keep a lot of unfinished SEI-frames. How can we make life easier for the main thread to know if something went wrong so proper actions can be made there as well.

You can mimic everything through the test code you have with proper settings and sleep(), free(), printf() and other actions.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -188,59 +240,61 @@ threaded_openssl_sign_hash(sv_threaded_plugin_t *self, const signature_info_t *s
...
-    status = SV_OK;
-    goto done;
-  }
-
   // If no |self->signature_info| exists. Allocate necessary memory for it and copy the
   // |private_key| and |algo|.
   if (!self->signature_info) {
     self->signature_info = local_signature_info_create(signature_info);
-    if (!self->signature_info) goto catch_error;
+    // Failed in memory allocation.
```
bjornvolcker, line 247: I think this comment should go _inside_ the if-statement for correct interpretation. Think of this as reading a book. Line 246 says ```signature_info``` is created. Then I read we failed allocating memory.

There are more places.

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
lurvas777: I second this. Generally I like to have comments just like docstrings in python, the first thing you read after you read the function name and parameters. Meaning that after I've read the context (function signature) I want to read any other info about that is useful.

Of course if the comment is "Check something..." then it should be before the statement. Example:

// Check if we exceed the max gop and do something about it
if (gop > max_gop) {
  do_something();
}

But generally it's better to have the statement after a check, since usually the check itself it clear:

if (gop > max_gop) {
  // When exceeding the max gop we need to do something
  do_something();
}

It makes it look cleaner and easier to read IMHO.

lurvas777 (commented):
Just one NIT.
When do we remove the code related to the sleep?

bjornvolcker (changes requested):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -188,59 +245,67 @@ threaded_openssl_sign_hash(sv_threaded_plugin_t *self, const signature_info_t *s
...
 
   status = SV_OK;
-  self->plugin_state = THREADED_SIGNING_HAS_HASH_TO_SIGN;
 
 catch_error:
-  if (status == SV_UNKNOWN_FAILURE) {
-    // Failed in memory allocation. Free all memory and set status to SV_MEMORY.
+  if (status != SV_OK) {
+    // Failed in memory allocation or failure due to full buffer. Free all memory.
     sv_threaded_plugin_reset(self);
```
bjornvolcker, line 298: I'm not sure it is a correct operation to clear all memory for all types of failure.
If we return a failure the ongoing SEI will be deleted and memory freed. But there may be others waiting for a signature, and those might wait for ever.
I suggest that we only reset if the thread has stopped.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -188,59 +245,67 @@ threaded_openssl_sign_hash(sv_threaded_plugin_t *self, const signature_info_t *s
...
 
-/* If |plugin_state| = THREADED_SIGNING_HAS_SIGNATURE a new |signature| has been written to
- * |signature_info|. The new |signature| is then copied to the output and the |plugin_state|
- * is set to THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN.
- *
- * Returns true if a new |signature| has been copied to output, otherwise false.
- * If the hash could not be signed due to a blocked openssl_sign_hash(), |signature_size| is set to
- * zero, but still returning true. */
+/* Returns true if the oldest signature in |output_buffer| has been copied to |signature|, otherwise
+ * false. Moves the signatures in |output_buffer| forward when the copy is done. */
```
bjornvolcker, line 308: Nit: "has been done"

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,31 +319,39 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
+
+  if (!self->is_running) {
+    // Thread is not running. Go to done and return error.
+    status = SV_EXTERNAL_ERROR;
+    goto done;
+  }
+
+  if (self->output_buffer_idx == 0) goto done;
+
+  *written_signature_size = 0;
```
bjornvolcker, line 331: This should be set before the sanity checks above.
Can even be moved before the lock.

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,31 +319,39 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
-    *written_signature_size = 0;
-    self->nbr_of_unsigned_hashes--;
+  } else if (self->output_buffer[0].size > max_signature_size) {
+    // There is no room to copy the signature, set status to invalid parameter.
+    status = SV_INVALID_PARAMETER;
+  } else {
+    // Copy the oldest signature
+    memcpy(signature, self->output_buffer[0].signature, self->output_buffer[0].size);
+    *written_signature_size = self->output_buffer[0].size;
+    // Change state and mark as copied.
```
bjornvolcker, line 342: Remove

lib/plugins/threaded-signing/plugin.c
```diff
@@ -254,31 +319,39 @@ threaded_openssl_get_signature(sv_threaded_plugin_t *self,
...
-    has_copied_signature = true;
-  } else if (self->plugin_state == THREADED_SIGNING_ERROR) {
-    *written_signature_size = 0;
-    // Change state and mark as copied.
-    self->plugin_state = THREADED_SIGNING_WAITS_FOR_HASH_TO_SIGN;
-    has_copied_signature = true;
+
+  if (!self->is_running) {
+    // Thread is not running. Go to done and return error.
+    status = SV_EXTERNAL_ERROR;
```
bjornvolcker, line 325: This error will not be picked up when calling ```sv_interface_get_signature()``` in ```signed_video_h26x_sign.c```. Note how errors propagate.

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #105, opened 2022-11-02
Adds new API to pass in a NALU in parts

bjornvolcker:
On the device side NALU data may be split in memory which the framework
cannot handle, since hashing has to be done in one go. This commit
enables partial updates of a hash. Hence, a NALU can now be added part
by part.

Two tests have been added for this scenario.


lusikamalo (approved):
OK.

Caroline-Holmstrom (approved):

PR #111, opened 2023-01-18
Extends parse_nalu_info() with is_auth_side parameter

bjornvolcker:
This prevents the signing side from hashing an incoming Signed
Video SEI frame.

In addition, pre-commit.sh made formatting changes to nalu_list.c.


lusikamalo (approved):
OK.

Caroline-Holmstrom (approved):

PR #114, opened 2023-03-17
Adds a setter to write SEIs w/wo emulation prevention

bjornvolcker:
This is a new API for the signing side which configures the Signed
Video session to produce SEIs with or without emulation prevention.
The default behavior is with emulation prevention turned on.

Two new tests have been added; one on the signing side to compare
SEIs w/wo emulation prevention, and one on the validation side to
verify that validation works even if emulation prevention bytes are
added later.

In addition, a name change of a member is done from tmp_tlv_memory
to nalu_data_wo_epb, since the entire nalu_data is copied instead
of just the tlv part.


lusikamalo (dismissed):
OK.

Caroline-Holmstrom (approved):

PR #116, opened 2023-05-02
Adds a queue in threaded-signing/plugin

bjornvolcker:
This solves the issue of failed signing if the signing
functionality was temporarily blocked when Signed Video made
the request.
Now incoming hashes are put in a queue, from which the signing
thread picks hash by hash.

A test has been added for this case.


bjornvolcker:
Code review done in a separate PR.

PR #118, opened 2023-08-07
Adds a new API to limit the SEI payload

bjornvolcker:
When the SEI exceeds the set maximum Signed Video falls back to
using the GOP level authentication for the current GOP.

Tests have been added and a new member for the struct sv_setting
to be able to modify the maximum payload size.


lusikamalo (approved):
LGTM.

PR #131, opened 2024-03-18
Simplify  prepend instruction.

mithydolphin:


lib/src/signed_video_h26x_sign.c
```diff
@@ -477,15 +479,15 @@ prepare_for_nalus_to_prepend(signed_video_t *self)
     // empty list item, the pull action has no impact. We can therefore silently remove it and
     // proceed. But if there are vital SEI-nalus waiting to be pulled we return an error message
     // (SV_NOT_SUPPORTED).
-    SVI_THROW_IF_WITH_MSG(self->num_nalus_to_prepend > 1, SVI_NOT_SUPPORTED,
+    //
```
bjornvolcker, line 482: Remove this line

lib/src/signed_video_h26x_sign.c
```diff
@@ -477,15 +479,15 @@ prepare_for_nalus_to_prepend(signed_video_t *self)
...
+    SVI_THROW_IF_WITH_MSG(self->num_of_completed_seis > 0, SVI_NOT_SUPPORTED,
         "There are remaining NALUs in list to prepend");
```
bjornvolcker, line 484: Update this debug message now when there is no list anymore

lib/src/signed_video_h26x_sign.c
```diff
@@ -477,15 +479,15 @@ prepare_for_nalus_to_prepend(signed_video_t *self)
...
-    if (self->num_nalus_to_prepend > 0) self->num_nalus_to_prepend = 0;
+    if (self->num_of_completed_seis > 0) self->num_of_completed_seis = 0;
```
bjornvolcker, line 485: Isn't this line redundant?

lib/src/signed_video_h26x_sign.c
```diff
@@ -625,30 +627,47 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
-signed_video_get_nalu_to_prepend(signed_video_t *self,
-    signed_video_nalu_to_prepend_t *nalu_to_prepend)
+signed_video_get_sei(signed_video_t *self, uint8_t *nalu_data, size_t *nalu_data_size)
 {
-  if (!self || !nalu_to_prepend) return SV_INVALID_PARAMETER;
 
-  if (self->num_nalus_to_prepend < 1) {
+  if (!self || !nalu_data_size) return SV_INVALID_PARAMETER;
+  *nalu_data_size = 0;
+  if (self->sei_data_buffer_idx < 1) {
```
bjornvolcker, line 635: Shouldn't you check `num_of_completed_seis` instead of the buffer index?

lib/src/signed_video_h26x_sign.c
```diff
@@ -451,12 +453,12 @@ add_nalu_to_prepend(signed_video_t *self,
   assert(data_size > 0);
 
   signed_video_nalu_to_prepend_t *nalu_to_prepend =
-      &(self->nalus_to_prepend_list[self->num_nalus_to_prepend]);
+      &(self->nalus_to_prepend_list[self->num_of_completed_seis]);
```
bjornvolcker, line 456: This list is not needed anymore. Hence, this function could be removed/replaced as well.

lib/src/signed_video_h26x_sign.c
```diff
@@ -55,7 +55,8 @@ add_nalu_to_prepend(signed_video_t *self,
     size_t data_size);
 static svi_rc
 prepare_for_nalus_to_prepend(signed_video_t *self);
-
+void
```
bjornvolcker, line 58: Make static

lib/src/signed_video_h26x_sign.c
```diff
@@ -158,22 +159,22 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
   SVI_DONE(status)
 
 done:
-  // Done with the SEI payload. Move |sei_data_buffer|. This should be done even if we caught a
-  // failure.
+
+  return status;
+}
+void
```
bjornvolcker, line 165: Add empty line between function definitions.

lib/src/signed_video_h26x_sign.c
```diff
@@ -158,22 +159,22 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
...
+free_and_reset_sei_data_buffer(signed_video_t *self)
```
bjornvolcker, line 166: The function name does not really reflect what it is doing. When I read the code it removes the first element. There is no free nor reset involved.

bjornvolcker (changes requested):

lib/src/includes/signed_video_sign.h
```diff
@@ -215,18 +223,14 @@ signed_video_add_nalu_for_signing(signed_video_t *self,
  *   }
  *
  * @param self Pointer to the signed_video_t object in use.
- * @param nalu_to_prepend A pointer to a signed_video_nalu_to_prepend_t object holding NALU data
- *   and prepend instructions.
+ * @param nalu_data Pointer to the SEI payload
+ * @param nalu_data_size Size of the NAlU
```
bjornvolcker, line 227:
```
@param sei Pointer to the memory to which a complete SEI will be copied. If a  NULL pointer is used, only the |sei_size| is updated.
@param sei_size Pointer to where the size of the SEI is written.
```

tests/check/check_h26xsigned_auth.c
```diff
@@ -1750,17 +1750,17 @@ START_TEST(vendor_axis_communications_operation)
   // Add an I-NALU to trigger a SEI.
   sv_rc = signed_video_add_nalu_for_signing(sv, i_nalu->data, i_nalu->data_size);
   ck_assert_int_eq(sv_rc, SV_OK);
-  sv_rc = signed_video_get_sei(sv, NULL, &data_size);
+  sv_rc = signed_video_get_sei(sv, NULL, &sei_size);
   ck_assert_int_eq(sv_rc, SV_OK);
-  uint8_t *nalu_data = malloc(data_size);
-  sv_rc = signed_video_get_sei(sv, nalu_data, &data_size);
+  uint8_t *nalu_data = malloc(sei_size);
```
bjornvolcker, line 1755: Check the size before malloc

tests/check/check_h26xsigned_auth.c
```diff
@@ -1715,12 +1715,12 @@ START_TEST(vendor_axis_communications_operation)
   SignedVideoCodec codec = settings[_i].codec;
   sign_algo_t algo = settings[_i].algo;
   SignedVideoAuthenticityLevel auth_level = settings[_i].auth_level;
-  //  signed_video_nalu_to_prepend_t nalu_to_prepend = {0};
+
   nalu_list_item_t *i_nalu = nalu_list_item_create_and_set_id('I', 0, codec);
   nalu_list_item_t *sei = NULL;
```
bjornvolcker, line 1720: Rename this to `sei_nalu` and the `nalu_data` below to `sei`

tests/check/check_h26xsigned_auth.c
```diff
@@ -1814,9 +1813,8 @@ generate_and_set_private_key_on_camera_side(struct sv_setting setting,
   SignedVideoReturnCode sv_rc;
   char *private_key = NULL;
   size_t private_key_size = 0;
-  //  signed_video_nalu_to_prepend_t nalu_to_prepend = {0};
   nalu_list_item_t *i_nalu = nalu_list_item_create_and_set_id('I', 0, setting.codec);
-  size_t data_size;
+  size_t sei_size;
```
bjornvolcker, line 1817: Declare this where you use it, or move the declaration of `sei` (former `nalu_data`) here.

lib/src/signed_video_h26x_sign.c
```diff
@@ -143,12 +136,11 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
     data_size = get_sign_and_complete_sei_nalu(self, &payload, payload_signature_ptr);
     SVI_THROW_IF(!data_size, SVI_UNKNOWN);
     // Add created SEI to the prepend list.
-    prepend_instruction = SIGNED_VIDEO_PREPEND_NALU;
     // TODO: Include setting |nalu_data| in add_nalu_to_prepend().
     // Transfer |payload| to |nalu_to_prepend|.
```
bjornvolcker, line 140: This comment is incorrect now

lib/src/signed_video_h26x_sign.c
```diff
@@ -116,7 +110,6 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
   // Get the oldest sei data
   const int sei_data_buffer_end = self->sei_data_buffer_idx;
   assert(sei_data_buffer_end <= MAX_NALUS_TO_PREPEND);
-  SignedVideoPrependInstruction prepend_instruction = SIGNED_VIDEO_PREPEND_NOTHING;
   size_t data_size = 0;
   svi_rc status = SVI_UNKNOWN;
   sei_data_t *nalu_to_prepend = &(self->sei_data_buffer[self->num_of_completed_seis]);
```
bjornvolcker, line 115: I think you should rename `nalu_to_prepend` to for example `sei_data` now when we deal with sei_data_buffer instead.

lib/src/signed_video_h26x_sign.c
```diff
@@ -162,8 +154,11 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
 
   return status;
 }
+
+/* Removes the first element from the SEI buffer of a `signed_video_t` structure,
+ * shifting remaining elements left and clearing the last slot. */
 void
```
bjornvolcker, line 160: nit: static void

lib/src/signed_video_h26x_sign.c
```diff
@@ -479,12 +432,10 @@ prepare_for_nalus_to_prepend(signed_video_t *self)
     // empty list item, the pull action has no impact. We can therefore silently remove it and
     // proceed. But if there are vital SEI-nalus waiting to be pulled we return an error message
     // (SV_NOT_SUPPORTED).
-    //
     SVI_THROW_IF_WITH_MSG(self->num_of_completed_seis > 0, SVI_NOT_SUPPORTED,
         "There are remaining NALUs in list to prepend");
     if (self->num_of_completed_seis > 0) self->num_of_completed_seis = 0;
 
-    assert(self->num_of_completed_seis == 0);
```
bjornvolcker: You removed the wrong line

lib/src/signed_video_h26x_sign.c
```diff
@@ -658,12 +609,17 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
     signed_video_nalu_to_prepend_t *nalu_to_prepend)
 {
```
bjornvolcker, line 610: Note that this function provides the user with SEIs in order from newest to oldest. You need to make sure that is still the case.

bjornvolcker (changes requested):
I'll continue the review tomorrow.

lib/src/signed_video_h26x_sign.c
```diff
@@ -578,32 +573,59 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
+  signed_video_nalu_data_free(self->sei_data_buffer[self->num_of_completed_seis].payload);
+
+  --(self->num_of_completed_seis);
+  shift_sei_buffer_index(self);
+  return SV_OK;
+}
+
 SignedVideoReturnCode
 signed_video_get_nalu_to_prepend(signed_video_t *self,
     signed_video_nalu_to_prepend_t *nalu_to_prepend)
```
bjornvolcker, line 631: This function does not work. It can write out of memory if there are more than one pending SEI.
How did you test it?

lib/src/signed_video_h26x_sign.c
```diff
@@ -578,32 +573,59 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
+  // Reset the fetched SEI information from the sei buffer.
   signed_video_nalu_data_free(self->sei_data_buffer[0].payload);
 
   --(self->num_of_completed_seis);
   shift_sei_buffer_index(self);
   return SV_OK;
 }
 
+static SignedVideoReturnCode
+signed_video_get_lastest_sei(signed_video_t *self, uint8_t *sei, size_t *sei_size)
```
bjornvolcker, line 603: Rename `get_latest_sei`
Only public APIs have the preface "signed_video".

This function does not work. How did you test it?
It will get you the oldest non-completed SEI and not the latest completed SEI. You will also lose the oldest SEI.

bjornvolcker (changes requested):
There are too many errors and the functionality does not work for multiple SEIs. Please, consider manipulating the code and tests so you can verify your changes.

Also rework the tests. They do not fully reflect what should be caught and there is a mix of names between like nalu_data and sei_size

lib/src/signed_video_h26x_sign.c
```diff
@@ -625,30 +621,186 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
+    return SV_OK;
+  }
+  // Assign the SEI size and SEI data to the provided pointers.
+  *sei_size = self->sei_data_buffer[self->num_of_completed_seis - 1].completed_sei_size;
+  memcpy(sei, self->sei_data_buffer[self->num_of_completed_seis - 1].payload, *sei_size);
+
+  // Reset the fetched SEI information from the sei buffer.
+
+  --(self->num_of_completed_seis);
+  shift_sei_buffer_last_index(self);
```
bjornvolcker, line 788: After this point the buffer has an empty item at the beginning which probably will break the system if we have more than one completed SEI. Which means that the added test is not testing things correctly.

lib/src/signed_video_h26x_sign.c
```diff
@@ -158,41 +174,35 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
...
+static void
+shift_sei_buffer_index(signed_video_t *self)
+{
+  const int sei_data_buffer_end = self->sei_data_buffer_idx;
   for (int j = 0; j < sei_data_buffer_end - 1; j++) {
     self->sei_data_buffer[j] = self->sei_data_buffer[j + 1];
   }
   self->sei_data_buffer[sei_data_buffer_end - 1].payload = NULL;
   self->sei_data_buffer[sei_data_buffer_end - 1].payload_signature_ptr = NULL;
   self->sei_data_buffer[sei_data_buffer_end - 1].last_two_bytes = LAST_TWO_BYTES_INIT_VALUE;
```
bjornvolcker, line 194: What about the members `completed_sei` and `completed_sei_size`? You need to set/initialize them as well.

bjornvolcker (changes requested):

tests/check/check_h26xsigned_sign.c
```diff
@@ -374,14 +372,18 @@ START_TEST(vendor_axis_communications_operation)
...
+  ck_assert(sei_size > 0);
+  uint8_t *nalu_data = malloc(sei_size);
+  sv_rc = signed_video_get_sei(sv, nalu_data, &sei_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sei = nalu_list_create_item(nalu_data, sei_size, codec);
   ck_assert(tag_is_present(sei, codec, VENDOR_AXIS_COMMUNICATIONS_TAG));
-  // Ownership of |nalu_to_prepend.nalu_data| has been transferred. Do not free memory.
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
+  // Ownership of |nalu_data| has been transferred. Do not free memory.
+  sv_rc = signed_video_get_sei(sv, nalu_data, &sei_size);
```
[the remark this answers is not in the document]
mithydolphin: there is a variable name sei in line 381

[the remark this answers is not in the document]
bjornvolcker: Then rename the one on line 381 to `sei_item` as suggested in the review comment above.

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
 }
 END_TEST
 
-START_TEST(two_seis_in_buffer)
+/* Test description
+ * Verify that after 2 completed Sei's created ,they will be transfered in
+ * correct order
+ * The operation is as follows:
+ * 1. Setup a signed_video_t session
+ * 2. Add  2 I NALU for signing that will trigger 2 Sei Nalu
```
bjornvolcker, line 589: Remove one white space between "Add" and "2"
Sei -> SEI

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
+ * 3. Get the Sei NALU's to prepend
```
bjornvolcker, line 590: NALU's -> NALUs

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
+ * 4. Check that Sei NALU's transfered in  correct order
```
bjornvolcker, line 591: - Sei -> SEI
- NALU's -> NALUs
- transfered -> were emitted
- remove one white space before "correct"

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
+  signed_video_nalu_to_prepend_t nalu_to_prepend_2 = {0};
+
   signed_video_t *sv = signed_video_create(codec);
+  sv->sv_test_on = true;
   ck_assert(sv);
   char *private_key = NULL;
   size_t private_key_size = 0;
   nalu_list_item_t *i_nalu = nalu_list_item_create_and_set_id('I', 0, codec);
-
+  nalu_list_item_t *i_nalu_2 = nalu_list_item_create_and_set_id('I', 0, codec);
```
bjornvolcker, line 606: You should have a different id for this I-frame otherwise the hash of `i_nalu_2` will be identical with that of `i_nalu`

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
+/* Test description
+ * Verify that after 2 completed Sei's created ,they will be transfered in
+ * correct order
+ * The operation is as follows:
+ * 1. Setup a signed_video_t session
+ * 2. Add  2 I NALU for signing that will trigger 2 Sei Nalu
+ * 3. Get the Sei NALU's to prepend
+ * 4. Check that Sei NALU's transfered in  correct order
+ */
+START_TEST(two_completed_seis_pending)
```
bjornvolcker, line 593: You should have the same test but when using `signed_video_get_sei(...)`
Name this test `two_completed_seis_pending_legacy` and keep the name with the new API

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
 {
-  // This test runs in a loop with loop index _i, corresponding to struct sv_setting _i in
-  // |settings|; See signed_video_helpers.h.
-
   SignedVideoCodec codec = settings[_i].codec;
   SignedVideoReturnCode sv_rc;
   signed_video_nalu_to_prepend_t nalu_to_prepend = {0};
+  signed_video_nalu_to_prepend_t nalu_to_prepend_2 = {0};
```
bjornvolcker, line 598: Since you name this one `nalu_to_prepend_2` the other one should be named `nalu_to_prepend_1`.
Same goes for `i_nalu` below

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
+      0);
 
-  // Remove the signature
-  update_hashable_data(&nalu);
-
-  // Verify that hashable data sizes and data contents are identical
-  ck_assert(nalu.hashable_data_size > 0);
-
-  free(nalu.nalu_data_wo_epb);
   nalu_list_free_item(i_nalu);
```
bjornvolcker, line 639: You also need to free `i_nalu_2` to avoid memory leaks

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
-  sv_rc = signed_video_add_nalu_for_signing_with_timestamp_for_test(
-      sv, i_nalu->data, i_nalu->data_size, NULL);
+  // After 2 seis are created we can copy the sei buffer
+  sei_data_t sei_buff[2];
+  memcpy(sei_buff, sv->sei_data_buffer, sizeof(sei_buff));
   ck_assert_int_eq(sv_rc, SV_OK);
   sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
   ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend_2);
+  ck_assert_int_eq(sv_rc, SV_OK);
```
bjornvolcker, line 630: You should make a third call and verify that you get a "nothing to do" instruction
This can preferably be done after the checks below

tests/check/check_h26xsigned_sign.c
```diff
@@ -581,53 +581,69 @@ START_TEST(undefined_nalu_in_sequence)
...
+  sv_rc = signed_video_add_nalu_for_signing_with_timestamp(
+      sv, i_nalu_2->data, i_nalu_2->data_size, NULL);
+  ck_assert_int_eq(sv_rc, SV_OK);
 
-  sv_rc = signed_video_add_nalu_for_signing_with_timestamp_for_test(
-      sv, i_nalu->data, i_nalu->data_size, NULL);
+  // After 2 seis are created we can copy the sei buffer
+  sei_data_t sei_buff[2];
+  memcpy(sei_buff, sv->sei_data_buffer, sizeof(sei_buff));
```
bjornvolcker, line 625: This is indeed a feasible way to check it, but it means that the test need to know what is done internally. That should if possible be avoided. And it can be done differently. For example, if you add I -> P -> I you can verify the order through the SEI sizes.
Sometimes you need to think outside the box ;-)

bjornvolcker (changes requested):
Some comments for now

tests/check/check_h26xsigned_sign.c
mithydolphin: Currently signed_video_get_sei(..) will  get the first element in SEI buffer  .Should I modify  it such a way that if there is more then 1 completed SEI waiting in the buffer it will behave like  signed_video_get_nalu_to_prepend(..)

bjornvolcker (changes requested):
It still does not work.
Let the state of the `sei_data_buffer` when calling either get_sei() or get_nalus_to_prepend() is
[C1, C2, C3, P1, P2, E, E] (C = Completed, P = in Progress, E = Empty)

After calling the get_sei() repeatedly the buffer should look like this
[C2, C3, P1, P2, E, E, E]
[C3, P1, P2, E, E, E, E]
[P1, P2, E, E, E, E, E]

and for the legacy API
[C1, C2, P1, P2, E, E, E]
[C1, P1, P2, E, E, E, E]
[P1, P2, E, E, E, E, E]

Can you verify that with pen & paper?

lib/src/includes/signed_video_sign.h
```diff
@@ -247,11 +247,10 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
  * bytes. Hence, the user can simply pull and prepend existing H26x NALUs. Pull NALUs to prepend
  * from signed_video_t one by one until no more generated SEIs exists, that is, when |sei_size|
  * is zero.
- * API provides the user with the NALU data as well as instructions on where to prepend that data.
  *
- * NOTE that as soon as the user pulls a new NALU to prepend, the ownership of the |nalu_data|
- * memory (see members of signed_video_nalu_to_prepend_t) is transferred. Free the |nalu_data|
- * memory with signed_video_nalu_data_free(...).
+ * NOTE: The user is responsible for allocating memory for the |nalu_data|
```
bjornvolcker, line 251: nalu_data -> sei, since that is the name of the input parameter

lib/src/includes/signed_video_sign.h
```diff
@@ -247,11 +247,10 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
...
+ * (see members of signed_video_nalu_to_prepend_t). Ensure that you manage the memory appropriately.
+ * To free the |nalu_data| memory, use signed_video_nalu_data_free(...).
```
bjornvolcker, line 253: This is incorrect now when the user allocates memory. Remove.

lib/src/includes/signed_video_sign.h
```diff
@@ -247,11 +247,10 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
  * bytes. Hence, the user can simply pull and prepend existing H26x NALUs. Pull NALUs to prepend
  * from signed_video_t one by one until no more generated SEIs exists, that is, when |sei_size|
  * is zero.
- * API provides the user with the NALU data as well as instructions on where to prepend that data.
  *
- * NOTE that as soon as the user pulls a new NALU to prepend, the ownership of the |nalu_data|
- * memory (see members of signed_video_nalu_to_prepend_t) is transferred. Free the |nalu_data|
- * memory with signed_video_nalu_data_free(...).
+ * NOTE: The user is responsible for allocating memory for the |nalu_data|
+ * (see members of signed_video_nalu_to_prepend_t). Ensure that you manage the memory appropriately.
```
bjornvolcker, line 252: Do not reference an API that is going to be deprecated.
Also, when describing things it should be written anonymous, that is, no "I", "we" or "you". For example, this sentence should be written as "The user is responsible for managing the memory appropriately". This sentence though is redundant since it is already stated that the user allocates memory.

lib/src/signed_video_h26x_sign.c
```diff
@@ -154,13 +154,13 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
...
-/* Removes the first element from the SEI buffer of a `signed_video_t` structure,
+/* Removes the specified index element from the SEI buffer of a `signed_video_t` structure,
  * shifting remaining elements left and clearing the last slot. */
 static void
-shift_sei_buffer_index(signed_video_t *self)
+shift_sei_buffer_at_index(signed_video_t *self, int index)
 {
   const int sei_data_buffer_end = self->sei_data_buffer_idx;
-  for (int j = 0; j < sei_data_buffer_end - 1; j++) {
+  for (int j = index; j < sei_data_buffer_end; j++) {
```
bjornvolcker, line 163: You lost the "-1". If it is not there you could potentially move in garbage into the buffer.

lib/src/signed_video_h26x_sign.c
```diff
@@ -452,7 +441,7 @@ prepare_for_nalus_to_prepend(signed_video_t *self)
     // (SV_NOT_SUPPORTED).
     if (!self->sv_test_on) {
       SVI_THROW_IF_WITH_MSG(
-          self->num_of_completed_seis > 0, SVI_NOT_SUPPORTED, "There are remainin SEI's.");
+          self->num_of_completed_seis > 0, SVI_NOT_SUPPORTED, "There are remainin SEIs.");
```
bjornvolcker, line 444: grammar: "remaining"

lib/src/signed_video_h26x_sign.c
```diff
@@ -603,12 +592,12 @@ signed_video_get_lastest_sei(signed_video_t *self, uint8_t *sei, size_t *sei_siz
...
   *sei_size = self->sei_data_buffer[self->num_of_completed_seis - 1].completed_sei_size;
   if (!sei) return SV_OK;
-  // Assign SEI data to the provided pointer.
+  // Copy SEI data to the provided pointer.
   memcpy(sei, self->sei_data_buffer[self->num_of_completed_seis - 1].payload, *sei_size);
 
   // Reset the fetched SEI information from the sei buffer.
   --(self->num_of_completed_seis);
-  reset_sei_buffer_at_index(self, self->num_of_completed_seis);
+  shift_sei_buffer_at_index(self, self->num_of_completed_seis);
```
bjornvolcker, line 600: You will leak memory on this line. Run valgrind to see if there are other memory leaks

lib/src/signed_video_h26x_sign.c
```diff
@@ -592,7 +581,7 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
 }
 
 static SignedVideoReturnCode
```
bjornvolcker, line 583: I would prefer that this static function returns `svi_rc` instead.

lib/src/signed_video_h26x_sign.c
```diff
@@ -647,16 +632,15 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
...
   if (SV_OK != status) return status;
   if (*sei_size != 0) {
     nalu_to_prepend->nalu_data = malloc(*sei_size);
     nalu_to_prepend->prepend_instruction = SIGNED_VIDEO_PREPEND_NALU;
   } else {
     nalu_to_prepend->prepend_instruction = SIGNED_VIDEO_PREPEND_NOTHING;
   }
-  return signed_video_get_lastest_sei(
-      self, (nalu_to_prepend->nalu_data), &(nalu_to_prepend->nalu_data_size));
+  return get_latest_sei(self, (nalu_to_prepend->nalu_data), &(nalu_to_prepend->nalu_data_size));
```
bjornvolcker, line 643: Remove unnecessary parentheses.

bjornvolcker (changes requested):
There is a memory leak. I only looked at the last changes. I will have a look at everything again.

lib/src/signed_video_internal.h
```diff
@@ -105,6 +105,8 @@ struct _sei_data_t {
   uint8_t *payload;  // Pointer to the allocated SEI data
   uint8_t *payload_signature_ptr;
   uint16_t last_two_bytes;
+  uint8_t *completed_sei;  // Number of completed sei
```
bjornvolcker, line 108: Remove this member. It is never used.
In another PR we should rename "payload" to "sei" and "payload_signature_ptr" to "write_position"

lib/src/signed_video_internal.h
```diff
@@ -105,6 +105,8 @@ struct _sei_data_t {
...
+  size_t completed_sei_size;  // nalu_data_size = payload_signature_ptr - payload
```
bjornvolcker, line 109: The comment used a variable that does not exist. Rename.

lib/src/signed_video_h26x_sign.c
```diff
@@ -141,14 +137,9 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
     // Add the signature to the SEI payload.
     data_size = get_sign_and_complete_sei_nalu(self, &payload, payload_signature_ptr);
     SVI_THROW_IF(!data_size, SVI_UNKNOWN);
```
bjornvolcker, line 139: Now when there is only one THROW I think the TRY/CATCH/DONE is overkill

lib/src/signed_video_h26x_sign.c
```diff
@@ -116,13 +112,13 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
   // Get the oldest sei data
   const int sei_data_buffer_end = self->sei_data_buffer_idx;
```
bjornvolcker, line 113: This variable is only used in the assert below. Merge the lines into one.

lib/src/signed_video_h26x_sign.c
```diff
@@ -159,41 +150,25 @@ complete_sei_nalu_and_add_to_prepend(signed_video_t *self)
...
-  self->sei_data_buffer[sei_data_buffer_end - 1].last_two_bytes = LAST_TWO_BYTES_INIT_VALUE;
-  self->sei_data_buffer_idx -= 1;
 
   return status;
 }
 
-/* Resets a signed_video_nalu_to_prepend_t object. It is assumed that any nalu_data has been freed
- * to avoid memory leakage.
- */
+/* Removes the specified index element from the SEI buffer of a `signed_video_t` structure,
```
bjornvolcker, line 157: grammar: Remove the extra white space and replace the "," with "by"

lib/src/signed_video_h26x_sign.c
```diff
@@ -631,29 +580,72 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
+static svi_rc
+get_latest_sei(signed_video_t *self, uint8_t *sei, size_t *sei_size)
+{
+  if (!self || !sei_size) return SVI_INVALID_PARAMETER;
+  *sei_size = 0;
+  if (self->num_of_completed_seis < 1) {
+    DEBUG_LOG("There are no completed seis.");
+    return SVI_OK;
+  }
+  // Assign the SEI size to the provided pointer.
```
bjornvolcker, line 592: Remove this line

lib/src/signed_video_h26x_sign.c
```diff
@@ -631,29 +580,72 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
+  svi_rc status = get_latest_sei(self, NULL, sei_size);
+  if (SVI_OK == status) {
+    if (*sei_size != 0) {
+      nalu_to_prepend->nalu_data = malloc(*sei_size);
+      nalu_to_prepend->prepend_instruction = SIGNED_VIDEO_PREPEND_NALU;
+    } else {
+      nalu_to_prepend->prepend_instruction = SIGNED_VIDEO_PREPEND_NOTHING;
+    }
+    return svi_rc_to_signed_video_rc(
+        get_latest_sei(self, nalu_to_prepend->nalu_data, &nalu_to_prepend->nalu_data_size));
```
bjornvolcker, line 645: It is better coding to set status here
`status = get_latest_sei(self, nalu_to_prepend->nalu_data, &nalu_to_prepend->nalu_data_size);`
and then skip the else below, by moving out that return.

lib/src/signed_video_internal.h
```diff
@@ -175,6 +173,7 @@ struct _signed_video_t {
   size_t arbitrary_data_size;  // Size of |arbitrary_data|.
 
   bool has_public_key;  // State to indicate if public key is received/added
+  bool sv_test_on;  // State to indicate if public key is received/added
```
bjornvolcker, line 176: Move this to the bottom of this struct well separated from the other members and add  like a section for test and debug variables.

tests/check/check_h26xsigned_auth.c
```diff
@@ -1715,11 +1715,11 @@ START_TEST(vendor_axis_communications_operation)
   SignedVideoCodec codec = settings[_i].codec;
   sign_algo_t algo = settings[_i].algo;
   SignedVideoAuthenticityLevel auth_level = settings[_i].auth_level;
-  signed_video_nalu_to_prepend_t nalu_to_prepend = {0};
   nalu_list_item_t *i_nalu = nalu_list_item_create_and_set_id('I', 0, codec);
   nalu_list_item_t *sei = NULL;
```
bjornvolcker, line 1719: Rename to `sei_item` ot `sei_nalu`

tests/check/check_h26xsigned_auth.c
```diff
@@ -1749,14 +1749,17 @@ START_TEST(vendor_axis_communications_operation)
   // Add an I-NALU to trigger a SEI.
   sv_rc = signed_video_add_nalu_for_signing(sv, i_nalu->data, i_nalu->data_size);
   ck_assert_int_eq(sv_rc, SV_OK);
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
+  sv_rc = signed_video_get_sei(sv, NULL, &sei_size);
+  ck_assert(sei_size > 0);
   ck_assert_int_eq(sv_rc, SV_OK);
-  sei = nalu_list_create_item(nalu_to_prepend.nalu_data, nalu_to_prepend.nalu_data_size, codec);
+  uint8_t *nalu_data = malloc(sei_size);
```
bjornvolcker, line 1755: rename to `sei`

tests/check/check_h26xsigned_auth.c
```diff
@@ -1828,18 +1829,23 @@ generate_and_set_private_key_on_camera_side(struct sv_setting setting,
...
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
+
+  size_t sei_size = 0;
+  sv_rc = signed_video_get_sei(sv, NULL, &sei_size);
+  ck_assert(sei_size > 0);
   ck_assert_int_eq(sv_rc, SV_OK);
-  *sei = nalu_list_create_item(
-      nalu_to_prepend.nalu_data, nalu_to_prepend.nalu_data_size, setting.codec);
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
+  uint8_t *nalu_data = malloc(sei_size);
```
bjornvolcker, line 1837: rename `nalu_data` to `sei` and `sei` to `sei_item` or `sei_nalu`

tests/check/check_h26xsigned_auth.c
```diff
@@ -1828,18 +1829,23 @@ generate_and_set_private_key_on_camera_side(struct sv_setting setting,
...
+  sv_rc = signed_video_get_sei(sv, nalu_data, &sei_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  *sei = nalu_list_create_item(nalu_data, sei_size, setting.codec);
+  sv_rc = signed_video_get_sei(sv, nalu_data, &sei_size);
```
bjornvolcker, line 1841: This is an incorrect operation. You should pass in `NULL` instead of `nalu_data`. Otherwise you could by accident overwrite the data in the just created item.

tests/check/check_h26xsigned_auth.c
```diff
@@ -2057,6 +2063,7 @@ START_TEST(no_emulation_prevention_bytes)
   nalu_list_item_t *i_nalu = nalu_list_item_create_and_set_id('I', 0, codec);
   nalu_list_item_t *sei = NULL;
```
bjornvolcker, line 2064: Rename to `sei_item` or `sei_nalu`

tests/check/check_h26xsigned_auth.c
```diff
@@ -2091,28 +2098,32 @@ START_TEST(no_emulation_prevention_bytes)
       sv, i_nalu->data, i_nalu->data_size, &g_testTimestamp);
   ck_assert_int_eq(sv_rc, SV_OK);
 
-  signed_video_nalu_to_prepend_t nalu_to_prepend = {0};
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
+  sv_rc = signed_video_get_sei(sv, NULL, &data_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  uint8_t *nalu_data = malloc(data_size);
```
bjornvolcker, line 2103: Rename to `sei`

tests/check/check_h26xsigned_auth.c
```diff
@@ -2057,6 +2063,7 @@ START_TEST(no_emulation_prevention_bytes)
   nalu_list_item_t *i_nalu = nalu_list_item_create_and_set_id('I', 0, codec);
   nalu_list_item_t *sei = NULL;
 
+  size_t data_size;
```
bjornvolcker, line 2066: Rename to `sei_size`

tests/check/check_h26xsigned_auth.c
```diff
@@ -2091,28 +2098,32 @@ START_TEST(no_emulation_prevention_bytes)
...
-  signed_video_nalu_to_prepend_t nalu_to_prepend = {0};
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
+  sv_rc = signed_video_get_sei(sv, NULL, &data_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  uint8_t *nalu_data = malloc(data_size);
+  sv_rc = signed_video_get_sei(sv, nalu_data, &data_size);
+
   ck_assert_int_eq(sv_rc, SV_OK);
-  ck_assert(nalu_to_prepend.prepend_instruction != SIGNED_VIDEO_PREPEND_NOTHING);
+  ck_assert(0 != nalu_data);
```
bjornvolcker, line 2107: Switch order of the check

tests/check/check_h26xsigned_auth.c
```diff
@@ -2091,28 +2098,32 @@ START_TEST(no_emulation_prevention_bytes)
...
 
   // Allocate memory for a new buffer to write to, and add emulation prevention bytes.
-  uint8_t *sei_data = malloc(nalu_to_prepend.nalu_data_size * 4 / 3);
+  uint8_t *sei_data = malloc(data_size * 4 / 3);
```
bjornvolcker, line 2110: Rename to `sei_with_epb` and change related variables the same way

tests/check/check_h26xsigned_sign.c
```diff
@@ -52,24 +52,19 @@ static SignedVideoReturnCode
...
+  sv_rc = signed_video_get_sei(sv, NULL, &sei_size);
   int num_pulled_nalus = 0;
-
-  if (num_nalus_to_pull == 0) goto done;
-
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
-  while (sv_rc == SV_OK && nalu_to_prepend.prepend_instruction != SIGNED_VIDEO_PREPEND_NOTHING) {
-
+  while (sv_rc == SV_OK && sei_size > 0) {
+    uint8_t *sei = malloc(sei_size);
```
bjornvolcker, line 59: You should probably check `sei` for NULL

tests/check/check_h26xsigned_sign.c
```diff
@@ -374,17 +372,21 @@ START_TEST(vendor_axis_communications_operation)
...
-  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend);
+  ck_assert(sei_size > 0);
+  uint8_t *sei = malloc(sei_size);
+  sv_rc = signed_video_get_sei(sv, sei, &sei_size);
   ck_assert_int_eq(sv_rc, SV_OK);
-  ck_assert(nalu_to_prepend.prepend_instruction == SIGNED_VIDEO_PREPEND_NOTHING);
+  sei_item = nalu_list_create_item(sei, sei_size, codec);
+  ck_assert(tag_is_present(sei_item, codec, VENDOR_AXIS_COMMUNICATIONS_TAG));
+  // Ownership of |sei| has been transferred. Do not free memory.
+  sv_rc = signed_video_get_sei(sv, sei, &sei_size);
```
bjornvolcker, line 384: You should pass in NULL instead of `sei`

tests/check/check_h26xsigned_sign.c
```diff
@@ -579,6 +581,132 @@ START_TEST(undefined_nalu_in_sequence)
...
+  // By construction, run the test for SV_AUTHENTICITY_LEVEL_FRAME only.
+  if (settings[_i].auth_level != SV_AUTHENTICITY_LEVEL_FRAME) return;
+
+  SignedVideoCodec codec = settings[_i].codec;
+  SignedVideoReturnCode sv_rc;
+  size_t sei_size_1 = 0;
+  size_t sei_size_2 = 0;
+  size_t sei_size_3 = 0;
+  signed_video_t *sv = signed_video_create(codec);
+  sv->sv_test_on = true;
```
bjornvolcker, line 603: Add a comment that you are bypassing code for checking if there are available SEIs

tests/check/check_h26xsigned_sign.c
```diff
@@ -579,6 +581,132 @@ START_TEST(undefined_nalu_in_sequence)
...
+  sv_rc = signed_video_set_private_key(sv, settings[_i].algo, private_key, private_key_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_set_authenticity_level(sv, settings[_i].auth_level);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_add_nalu_for_signing(sv, i_nalu_1->data, i_nalu_1->data_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_add_nalu_for_signing(sv, i_nalu_2->data, i_nalu_2->data_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+
+  // After 2 seis are created we can copy the sei buffer
```
bjornvolcker, line 623: grammar: "... we can get the SEIs"

tests/check/check_h26xsigned_sign.c
```diff
@@ -579,6 +581,132 @@ START_TEST(undefined_nalu_in_sequence)
...
+  sv_rc = signed_video_get_sei(sv, NULL, &sei_size_1);
```
bjornvolcker, line 624: You should sanity check the sizes

tests/check/check_h26xsigned_sign.c
```diff
@@ -579,6 +581,132 @@ START_TEST(undefined_nalu_in_sequence)
...
+  sv_rc = signed_video_get_sei(sv, NULL, &sei_size_2);
+  uint8_t *sei_2 = malloc(sei_size_2);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_get_sei(sv, sei_2, &sei_size_2);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_get_sei(sv, NULL, &sei_size_3);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  // Check if there are no more SEIs remaining in the buffer
+  ck_assert_int_eq(sei_size_3, 0);
+  // Verify the transfer order of NALUs
```
bjornvolcker, line 638: Add information why `sei_size_1` should be smaller than `sei_size_2`

tests/check/check_h26xsigned_sign.c
```diff
@@ -579,6 +581,132 @@ START_TEST(undefined_nalu_in_sequence)
...
+
+  // After 2 seis are created we can copy the sei buffer
+  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend_1);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend_2);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_get_nalu_to_prepend(sv, &nalu_to_prepend_3);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  ck_assert_int_eq(nalu_to_prepend_3.prepend_instruction, SIGNED_VIDEO_PREPEND_NOTHING);
+  // Verify the transfer order of NALUs
```
bjornvolcker, line 698: Add comment on why size 1 should be greater than 2

tests/check/check_h26xsigned_sign.c
```diff
@@ -579,6 +581,132 @@ START_TEST(undefined_nalu_in_sequence)
...
+  sv_rc = signed_video_set_private_key(sv, settings[_i].algo, private_key, private_key_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_set_authenticity_level(sv, settings[_i].auth_level);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_add_nalu_for_signing(sv, i_nalu_1->data, i_nalu_1->data_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_add_nalu_for_signing(sv, i_nalu_2->data, i_nalu_2->data_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+
+  // After 2 seis are created we can copy the sei buffer
```
bjornvolcker, line 690: Update comment like in previous test

bjornvolcker (changes requested):

lib/src/includes/signed_video_sign.h
```diff
@@ -243,8 +243,7 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
  * is zero.
  *
  * NOTE: The user is responsible for allocating memory for the |sei|
```
bjornvolcker, line 245: grammar: Add a period at end of this line

lib/src/includes/signed_video_sign.h
```diff
@@ -243,8 +243,7 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
...
- * (see members of signed_video_nalu_to_prepend_t). Ensure that the memory is managed appropriately.
- *
+ * Ensure that the user manages the memory appropriately.
```
bjornvolcker, line 246: Remove this line. It is redundant with the word "responsible" on previous line

lib/src/includes/signed_video_sign.h
```diff
@@ -243,8 +243,7 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
  * is zero.
  *
  * NOTE: The user is responsible for allocating memory for the |sei|
- * (see members of signed_video_nalu_to_prepend_t). Ensure that the memory is managed appropriately.
- *
```
bjornvolcker: Add back this line

lib/src/includes/signed_video_sign.h
```diff
@@ -262,11 +261,11 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
...
- *     // The first call of the function is for getiing the sei_size.
- *     // The Second call is to  get the sei.
- *     uint8_t *sei = malloc(sei_size);
- *     status = signed_video_get_sei(sv, sei, &sei_size);
+ *     // The first call of the function is for geting the |sei_size|.
+ *     // The second call is to  get the sei.
  *     while (status == SV_OK && sei_size != 0) {
+ *       uint8_t *sei = malloc(sei_size);
+ *       status = signed_video_get_sei(sv, sei, &sei_size);
  *         break;
```
bjornvolcker, line 269: This is incorrect. It will not compile if used and it dose not match the principle used in our tests for example

lib/src/signed_video_h26x_sign.c
```diff
@@ -641,11 +640,10 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
     } else {
       nalu_to_prepend->prepend_instruction = SIGNED_VIDEO_PREPEND_NOTHING;
```
bjornvolcker, line 641: You should should reset the input `nalu_to_prepend` before calling get_latest_sei

lib/src/signed_video_h26x_sign.c
```diff
@@ -641,11 +640,10 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
...
     }
-    return svi_rc_to_signed_video_rc(
-        get_latest_sei(self, nalu_to_prepend->nalu_data, &nalu_to_prepend->nalu_data_size));
-  } else {
+    status = get_latest_sei(self, nalu_to_prepend->nalu_data, &nalu_to_prepend->nalu_data_size);
```
bjornvolcker, line 643: This should only be called if `*sei_size > 0`

lib/src/signed_video_h26x_sign.c
```diff
@@ -641,11 +640,10 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
...
     return svi_rc_to_signed_video_rc(status);
```
bjornvolcker, line 644: Remove

lib/src/signed_video_internal.h
```diff
@@ -173,13 +173,15 @@ struct _signed_video_t {
...
   bool has_public_key;  // State to indicate if public key is received/added
-  bool sv_test_on;  // State to indicate if public key is received/added
 
   // Handle for vendor specific data. Only works with one vendor.
   void *vendor_handle;
   // Vendor encoders for signing. Only works with one vendor.
   const sv_tlv_tag_t *vendor_encoders;
   size_t num_vendor_encoders;
+
+  // For tests
```
bjornvolcker, line 183: I think you should elaborate a little bit more than simply saying "For tests"

lib/src/signed_video_internal.h
```diff
@@ -173,13 +173,15 @@ struct _signed_video_t {
...
+  bool sv_test_on;  // State to indicate if public key is received/added
```
bjornvolcker, line 184: The comment is incorrect. Rewrite and explain what this flag can be used for.

tests/check/check_h26xsigned_auth.c
```diff
@@ -2098,32 +2099,32 @@ START_TEST(no_emulation_prevention_bytes)
...
+  memcpy(sei_p, sei, 4);
   sei_p += 4;  // Move past the start code to avoid an incorrect emulation prevention byte.
-  char *src = (char *)(nalu_data + 4);
-  size_t src_size = data_size - 4;
+  char *src = (char *)(sei + 4);
+  size_t src_size = sei_size - 4;
   write_byte_many(&sei_p, src, src_size, &last_two_bytes, true);
-  size_t sei_data_size = sei_p - sei_data;
-  signed_video_nalu_data_free(nalu_data);
+  size_t sei_data_size = sei_p - sei_with_epb;
```
bjornvolcker, line 2119: Rename to `sei_with_epb_size`

tests/check/check_h26xsigned_sign.c
```diff
@@ -600,6 +600,7 @@ START_TEST(two_completed_seis_pending)
   size_t sei_size_2 = 0;
   size_t sei_size_3 = 0;
   signed_video_t *sv = signed_video_create(codec);
+  // Enable testing mode to add multiple SEIs.
```
bjornvolcker, line 603: Implicitly this is correct, but there is no such mode in the code. Rephrase this. For example that this will bypass a check so multiple NAL Units can be added before getting SEIs

tests/check/check_h26xsigned_sign.c
```diff
@@ -636,6 +639,8 @@ START_TEST(two_completed_seis_pending)
   // Check if there are no more SEIs remaining in the buffer
   ck_assert_int_eq(sei_size_3, 0);
   // Verify the transfer order of NALUs
+  // Expect |sei_size_1| to be less than |sei_size_2|
+  // as the first SEI is at least 32 bytes smaller than the second.
```
bjornvolcker, line 643: This is not the correct reason. The reason is that the second SEI includes more one more hash.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_sign.c
mithydolphin: If I call it before the get_latest_sei I would also reset the |sei_size|.
Should I reset it  before we get the |sei_size|?

lib/src/signed_video_h26x_sign.c
mithydolphin: Would it work if I call it  while I am checking  if the |sei_size != 0|?

lib/src/signed_video_h26x_sign.c
mithydolphin: done

lib/src/includes/signed_video_sign.h
mithydolphin: *     while (status == SV_OK && sei_size != 0) {
 *       uint8_t *sei = malloc(sei_size);
 *       status = signed_video_get_sei(sv, sei, &sei_size);
 *       if (status != SV_OK) {
 *          // True error. Handle it properly.
 *          }
 *       status = signed_video_get_sei(sv, NULL, &sei_size); 
 *       }
 For it to work I need to update  the sei_size  every iteration to  allocate  the write size for the next sei .
 Am I missing something?

lib/src/signed_video_internal.h
bjornvolcker: I think it is better to have one single generic flag like we had from the beginning `sv_test_on` or `test_override`.
This flag can then be used to enable/disable other parts if there are needs for that later. There should not be multiple flags for each test functionality. That can easily get out of hands and introduce bugs. In fact, if this should be done properly, this flag should not be possible to set if we build in release mode.

Then you should explain what type of functionality and/or sanity checks that is affected if the flag is turned on.

lib/src/signed_video_h26x_sign.c
bjornvolcker: Yes. When you have outputs returned through a pointer like here you need to reset the content before doing anything. It is good practice to do that. For example, if the pointer garbage data when passed in and some parts are not changed the user may read the data in this pointer and make conclusions based on garbage.

lib/src/includes/signed_video_sign.h
bjornvolcker: The code is there in `pull_nalus()` in the test helper function

bjornvolcker (changes requested):
There are a few old remaining review comments to address and a few new one.

lib/src/includes/signed_video_sign.h
```diff
@@ -260,16 +261,19 @@ signed_video_get_nalu_to_prepend(signed_video_t *self,
...
- *         num_nalus_to_pull--;
- *         if (num_nalus_to_pull == 0) break;
- *         sv_rc = signed_video_get_sei(sv, NULL, &sei_size);
+ *         status = signed_video_get_sei(sv, sei, &sei_size);
+ *         if (status != SV_OK) {
+ *          // True error. Handle it properly.
+ *         }
+ *         // The user is responsible for freeing |sei|.
+ *         status = signed_video_get_sei(sv, NULL, &sei_size);
+ *         if (status != SV_OK) {
```
bjornvolcker, line 274: This check is unnecessary since it is done in the while statement

bjornvolcker (changes requested):
There is still some unresolved old review comments.

bjornvolcker (changes requested):
There are still 3 unresolved comments that I think you should fix.

bjornvolcker (approved):

PR #132, opened 2024-03-20
Adds new APIs to signed_video_interfaces.h

bjornvolcker:
as part of a refactoring work.
The old APIs are commented with TO BE DEPRECATED and the plugin init
and exit is redefined, since it is not used by the library.

Implementations to the unthreaded plugin is made in this commit
and the threaded plugin will be updated in a separate one before
removing the deprecated APIs.

A new function to get the maximum signature size from the private key
has been added and the openssl_sign_hash() no longer relies on the,
by the user, set signing algo and instead reads it from the private
key.

The library still uses the, to be deprecated, calls until the
threaded plugin has been updated.


mithydolphin (approved):
Looks good to me !!

PR #135, opened 2024-04-08
Removes uses of OPENSSL_free

bjornvolcker:
since it is simply a macro that adds debug info __FUNC__ and
__LINE__ to a normal free() call.
The code now use free() throughout.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #136, opened 2024-04-09
Splits includes/signed_video_interfaces.h

bjornvolcker:
into a new public header file (signed_video_signing_plugin.h) with
all new/renamed APIs to better reflect their usage.
Further, moves the sign_algo_t and signature_info_t to
signed_video_openssl.h since it is related to the cryptographic
parts used by the wrappers in signed_video_openssl.

All old APIs in signed_video_interfaces.h have been removed and
the only remaining code is the redefinition of the plugin init/exit
APIs. These need to be kept since they are not used by the library,
but by the user.


mithydolphin (approved):

lib/plugins/threaded-signing/plugin.c
```diff
@@ -24,21 +24,21 @@
...
  * The thread is stopped if |out| is full, if there was a failure in the memory allocation
- * for a new signature or if sv_interface_teardown() is called.
+ * for a new signature or if sv_signing_plugin_session_teardown() is called.
  *
- * If the plugin is initialized, sv_interface_init(), one single central thread is spawned. Each
- * Signed Video session will then get an id to distiguish between tehm since they use common input
- * and output buffers. The thread is stopped if |out| is full, if there was a failure in
- * the memory allocation for a new signature or if sv_interface_exit() is called.
+ * If the plugin is initialized, sv_signing_plugin_init(), one single central thread is spawned.
+ * Each Signed Video session will then get an id to distiguish between tehm since they use common
```
lusikamalo, line 30: nit: tehm->them

lusikamalo (changes requested):
I think the first sentence in the commit message needs to be rephrased.

Also it could be worth mentioning in the commit message that it is deprecated interfaces that is removed.

lusikamalo (approved):
OK.

PR #137, opened 2024-04-09
Removes usage of sign_algo_t in code

bjornvolcker:
More precisely, removes usage of the algo member in
signature_info_t since that information can be extracted from the
key pair. The member itself is NOT removed due to backward
compatibility.

A new function for setting the private key has been added
temporarily until the old one is deprecated.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #138, opened 2024-04-11
Stores the private key in EVP_PKEY format

bjornvolcker:
instead of PEM format. This saves some operations when signing a
hash, since the EVP_PKEY otherwise has to be created from the PEM
format for every signature.

At the same time as the conversion is done, memory for the largest
possible signature is allocated.

A new struct pem_pkey_t to store a private or public key in PEM
format has been added. This struct is used in the new version of
threaded signing plugin at init (optional by user).

Further, two new openssl helpers are added openssl_free_key() and
openssl_private_key_malloc(), where the latter more or less
replaces openssl_signature_malloc() (now removed).

### DIFF
--- a/lib/src/signed_video_openssl.c
+++ b/lib/src/signed_video_openssl.c
@@ -111,8 +111,6 @@ openssl_private_key_malloc(signature_info_t *signature_info,
   EVP_PKEY *signing_key = NULL;
   svi_rc status = SVI_UNKNOWN;
   SVI_TRY()
-    SVI_THROW_IF(!private_key, SVI_INVALID_PARAMETER);
-
     // Read private key
     BIO *bp = BIO_new_mem_buf(private_key, private_key_size);
     signing_key = PEM_read_bio_PrivateKey(bp, NULL, NULL, NULL);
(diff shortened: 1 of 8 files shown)


lib/src/signed_video_openssl.c
```diff
@@ -90,33 +90,46 @@ openssl_create_private_key(sign_algo_t algo, const char *path_to_key, key_paths_
...
-  if (!signature_info) return SV_INVALID_PARAMETER;
+  if (!signature_info || !private_key || private_key_size == 0) return SV_INVALID_PARAMETER;
 
-  const void *private_key = signature_info->private_key;
-  int private_key_size = (int)signature_info->private_key_size;
+  EVP_PKEY *signing_key = NULL;
   svi_rc status = SVI_UNKNOWN;
   SVI_TRY()
-    SVI_THROW_IF(!private_key || private_key_size == 0, SV_INVALID_PARAMETER);
+    SVI_THROW_IF(!private_key, SVI_INVALID_PARAMETER);
```
lusikamalo, line 114: Check of "!private_key" is already performed at line 109.

bjornvolcker: Correct, and it is actually removed in another PR. I'll fix it in this PR to make this one more correct

lusikamalo (approved):
OK.

mithydolphin (approved):

PR #139, opened 2024-04-12
Stores the public key in EVP_PKEY format

bjornvolcker:
instead of PEM format in the signature_info struct. This saves some
operations when verifying a signature, since the EVP_PKEY otherwise
has to be created from the PEM format for every verification.

The PEM format public key is intermediately stored in the pem_public_key
member of signed_video_t since the PEM format is written to/read from
the SEIs.

Further, a new openssl helpers are added openssl_public_key_malloc(),
where the latter turns a PEM public key into an EVP_PKEY object.


lusikamalo (dismissed):
OK.

lusikamalo (approved):
OK.

PR #140, opened 2024-04-15
Uses EVP_PKEY_CTX instead of EVP_PKEY

bjornvolcker:
To further reduce number of openssl operation the lib keeps
the private and public keys in an algorithm context which
can be used directly when signing or verifying.


lusikamalo (approved):
LGTM.

PR #141, opened 2024-04-16
Removes unused functions

bjornvolcker:
This commit removes two functions; openssl_key_memory_allocated()
and openssl_get_algo_of_public_key(). It is safe to remove
the first one from the public header file since it by nature is
a helper function.


lusikamalo (approved):
OK.

PR #142, opened 2024-04-16
Moves some openssl declarations to internal

bjornvolcker:
The public header file signed_video_openssl.h is primarily meant
for signing plugins + signing users that need a helper function
to generate a software private key for testing. Everything else
are APIs used by the library, hence can/should be placed in an
internal header file.


lusikamalo (approved):
OK.

PR #144, opened 2024-04-18
Prepares for arbitrary hash algorithm

bjornvolcker:
To be able to support arbitrary hash algorithms when hashing NAL
Units, the library needs to use EVP_MD_CTX. If so, it is
possible to set the EVP_MD to use, for example, EVP_sha256().

The context is stored in the |crypto_handle|. Separate calls for
openssl 3.0 or openssl 1.1 was used, but the openssl 3.0
functions already existed in openssl 1.1, hence they can be used
for both. This change is made in this commit.

To use the context, every hash needs to do init/update/finalize
instead of the short-cut function SHA256(). Using short-cut
functions will not scale to arbitrary hashing algorithms.
Some internal functions needed the |crypto_handle| as input and
that change is made in this commit.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #145, opened 2024-04-24
Adds a setter for hash algorithm

bjornvolcker:
To customize usage of hash algorithms a new setter is added.
This new function sets the hashing algorithm, used to hash NAL
Units, given a name or OID that OpenSSL can interpret. If the
hash algorithm is not set, SHA256 is used.

A change of hash algorithm cannot be applied once the session has
started hashing NAL Units. Therefore a new member
|signing_started| has been added.

Basic signing tests is added.

NOTE that the validation side is not updated. The library still
assumes the video is hashed with SHA256. Transferring the
knowledge of which hash algorithm is in use is done in separate
commits.

### DIFF
--- a/.github/workflows/meson-build.yml
+++ b/.github/workflows/meson-build.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Update system
-      run: sudo apt-get update
+      run: sudo apt-get update || true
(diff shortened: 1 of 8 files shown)


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #146, opened 2024-04-26
Adds a crypto TLV tag

bjornvolcker:
A new TLV tag CRYPTO_INFO_TAG is created which is associated
with crypto information like the hashing algorithm used.
At this point the length of the data is assumed to be at most
255 bytes, hence one byte is enough to represent the length.

Both encode and decode functions are added and helper functions
are added to read and write the OID on serialized form, i.e.,
ASN.1/DER format.

A new struct to keep a message digest in both type and OID is
added. This will in particular help when a message digest for
the signature is added later.


lib/src/signed_video_openssl.c
```diff
@@ -575,19 +644,66 @@ openssl_set_hash_algo(void *handle, const char *name_or_oid)
...
+  return status;
+}
+
+svi_rc
+openssl_set_hash_algo_by_encoded_oid(void *handle,
+    const unsigned char *encoded_oid,
+    size_t encoded_oid_size)
+{
+  openssl_crypto_t *self = (openssl_crypto_t *)handle;
+  if (!self || !encoded_oid || !encoded_oid_size) return SVI_INVALID_PARAMETER;
```
lusikamalo, line 672: nit: I have seen that size is checked differently in other parts of the code. "!encoded_oid_size" works fine but check this out so it's consistent in the code.

lib/src/signed_video_openssl.c
```diff
@@ -575,19 +644,66 @@ openssl_set_hash_algo(void *handle, const char *name_or_oid)
...
+  SVI_DONE(status)
+
+  return status;
+}
+
+const unsigned char *
+openssl_get_hash_algo_encoded_oid(void *handle, size_t *encoded_oid_size)
+{
+  openssl_crypto_t *self = (openssl_crypto_t *)handle;
+  if (!self || !encoded_oid_size) return NULL;
```
lusikamalo, line 703: nit:Same as comment above. I have no problem to keep size check as it is.

lusikamalo (approved):
OK. You can skip my comments if you want.
LGTM.

mithydolphin (dismissed):

bjornvolcker:
I will make suggested changes. I think it is good practice to distinguish between pointers and integers by using "!" vs "!= 0"

lusikamalo (approved):
OK.

PR #147, opened 2024-04-29
First preparation for arbitrary hash algo

bjornvolcker:
As a first step to support other hashing algorithms than SHA256
a split of the macro HASH_DIGEST_SIZE into a MAX_HASH_SIZE and
SHA256_HASH_SIZE is done. One is used for allocating enough
memory and the other to control the actual hash size (currently
fixed to SHA256_HASH_SIZE).

A new setting member (hash_algo_name) has been added as well as
an extra setting covering non-custom configurations.

### DIFF
--- a/lib/src/signed_video_internal.h
+++ b/lib/src/signed_video_internal.h
@@ -45,8 +45,10 @@ typedef struct _h26x_nalu_t h26x_nalu_t;
 #else
 #define ATTR_UNUSED __attribute__((unused))
 #endif
-// Currently only support SHA-256 which produces hashes of size 256 bits.
-#define HASH_DIGEST_SIZE (256 / 8)
+// Currently the SHA-512 produces maximum hashe sizes.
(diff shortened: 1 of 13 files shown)


lib/src/signed_video_internal.h
```diff
@@ -45,8 +45,10 @@ typedef struct _h26x_nalu_t h26x_nalu_t;
 #else
 #define ATTR_UNUSED __attribute__((unused))
 #endif
-// Currently only support SHA-256 which produces hashes of size 256 bits.
-#define HASH_DIGEST_SIZE (256 / 8)
+// Currently the SHA-512 produces maximum hashe sizes.
```
lusikamalo, line 48: nit: hashe->hash

lusikamalo (dismissed):
Skip my "nit" comment if you want.

lusikamalo (approved):
OK.

PR #148, opened 2024-04-29
Fix setting hash algo

bjornvolcker:
This commit makes sure to assign the new hash algo to the message
digest context.
Further, does not free and create new contexts for each hash.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #149, opened 2024-04-30
Adds hash size getter

bjornvolcker:
Since the hash algorithm can be arbitrary there is a need to know
the hash size when for example allocating memory.
A getter for this has been added and the size is extracted from
the EVP_MD when a hash algo is set.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #150, opened 2024-04-30
Signing side can handle arbitrary hashes

bjornvolcker:
The signing side of the library is now set up to handle any type
of hash. Only necessary changes to compile has been made to the
validation side. Complete functionality is still only guaranteed
for default hash algorithm (SHA256).


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #151, opened 2024-04-30
Solves validation side with arbitrary hash algo

bjornvolcker:
The hash algorithm is not known until the first SEI arrives.
Therefore, the hashable data is copied and kept until the
hash algo is known. When the first SEI arrives, all stored
NALUs are re-run and hashed.

The side effect is that when the video is not signed the
memory impact becomes larger.

Further, the hash_and_add_for_auth() function now takes
a list item rather than the h26x_nalu_t struct as input.


lusikamalo (dismissed):
OK.

lusikamalo (approved):
OK.

mithydolphin (approved):

PR #153, opened 2024-05-02
Changes tests to use new key generation APIs

bjornvolcker:
The tests now take a function pointer as input instead of
sign_algo_t. The test scope has been reduced as well. RSA
keys have been moved to "Special cases".


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #154, opened 2024-05-02
Removes key_paths_t

bjornvolcker:
There is no gain in using a struct for key paths and buffers.
Instead use the pem_pkey_t struct and separate the path.


lusikamalo (dismissed):
OK.

mithydolphin (dismissed):

mithydolphin (approved):

lusikamalo (approved):
OK.

PR #158, opened 2024-05-03
Adds a new struct to replace signature_info_t

bjornvolcker:
To add a clearer separation between the signing side and the
validation side in the code the signature_info_t is split
into one siging_info_t part and one for validation.
This commit address the signing part only and a new struct
signing_info_t is added, which holds the hash to sign, the
private key and the generated signature.

Three functions now use this new struct openssl_sign_hash(),
openssl_private_key_malloc() and
openssl_read_pubkey_from_private_key().

The change affects the signing plugins in particular. The src
code still has a member signature_info, which is used and the
necessary parts have been copied from signing_info for now.

### DIFF
--- a/lib/plugins/README.md
+++ b/lib/plugins/README.md
@@ -20,8 +20,8 @@ signing plugin, hence should preferably be built with the unthreaded one.
 
 ## Threaded plugin
 The threaded plugin calls the OpenSSL signing APIs from a separate thread. If the plugin is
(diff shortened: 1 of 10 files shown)


mithydolphin (dismissed):

lusikamalo (approved):
OK.

mithydolphin (approved):

lusikamalo (approved):
OK.

PR #161, opened 2024-05-06
Removes last dependency on signature_info_t in openssl

bjornvolcker:
The last function, openssl_public_key_malloc(), now takes the
new sign_or_verify_data_t as input.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #163, opened 2024-05-07
Adds hash size as input to hash functions

bjornvolcker:
The hash size has more or less been implicitly used when
hash function wrappers are used. They now take a size as
input.


lusikamalo (approved):
OK.

PR #165, opened 2024-05-07
Removes last dependency on signature_info_t in code

bjornvolcker:
Adds a verify_data member for storing the public key etc
for verifying the signature.
Renames sign_data_create and sign_data_free, since it now also
is used by verify_data. The free function now frees the key as
well, which was not needed for sign_data, since it is handled
by the signing plugin.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #166, opened 2024-05-08
Renames nalu_list to test_stream

bjornvolcker:
To be more clear between the nalu list in src and in tests this
commit renames the one in tests to test_stream. That should
better reflect that the linked list mimics a video stream.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #168, opened 2024-05-08
Updates test_stream_item_check_type()

bjornvolcker:
Now takes a single character as type instead of a string.
Related functions in signed_video_helper.s has been updated
as well.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #169, opened 2024-05-08
Makes some functions static

bjornvolcker:
Some functions are not used outside test_stream.c. These
have been made static instead. The print functions are still
declared in the header file for manual debug purposes.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #176, opened 2024-05-14
Support only OpenSSL 3

bjornvolcker:
OpenSSL version older than 3.0.0 has been deprecated. Therefore,
all code for 1.1.1 is removed. The validation side has not been
limited by the older version, so old recordings are still
backwards compatible.


lusikamalo (approved):
OK.

PR #184, opened 2024-05-17
Removes internal return code SVI_NULL_PTR

bjornvolcker:
This internal return code is not exposed publicly and rarely
used. Replaced with SVI_INVALID_PARAMETER.


lusikamalo (approved):
OK.

lusikamalo (approved):
OK.

mithydolphin (approved):

PR #185, opened 2024-05-17
Removes SVI_FILE

bjornvolcker:
This internal return code is only used once and has
now been replaced with SVI_EXTERNAL_FAILURE.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #189, opened 2024-05-20
enable_golden_sei_principle

mithydolphin:
While using the principle, after the golden SEI is generated, the rest
of the SEIs only need to carry mandatory information. A new test is added
to confirm the principle.

### Describe your changes

- Fixes #(issue)


lib/src/includes/signed_video_sign.h
```diff
@@ -499,6 +499,22 @@ signed_video_set_recurrence_interval_frames(signed_video_t *self, unsigned recur
 SignedVideoReturnCode
 signed_video_set_sei_epb(signed_video_t *self, bool sei_epb);
 
+/**
+ * @brief Configures Signed Video use golden SEI principle
+ *
+ * The principle of the golden sei sends all information only needed once such as the Public key at
+ * the start of the stream with golden sei after that rest of the SEIs in the stream only include
+ * mandatory information.
```
bjornvolcker, line 507: You should mention that it is the user's responsibility to add the golden SEI at the beginning of the stream. To be more precise, it does not have to start with a golden SEI, the only thing that matters is that the first SEI should be a golden SEI.

Also, I don't think it should be feasible to set this on an ongoing session. That has to be clarified in the description and fixed in the code + tests.

tests/check/check_signed_video_sign.c
```diff
@@ -721,6 +721,81 @@ START_TEST(golden_sei_created)
 }
 END_TEST
 
+/* Test description
+ * Generates SEIs using golden SEI prinsiple and verifies them.
+ * The operation is as follows:
+ * 1. Setup a signed_video_t session
+ * 2. Generate and store golden SEI
+ * 3. End signed_video_t and start a new one
```
bjornvolcker, line 729: The operations 1-3 do not add any value to the tests, since it is already checked in golden_sei_created.
The operations 4-7 are independent from 1-3 and is fine to test here, but I think you should add a check that the public key is not part of the SEI.

The test you've written better fits the auth side, where you should verify that the validation is successful with the golden SEI principle.

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -2188,6 +2188,84 @@ START_TEST(with_blocked_signing)
...
+  ck_assert(sv);
+  char *private_key = NULL;
+  size_t private_key_size = 0;
+  // Setup the key
+  sv_rc = settings[_i].generate_key(NULL, &private_key, &private_key_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+
+  sv_rc = signed_video_set_private_key_new(sv, private_key, private_key_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_set_hash_algo(sv, settings[_i].hash_algo_name);
```
bjornvolcker, line 2215: Just like you apply the hash algo, you should apply other appropriate settings from `settings[_i]` which I guess is only the authenticity level

tests/check/check_signed_video_auth.c
```diff
@@ -2188,6 +2188,84 @@ START_TEST(with_blocked_signing)
...
+  ck_assert(golden_sei_size != 0);
+  uint8_t *golden_sei = malloc(golden_sei_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  // Store the |golden_sei|
+  sv_rc = signed_video_get_sei(sv, golden_sei, &golden_sei_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+
+  // End the |sv| session
+  signed_video_free(sv);
+  // Start a new stream
```
bjornvolcker, line 2231: Add an empty line to make a clear separation between the pre-start operation and the actual test stream.

tests/check/check_signed_video_auth.c
```diff
@@ -2188,6 +2188,84 @@ START_TEST(with_blocked_signing)
...
+  ck_assert_int_eq(sv_rc, SV_OK);
+  ck_assert(sei_size != 0);
+  uint8_t *sei = malloc(sei_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  sv_rc = signed_video_get_sei(sv, sei, &sei_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+
+  test_stream_item_t *golden_sei_item = test_stream_item_create(golden_sei, golden_sei_size, codec);
+  test_stream_item_t *sei_item = test_stream_item_create(sei, sei_size, codec);
+
```
bjornvolcker, line 2254: This test does not test if the golden SEI principle works. You should create a `test_stream_t` and validate it to make sure it works. You're doing operations below, that only deals with the signing side, hence should be part of the signing tests.

tests/check/check_signed_video_sign.c
```diff
@@ -721,81 +721,6 @@ START_TEST(golden_sei_created)
...
-  // Store the |golden_sei|
-  sv_rc = signed_video_get_sei(sv, golden_sei, &golden_sei_size);
-  ck_assert_int_eq(sv_rc, SV_OK);
-
-  // End the |sv| session
-  signed_video_free(sv);
-  // Start a new stream
-  sv = signed_video_create(codec);
-  ck_assert(sv);
-  sv_rc = signed_video_set_using_golden_sei(sv, true);
```
bjornvolcker: You should still test that this API fulfills what it says in the public description, that is, returns the correct error codes and cannot be used after starting a stream.

bjornvolcker (changes requested):

lib/src/includes/signed_video_sign.h
```diff
@@ -499,6 +499,29 @@ signed_video_set_recurrence_interval_frames(signed_video_t *self, unsigned recur
...
+ * be the first SEI included.
+ *
+ * Note: It is not feasible to set this on an ongoing session.
+ *
+ * @param self Session struct pointer
+ * @param using_golden_sei Flag to enable or disable the golden SEI principle.
+ *
+ * @returns SV_OK SEI |using_golden_sei| was successfully set,
+ *          SV_INVALID_PARAMETER Invalid parameter.
+ *          SV_NOT_SUPPORTED if set on the ongoing session.
```
bjornvolcker, line 520: grammar: on the -> during an

bjornvolcker, line 520: Is there a test for verifying that you get SV_NOT_SUPPORTED if set during a session?

lib/src/signed_video_h26x_auth.c
```diff
@@ -1065,6 +1065,25 @@ signed_video_add_h26x_nalu(signed_video_t *self, const uint8_t *nalu_data, size_
         DEBUG_LOG("No cryptographic information found in SEI. Using default hash algo");
         self->validation_flags.hash_algo_known = true;
       }
+      // If the NALU is golden SEI, proceed with the validation process.
+      if (nalu.is_golden_sei) {
+        prepare_for_validation(self);
```
bjornvolcker, line 1070: Remove this redundant call.

lib/src/signed_video_h26x_auth.c
```diff
@@ -1065,6 +1065,25 @@ signed_video_add_h26x_nalu(signed_video_t *self, const uint8_t *nalu_data, size_
...
+        // Check the status of the verified signature hash for the GOP info.
+        prepare_for_validation(self);
+        switch (self->gop_info->verified_signature_hash) {
```
bjornvolcker, line 1073: This switch is a copy-paste from another part of the code. Then I suggest to make a function for it.

tests/check/check_signed_video_auth.c
```diff
@@ -2202,18 +2202,23 @@ START_TEST(golden_sei_principle)
...
   SignedVideoCodec codec = settings[_i].codec;
   SignedVideoReturnCode sv_rc;
-  signed_video_t *sv = signed_video_create(codec);
-  ck_assert(sv);
   char *private_key = NULL;
   size_t private_key_size = 0;
   // Setup the key
   sv_rc = settings[_i].generate_key(NULL, &private_key, &private_key_size);
   ck_assert_int_eq(sv_rc, SV_OK);
+  // Generate golden SEI
```
bjornvolcker, line 2210: Add an empty line before this comment to make a clear distinction between the different parts of the test.

tests/check/check_signed_video_auth.c
```diff
@@ -2202,18 +2202,23 @@ START_TEST(golden_sei_principle)
```
bjornvolcker, line 2202: The text on operations 3-5 above does not match the test

lib/src/signed_video_h26x_auth.c
```diff
@@ -1065,6 +1065,25 @@ signed_video_add_h26x_nalu(signed_video_t *self, const uint8_t *nalu_data, size_
...
+          case 1:
+            nalu_list->first_item->validation_status = '.';
+            break;
+          case 0:
+            nalu_list->first_item->validation_status = 'N';
```
bjornvolcker, line 1078: There is a flaw in the code if we end up here. If the golden SEI is not authentic the library will report SV_AUTH_RESULT_SIGNATURE_PRESENT and move on.
So, if I make a fake recording and sign it with my own private key, and then replace the video content of another recording using a golden SEI, and also change the public key in that golden SEI, the verification of the golden SEI will fail, but the video will be validated as authentic since it using the tampered public key.

bjornvolcker (changes requested):

lib/src/includes/signed_video_sign.h
mithydolphin: I can add new tests in the next pull request

lib/src/signed_video_h26x_auth.c
```diff
@@ -1028,6 +1028,25 @@ reregister_nalus(signed_video_t *self)
...
+validate_golden_sei(signed_video_t *self, h26x_nalu_list_t *nalu_list)
+{
+  // Check the status of the verified signature hash for the GOP info.
+  switch (self->gop_info->verified_signature_hash) {
+    case 1:
+      nalu_list->first_item->validation_status = '.';
+      break;
+    case 0:
+      nalu_list->first_item->validation_status = 'N';
+      self->latest_validation->authenticity = SV_AUTH_RESULT_NOT_OK;
```
bjornvolcker, line 1041: I think you should add a TODO here that this result will be overwritten in maybe_validate_gop, so we can fix it in a near future.

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #194, opened 2024-05-23
Removes svi_rc sv_rc helpers

bjornvolcker:
Since the library no longer use internal return codes
mapping functions between them has been removed.


lusikamalo (approved):
OK.

PR #198, opened 2024-05-24
Organizes signed_video_create

bjornvolcker:
The signed_video_create function now initilializes members in
the same order as they appear in the object.

Also moved some resets to gop_info_reset() which in addition
has been made static.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #199, opened 2024-05-24
Removes signing_present

bjornvolcker:
The member in signed_video_t is not used anywhere, since it
is part of validation_flags.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #200, opened 2024-05-27
Removes and renames 2 members

bjornvolcker:
Renames has_gop_sei to has_sei in gop_state_t.
Removes version from gop_info_t, since it is never used.


lusikamalo (approved):
OK.

PR #204, opened 2024-06-12
Modification for Golden SEI Principle

mithydolphin:
Authentication mechanisms will be updated to align with the new golden SEI API
requirements. The existing code will be refactored, validation checks added,
error handling implemented.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -691,6 +691,52 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei)
   return status;
 }
 
+/**
+ * Decodes the SEI message, retrieves necessary parameters for authentication, and computes the hash
+ * for authenticity.
+ */
+static svrc_t
+prepare_golden_sei(signed_video_t *self)
+{
+  assert(self);
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
+  svrc_t status = SV_UNKNOWN_FAILURE;
+  sign_or_verify_data_t *verify_data = self->verify_data;
+  const size_t hash_size = verify_data->hash_size;
+
+  SV_TRY()
+    // Get the next SEI item from the NALU list.
+    h26x_nalu_list_item_t *sei = h26x_nalu_list_get_next_sei_item(nalu_list);
+
+    // If an SEI item is found and it hasn't been decoded yet, decode it.
+    if (sei && !sei->has_been_decoded) {
+      // Extract the TLV data and size from the NALU.
+      const uint8_t *tlv_data = sei->nalu->tlv_data;
+      size_t tlv_size = sei->nalu->tlv_size;
+
+      // Decode the SEI data and update the status.
+      SV_THROW(decode_sei_data(self, tlv_data, tlv_size));
+      sei->has_been_decoded = true;  // Mark the SEI as decoded.
+      if (self->gop_info->signature_hash_type == DOCUMENT_HASH) {
+        SV_THROW(hash_and_add_for_auth(self, sei));
+        memcpy(verify_data->hash, sei->hash, hash_size);
+      }
+    }
+
+    // Check if the SEI is decoded, not yet used in the GOP hash, and the signature hash type is
+    // GOP_HASH.
+    if (sei && sei->has_been_decoded && !sei->used_in_gop_hash &&
+        self->gop_info->signature_hash_type == GOP_HASH) {
+      // Compute the hash and add it for authentication.
+      SV_THROW(hash_and_add_for_auth(self, sei));
+    }
+    SV_THROW(prepare_for_validation(self));
+  SV_CATCH()
+  SV_DONE(status)
+
+  return status;
+}
+
 /* prepare_for_validation()
  *
  * 1) finds the oldest available and pending SEI in the |nalu_list|.
@@ -864,6 +910,7 @@ static bool
 validation_is_feasible(const h26x_nalu_list_item_t *item)
 {
   if (!item->nalu) return false;
+  if (item->nalu->is_golden_sei) return false;
   if (!item->nalu->is_hashable) return false;
   if (item->validation_status != 'P') return false;
 
@@ -906,6 +953,24 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
   if (!validation_feasible) {
     // If this is the first arrived SEI, but could still not validate the authenticity, signal to
     // the user that the Signed Video feature has been detected.
+    if (nalu->is_golden_sei) {
+      switch (self->gop_info->verified_signature_hash) {
+        case 1:
+          nalu_list->first_item->validation_status = '.';
+          break;
+        case 0:
+          nalu_list->first_item->validation_status = 'N';
+          self->latest_validation->authenticity = SV_AUTH_RESULT_NOT_OK;
+          self->has_public_key = false;
+          break;
+        case -1:
+        default:
+          // Got an error when verifying the gop_hash. Verify without a SEI.
+          nalu_list->first_item->validation_status = 'E';
+          self->latest_validation->authenticity = SV_AUTH_RESULT_NOT_OK;
+          self->has_public_key = false;
+      }
+    }
     if (validation_flags->is_first_sei) {
       latest->authenticity = SV_AUTH_RESULT_SIGNATURE_PRESENT;
       latest->number_of_expected_picture_nalus = -1;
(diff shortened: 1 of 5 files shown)


tests/check/check_signed_video_auth.c
```diff
@@ -2257,7 +2257,7 @@ START_TEST(golden_sei_principle)
   // Final validation is OK and all received NAL Units, but the last one, are validated.
```
bjornvolcker, line 2257: I think you should make 2 experiments.

First experiment should not affect the test result
Change lines 2254 and 2255 to
```
test_stream_item_t *first_sei = test_stream_pop_first_item(list);
test_stream_check_types(list, "IPPSIPPSIPPSI");
test_stream_append_item(list, first_sei, 1);
test_stream_check_types(list, "ISPPSIPPSIPPSI");
test_stream_append_item(list, golden_sei_item, 1);
test_stream_check_types(list, "ISSPPSIPPSIPPSI");
```

The second experiment  modifies the Golden SEI, which you can do by, right before line 2254, add
```
golden_sei_item->data[golden_sei_item->data_size - 5] = ~golden_sei_item->data[golden_sei_item->data_size - 5];
```

lib/src/signed_video_h26x_auth.c
```diff
@@ -691,6 +691,52 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei)
...
+    }
+
+    // Check if the SEI is decoded, not yet used in the GOP hash, and the signature hash type is
+    // GOP_HASH.
+    if (sei && sei->has_been_decoded && !sei->used_in_gop_hash &&
+        self->gop_info->signature_hash_type == GOP_HASH) {
+      // Compute the hash and add it for authentication.
+      SV_THROW(hash_and_add_for_auth(self, sei));
+    }
+    SV_THROW(prepare_for_validation(self));
```
bjornvolcker, line 733: If the Golden SEI is not the first NALU of the stream, the signature will not be verified. This is what will happen if signing is done in a separate thread; See line 814.

bjornvolcker (changes requested):
Try the 2 experiments mentioned as a comment of check_signed_video_auth.c

bjornvolcker (changes requested):
The overwrite of NOT_OK has not been addressed yet

tests/check/check_signed_video_auth.c
```diff
@@ -2257,7 +2259,7 @@ START_TEST(golden_sei_principle)
...
   signed_video_accumulated_validation_t final_validation = {
       SV_AUTH_RESULT_OK, false, 15, 14, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
-  // One pending NAL Unit per GOP.
+
```
bjornvolcker, line 2262: I don't remember why we skip this test if auth_level is not FRAME (line 2204).
Shouldn't it work for both cases now?

lib/src/signed_video_h26x_auth.c
```diff
@@ -691,6 +691,39 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei)
...
+prepare_golden_sei(signed_video_t *self, h26x_nalu_list_item_t *sei)
+{
+  assert(self);
+  svrc_t status = SV_UNKNOWN_FAILURE;
+  sign_or_verify_data_t *verify_data = self->verify_data;
+  const size_t hash_size = verify_data->hash_size;
+  self->gop_state.has_sei = true;
+  SV_TRY()
+    // If an SEI item is found and it hasn't been decoded yet, decode it.
+    if (sei && !sei->has_been_decoded) {
```
bjornvolcker, line 708: This should always be true if the SEI is a Golden SEI, so there is no need for an if-statement.

lib/src/signed_video_h26x_auth.c
```diff
@@ -691,6 +691,39 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei)
...
 
+/**
+ * Decodes the SEI message, retrieves necessary parameters for authentication, and computes the hash
+ * for authenticity.
+ */
+static svrc_t
+prepare_golden_sei(signed_video_t *self, h26x_nalu_list_item_t *sei)
+{
+  assert(self);
+  svrc_t status = SV_UNKNOWN_FAILURE;
```
bjornvolcker, line 702: Put this right before `SV_TRY()` with an empty line above.

lib/src/signed_video_h26x_auth.c
```diff
@@ -691,6 +691,39 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei)
...
+  sign_or_verify_data_t *verify_data = self->verify_data;
+  const size_t hash_size = verify_data->hash_size;
+  self->gop_state.has_sei = true;
```
bjornvolcker, line 705: This is used in `prepare_for_validation`, so move it to the line right before it.

lib/src/signed_video_h26x_auth.c
```diff
@@ -691,6 +691,39 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei)
...
+      // Extract the TLV data and size from the NALU.
+      const uint8_t *tlv_data = sei->nalu->tlv_data;
+      size_t tlv_size = sei->nalu->tlv_size;
+
+      // Decode the SEI data and update the status.
+      SV_THROW(decode_sei_data(self, tlv_data, tlv_size));
+      sei->has_been_decoded = true;  // Mark the SEI as decoded.
+      // Assuming the signature hash type is always DOCUMENT_HASH.
+      SV_THROW(hash_and_add_for_auth(self, sei));
+      memcpy(verify_data->hash, sei->hash, hash_size);
```
bjornvolcker, line 718: I think using `verify_data->hash_size` fits on this line, so you can remove the declaration on line 704, since it is only used here.

bjornvolcker (changes requested):

lib/src/signed_video_tlv.c
```diff
@@ -920,6 +920,7 @@ decode_crypto_info(signed_video_t *self, const uint8_t *data, size_t data_size)
     data_ptr += hash_algo_encoded_oid_size;
 
     SV_THROW_IF(data_ptr != data + data_size, SV_AUTHENTICATION_ERROR);
+    self->gop_info->nalu_hash = self->gop_info->hashes + self->verify_data->hash_size;
```
bjornvolcker, line 923: This is an unrelated bug fix and should be in a separate PR.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -907,13 +941,36 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
     // If this is the first arrived SEI, but could still not validate the authenticity, signal to
     // the user that the Signed Video feature has been detected.
     if (validation_flags->is_first_sei) {
+      // Only set authenticity to SV_AUTH_RESULT_SIGNATURE_PRESENT if it has not been marked as
+      // NOT_OK This ensures that an earlier failure is not overwritten.
       latest->authenticity = SV_AUTH_RESULT_SIGNATURE_PRESENT;
```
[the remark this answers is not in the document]
bjornvolcker: Your latest change adds no value. In fact it will mark latest to "signature present" even if it is not. What I meant was that you should **replace** the line and not **complement** it.

bjornvolcker (changes requested):
There is still one unresolved review comment, possibly due to a misunderstanding

lib/src/signed_video_h26x_auth.c
```diff
@@ -939,36 +939,36 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
...
+            latest->authenticity = SV_AUTH_RESULT_NOT_OK;
+            self->has_public_key = false;
+        }
+      } else {
+        latest->authenticity = SV_AUTH_RESULT_SIGNATURE_PRESENT;
+      }
       latest->number_of_expected_picture_nalus = -1;
       latest->number_of_received_picture_nalus = -1;
       latest->number_of_pending_picture_nalus = h26x_nalu_list_num_pending_items(nalu_list);
       latest->public_key_has_changed = false;
```
bjornvolcker, line 969: This should probably be moved into the "else" since this was updated when the Golden SEI was decoded. In general the public key should not have been changed since it is the first SEI, but it is a safety risk that if such an event happens, now overwriting it and  sending a signal that the authenticity is OK.

bjornvolcker (changes requested):
One minor thing

bjornvolcker (approved):

PR #205, opened 2024-06-13
Support multi-threaded signing for golden SEI generation

mithydolphin:
Updated the signing process to support generating golden SEIs in a multi-threaded
environment. Finalizing SEI is no longer done in the same function, allowing signing
to occur in a separate thread. Users are now responsible for ensuring SEI completion
before fetching the signature from the signing plugin.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/plugins/unthreaded-signing/plugin.c
+++ b/lib/plugins/unthreaded-signing/plugin.c
@@ -47,7 +47,6 @@ typedef struct _signature_data_t {
 
 // Plugin handle to store the signature, etc.
 typedef struct _sv_unthreaded_plugin_t {
-  bool signature_generated;
   int out_buffer_idx;
   sign_or_verify_data_t sign_data;
   signature_data_t out_buffer[MAX_BUFFER_LENGTH];  // Buffer to store signature information
@@ -55,16 +54,22 @@ typedef struct _sv_unthreaded_plugin_t {
 
 /**
  * Shifts the elements in the signature data buffer to the left.
- * This effectively removes the oldest signature from the buffer.
  */
 static void
 shift_out_buffer(sv_unthreaded_plugin_t *self)
 {
   const int idx = self->out_buffer_idx;
-  for (int j = 0; j < idx; j++) {
+  if (idx <= 0) return;
+
+  // Store the address of the oldest signature
+  uint8_t *oldest_signature = self->out_buffer[0].signature;
+
+  for (int j = 0; j < idx - 1; j++) {
     self->out_buffer[j] = self->out_buffer[j + 1];
   }
 
+  self->out_buffer[idx - 1].signature = oldest_signature;
+  self->out_buffer[idx - 1].signature_size = 0;
   self->out_buffer_idx -= 1;
 }
 
@@ -82,18 +87,11 @@ unthreaded_openssl_sign_hash(sv_unthreaded_plugin_t *self, const uint8_t *hash, 
 
   status = openssl_sign_hash(&self->sign_data);
   int idx = self->out_buffer_idx;
-  if (!self->out_buffer[idx].signature) {
-    self->out_buffer[idx].signature = calloc(1, self->sign_data.signature_size);
-    if (!self->out_buffer[idx].signature) {
-      // Failed in memory allocation.
-      status = SV_MEMORY;
-    }
-    self->out_buffer[idx].signature_size = self->sign_data.signature_size;
-  }
-  memcpy(
-      self->out_buffer[idx].signature, self->sign_data.signature, self->sign_data.signature_size);
 
-  if ((status == SV_OK) && (self->sign_data.signature_size > 0)) {
+  if (status == SV_OK && self->sign_data.signature_size > 0) {
+    memcpy(
+        self->out_buffer[idx].signature, self->sign_data.signature, self->sign_data.signature_size);
+    self->out_buffer[idx].signature_size = self->sign_data.signature_size;
     self->out_buffer_idx++;
   }
 
@@ -106,10 +104,7 @@ unthreaded_openssl_sign_hash(sv_unthreaded_plugin_t *self, const uint8_t *hash, 
 static bool
 unthreaded_openssl_has_signature(sv_unthreaded_plugin_t *self)
 {
-  if (self->out_buffer_idx > 0) {
-    return true;
-  }
-  return false;
+  return self->out_buffer_idx > 0;
 }
 
 /**
@@ -146,10 +141,9 @@ sv_signing_plugin_get_signature(void *handle,
     } else {
       memcpy(signature, self->out_buffer[0].signature, self->out_buffer[0].signature_size);
       *written_signature_size = self->out_buffer[0].signature_size;
-      free(self->out_buffer[0].signature);
+      shift_out_buffer(self);
     }
   }
-  shift_out_buffer(self);
   if (error) *error = SV_OK;
 
   return has_signature;
@@ -169,7 +163,29 @@ sv_signing_plugin_session_setup(const void *private_key, size_t private_key_size
     self = NULL;
   }
 
+  // Preallocate memory for the signatures in the buffer
+  for (int i = 0; i < MAX_BUFFER_LENGTH; i++) {
+    self->out_buffer[i].signature = calloc(1, self->sign_data.max_signature_size);
+    if (!self->out_buffer[i].signature) {
+      // Handle allocation failure
+      sv_signing_plugin_session_teardown((void *)self);
+      return NULL;
+    }
+  }
+
   return self;
+}
+
+static void
+out_buffer_teardown(sv_unthreaded_plugin_t *self)
+{
+  for (int i = 0; i < MAX_BUFFER_LENGTH; i++) {
+    if (self->out_buffer[i].signature) {
+      free(self->out_buffer[i].signature);
+      self->out_buffer[i].signature = NULL;
+    }
+    self->out_buffer[i].signature_size = 0;
+  }
 }
 
 void
@@ -177,7 +193,7 @@ sv_signing_plugin_session_teardown(void *handle)
 {
   sv_unthreaded_plugin_t *self = (sv_unthreaded_plugin_t *)handle;
   if (!self) return;
-
+  out_buffer_teardown(self);
   openssl_free_key(self->sign_data.key);
   free(self->sign_data.signature);
   free(self);
--- a/lib/plugins/unthreaded-signing/plugin.c
+++ b/lib/plugins/unthreaded-signing/plugin.c
@@ -27,6 +27,7 @@
 
 #include "includes/signed_video_openssl.h"
 #include "includes/signed_video_signing_plugin.h"
+#define MAX_BUFFER_LENGTH 60  // Maximum length of the signature buffer
 
 #ifndef ATTR_UNUSED
 #if defined(_WIN32) || defined(_WIN64)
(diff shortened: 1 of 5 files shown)


tests/check/check_signed_video_sign.c
```diff
@@ -219,11 +219,11 @@ START_TEST(api_inputs)
...
-  uint8_t *sei = malloc(sei_size);
-  sv_rc = signed_video_get_sei(NULL, sei, &sei_size);
-  ck_assert_int_eq(sv_rc, SV_INVALID_PARAMETER);
-  sv_rc = signed_video_get_sei(sv, sei, &sei_size);
-  ck_assert_int_eq(sv_rc, SV_OK);
+  // uint8_t *sei = malloc(sei_size);
+  // sv_rc = signed_video_get_sei(NULL, sei, &sei_size);
+  // ck_assert_int_eq(sv_rc, SV_INVALID_PARAMETER);
+  // sv_rc = signed_video_get_sei(sv, sei, &sei_size);
+  // ck_assert_int_eq(sv_rc, SV_OK);
```
bjornvolcker, line 226: Why have these been commented out?

tests/check/check_signed_video_sign.c
```diff
@@ -648,18 +648,18 @@ START_TEST(two_completed_seis_pending)
...
   sv_rc = signed_video_get_sei(sv, NULL, &sei_size_1);
   ck_assert_int_eq(sv_rc, SV_OK);
   ck_assert(sei_size_1 != 0);
   uint8_t *sei_1 = malloc(sei_size_1);
   ck_assert_int_eq(sv_rc, SV_OK);
   sv_rc = signed_video_get_sei(sv, sei_1, &sei_size_1);
   ck_assert_int_eq(sv_rc, SV_OK);
-  // Now get the second one.
+
+  sv_rc = signed_video_add_nalu_for_signing(sv, i_nalu_2->data, i_nalu_2->data_size);
```
bjornvolcker, line 660: This should still be at top. Otherwise there will not be a queue of signatures/SEIs

lib/src/signed_video_h26x_sign.c
```diff
@@ -615,11 +616,45 @@ SignedVideoReturnCode
...
+
+  SV_CATCH()
+  SV_DONE(status)
   if (!self || !sei_size) return SV_INVALID_PARAMETER;
   *sei_size = 0;
   if (self->num_of_completed_seis < 1) {
     DEBUG_LOG("There are no completed seis.");
-    return SV_OK;
+    status = SV_OK;
+    return status;
```
bjornvolcker, line 657: Why has this been split into two lines?

lib/src/signed_video_h26x_sign.c
```diff
@@ -615,11 +616,45 @@ SignedVideoReturnCode
 signed_video_get_sei(signed_video_t *self, uint8_t *sei, size_t *sei_size)
 {
 
+  svrc_t status = SV_UNKNOWN_FAILURE;
```
bjornvolcker, line 619: This copied code snippet is pasted in two places. You should try to turn it into a function to avoid divergence between them in the future.

lib/src/includes/signed_video_signing_plugin.h
```diff
@@ -123,7 +123,8 @@ sv_signing_plugin_init(void *user_data);
 /* Temporary function for backwards compatibility while re-interpreting |user_data|. */
 int
 sv_signing_plugin_init_new(void *user_data);
-
+void
+sv_signing_pluging_reset_signature_generated(void *handle);
```
bjornvolcker, line 127: This is declared but not defined

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -36,35 +37,81 @@
 #endif
 #endif
 
+/**
+ * Structure to store the signature information.
+ */
+typedef struct _sv_unthreaded_signature_info_t {
+  uint8_t *signature;  // The signature of the |hash|.
+  size_t signature_size;  // The size of the |signature|.
+  size_t max_signature_size;  // The allocated size of the |signature|.
```
bjornvolcker, line 46: There is no need to store max_signature_size

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -36,35 +37,81 @@
...
+} sv_unthreaded_signature_info_t;
```
bjornvolcker, line 47: This is not a public struct, hence keep the name shorter, for example `signature_data_t`

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -36,35 +37,81 @@
...
+
 // Plugin handle to store the signature, etc.
 typedef struct _sv_unthreaded_plugin_t {
   bool signature_generated;
+  int signature_data_buffer_idx;
```
bjornvolcker, line 52: Put this next to the buffer. It is fine to use a shorter name like `out_idx` and below `out_buffer`

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -36,35 +37,81 @@
...
   SignedVideoReturnCode status = SV_UNKNOWN_FAILURE;
   // Borrow the |hash| by passing the pointer to |sign_data| for signing.
+
   self->sign_data.hash = (uint8_t *)hash;
   self->sign_data.hash_size = hash_size;
 
   status = openssl_sign_hash(&self->sign_data);
-  self->signature_generated = (status == SV_OK) && (self->sign_data.signature_size > 0);
+  int idx = self->signature_data_buffer_idx;
+  if (!self->signature_data_buffer[idx].signature) {
```
bjornvolcker, line 88: Looks like you can gain a lot from a helper variable of `self->signature_data_buffer[idx]`

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -36,35 +37,81 @@
...
+    self->signature_data_buffer[idx].signature = calloc(1, self->sign_data.max_signature_size);
```
bjornvolcker, line 89: Here you might write outside memory.

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -36,35 +37,81 @@
...
+static void
+shift_signature_data_buffer(sv_unthreaded_plugin_t *self)
+{
+  const int idx = self->signature_data_buffer_idx;
+  for (int j = 0; j < idx; j++) {
+    self->signature_data_buffer[j] = self->signature_data_buffer[j + 1];
+  }
+
+  self->signature_data_buffer_idx -= 1;
+  if (self->signature_data_buffer_idx < 1) self->signature_generated = false;
```
bjornvolcker, line 71: I can't really see how this helps. `signature_generated` is not really used anymore.

bjornvolcker (changes requested):
Some initial comments. There are some flaws. See for example valgrind, since you're leaking memory and it can happen that you write outside memory.

bjornvolcker (changes requested):
You can still end up writing outside memory.

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -146,15 +141,15 @@ sv_signing_plugin_get_signature(void *handle,
   bool has_signature = unthreaded_openssl_has_signature(self);
```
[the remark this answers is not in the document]
bjornvolcker: And now when it is trivial, there is no need to have a function for it...

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -99,11 +136,12 @@ sv_signing_plugin_get_signature(void *handle,
...
     // Copy signature if there is room for it.
-    if (max_signature_size < self->sign_data.signature_size) {
+    if (max_signature_size < self->out_buffer[0].signature_size) {
       *written_signature_size = 0;
     } else {
-      memcpy(signature, self->sign_data.signature, self->sign_data.signature_size);
-      *written_signature_size = self->sign_data.signature_size;
+      memcpy(signature, self->out_buffer[0].signature, self->out_buffer[0].signature_size);
+      *written_signature_size = self->out_buffer[0].signature_size;
+      shift_out_buffer(self);
```
bjornvolcker, line 144: The shift operation should be done also in the other case, right?

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -125,15 +163,37 @@ sv_signing_plugin_session_setup(const void *private_key, size_t private_key_size
...
 
+static void
+out_buffer_teardown(sv_unthreaded_plugin_t *self)
+{
+  for (int i = 0; i < MAX_BUFFER_LENGTH; i++) {
+    if (self->out_buffer[i].signature) {
+      free(self->out_buffer[i].signature);
+      self->out_buffer[i].signature = NULL;
+    }
+    self->out_buffer[i].signature_size = 0;
```
bjornvolcker, line 187: Not needed. Remove.

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -125,15 +163,37 @@ sv_signing_plugin_session_setup(const void *private_key, size_t private_key_size
...
   return self;
 }
 
+static void
+out_buffer_teardown(sv_unthreaded_plugin_t *self)
+{
+  for (int i = 0; i < MAX_BUFFER_LENGTH; i++) {
+    if (self->out_buffer[i].signature) {
+      free(self->out_buffer[i].signature);
+      self->out_buffer[i].signature = NULL;
```
bjornvolcker, line 185: This line is not needed since the buffer teardown is done in plugin teardown only which ends all operations.

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -125,15 +163,37 @@ sv_signing_plugin_session_setup(const void *private_key, size_t private_key_size
...
+  }
+}
+
 void
 sv_signing_plugin_session_teardown(void *handle)
 {
   sv_unthreaded_plugin_t *self = (sv_unthreaded_plugin_t *)handle;
   if (!self) return;
-
+  out_buffer_teardown(self);
```
bjornvolcker, line 196: nit: Add an empty line between the if and this line

lib/src/signed_video_h26x_sign.c
```diff
@@ -53,7 +53,6 @@ static svrc_t
 prepare_for_nalus_to_prepend(signed_video_t *self);
 static void
 shift_sei_buffer_at_index(signed_video_t *self, int index);
-
```
bjornvolcker, line 56: Keep this empty line to separate declarations from definitions

tests/check/check_signed_video_sign.c
```diff
@@ -302,7 +302,7 @@ START_TEST(incorrect_operation)
   // signing plugin.
```
bjornvolcker, line 302: Update the comment now when the behavior has changed

bjornvolcker (changes requested):
Now when the get_signature has been moved is `sv_test_on` really needed?

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -182,7 +172,6 @@ out_buffer_teardown(sv_unthreaded_plugin_t *self)
   for (int i = 0; i < MAX_BUFFER_LENGTH; i++) {
     if (self->out_buffer[i].signature) {
```
bjornvolcker, line 173: No need to check for nullptr since free() can handle it

bjornvolcker (changes requested):

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -89,7 +90,7 @@ unthreaded_openssl_sign_hash(sv_unthreaded_plugin_t *self, const uint8_t *hash,
   if (status != SV_OK) return status;
   int idx = self->out_buffer_idx;
 
-  if (status == SV_OK && self->sign_data.signature_size > 0 && idx < MAX_BUFFER_LENGTH) {
+  if (self->sign_data.signature_size > 0 && idx < MAX_BUFFER_LENGTH) {
```
bjornvolcker, line 93: I think it is better practice to check availability before signing the hash. There is no point in signing if the generated signature can't be stored.

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -133,11 +132,13 @@ sv_signing_plugin_get_signature(void *handle,
 
   if (!self || !signature || !written_signature_size) return false;
 
-  bool has_signature = unthreaded_openssl_has_signature(self);
+  bool has_signature = self->out_buffer_idx > 0;
   if (has_signature) {
     // Copy signature if there is room for it.
     if (max_signature_size < self->out_buffer[0].signature_size) {
       *written_signature_size = 0;
+      shift_out_buffer(self);
```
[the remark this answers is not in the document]
bjornvolcker: And now it is not executed when entering the first if-statement...

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -85,12 +85,12 @@ unthreaded_openssl_sign_hash(sv_unthreaded_plugin_t *self, const uint8_t *hash,
   // Borrow the |hash| by passing the pointer to |sign_data| for signing.
   self->sign_data.hash = (uint8_t *)hash;
   self->sign_data.hash_size = hash_size;
-
-  status = openssl_sign_hash(&self->sign_data);
-  if (status != SV_OK) return status;
   int idx = self->out_buffer_idx;
```
bjornvolcker, line 88: What will now happen if `idx = MAX_BUFFER_LENGTH`? What kind of value is returned and is that what you want to return?

bjornvolcker (changes requested):

lib/plugins/unthreaded-signing/plugin.c
mithydolphin: This case it will return to SV_UNKNOWN_FAILURE. 
Since its a  unsupported behavior it should return SV_NOT_SUPPORTED.

lib/plugins/unthreaded-signing/plugin.c
```diff
@@ -88,6 +88,8 @@ unthreaded_openssl_sign_hash(sv_unthreaded_plugin_t *self, const uint8_t *hash,
...
   if (idx < MAX_BUFFER_LENGTH) {
     status = openssl_sign_hash(&self->sign_data);
+  } else {
+    return SV_NOT_SUPPORTED;
```
bjornvolcker, line 92: You mix a lot of returns wrapped in if-else-statements which makes it hard to read and understand what actually happens. I would have written this something like

```
int idx = self->out_buffer_idx;
if (idx >= MAX_BUFFER_LENGTH) {
  status = SV_NOT_SUPPORTED;
  goto done;
}

status = openssl_sign_hash(&self->sign_data);
if (status != SV_OK) goto done;

if (self->sign_data.signature_size == 0) {
  status = SV_NOT_SUPPORTED;
  goto done;
}

signature_data_t *out = &self->out_buffer[idx];
if (!out->signature) {
  out->signature = calloc(1, self->sign_data.max_signature_size);
  if (!out->signature) {
    status = SV_MEMORY;
    goto done;
  }
}

memcpy(out->signature, self->sign_data.signature, self->sign_data.signature_size);
out->signature_size = self->sign_data.signature_size;
self->out_buffer_idx++;

done:
return status;
```

bjornvolcker (changes requested):

bjornvolcker (changes requested):
Good work. Only one review comment left it looks like. The one on sv_test_on.

bjornvolcker (approved):
You should actually build on camera to make sure nothing is broken, but there are already quite some untested commits, so if that fails it is not necessarily this commit that is the culprit.

PR #206, opened 2024-07-04
Threaded signing

mithydolphin:
Disable test for blocked signature

The current implementation does not support this case.


bjornvolcker (changes requested):
I think a more correct way would be to

1. Make the "new" changes to the tests
```
#ifdef SIGNED_VIDEO_DEBUG
  // Verification the signature is not yet working for buffered signatures.
  return;
#endif
```
2. And then Revert the Revert

PR #209, opened 2024-07-04
Link previous GOP

mithydolphin:
Link previous GOP by adding I-frame hashes from the previous GOP in
the General tag

### Describe your changes

- Fixes #(issue)


lib/src/includes/signed_video_sign.h
```diff
@@ -577,6 +577,25 @@ signed_video_set_max_sei_payload_size(signed_video_t *self, size_t max_sei_paylo
...
+ *
+ * @param self Session struct pointer
+ * @param linked_hash_on Flag to enable or disable the linked hash.
+ *
+ * @returns SV_OK if the linked hash was successfully set,
+ *          SV_INVALID_PARAMETER if an invalid parameter was passed,
+ *          SV_NOT_SUPPORTED if set during an ongoing session.
+ */
+SignedVideoReturnCode
+signed_video_set_linked_hash(signed_video_t *self, bool linked_hash_on);
```
bjornvolcker, line 597: As discussed. It is no option for the user to select which linking method to use. Once this new feature is in, the camera will always link the previous GOP.
Hence, remove this function.

lib/src/signed_video_tlv.c
```diff
@@ -236,7 +236,9 @@ encode_general(signed_video_t *self, uint8_t *data)
   if (gop_info->has_timestamp) {
     data_size += sizeof(timestamp);
   }
-
+  if (self->linked_hash_on) {
```
bjornvolcker, line 239: As mentioned above. There is a misunderstanding of usage of the new linked hash. It should always be written. Remove this if.

lib/src/signed_video_tlv.c
```diff
@@ -278,6 +280,14 @@ encode_general(signed_video_t *self, uint8_t *data)
     write_byte(last_two_bytes, &data_ptr, (uint8_t)((timestamp)&0x000000ff), epb);
   }
 
+    // Write hash_size; 2 bytes
```
bjornvolcker, line 283: This comment is wrong. Remove.

lib/src/signed_video_tlv.c
```diff
@@ -321,6 +331,15 @@ decode_general(signed_video_t *self, const uint8_t *data, size_t data_size)
       }
     }
 
+    // Calculate the size of linked_hash_data and gop_hash_data
+    size_t remaining_size = data_size - (data_ptr - data);
+    size_t hash_size = remaining_size / 2;
```
bjornvolcker, line 336: This does not look correct. As far as I can see, only one hash is read.

bjornvolcker (changes requested):
There are some misunderstandings on this feature. Some could have been avoided if this PR would have been split into a handful smaller ones.
My suggestion is to upload a clean PR only for the TLV and continue from there.

lib/src/signed_video_tlv.c
```diff
@@ -236,9 +236,8 @@ encode_general(signed_video_t *self, uint8_t *data)
   if (gop_info->has_timestamp) {
     data_size += sizeof(timestamp);
   }
-  if (self->linked_hash_on) {
-    data_size += gop_info->linked_hash_data.hash_size;
-  }
+  data_size += gop_info->linked_hash_data.hash_size;
```
bjornvolcker, line 239: The hash size is available in `sign_data`. No need to hold a separate member for that.

lib/src/signed_video_tlv.c
```diff
@@ -278,6 +279,12 @@ encode_general(signed_video_t *self, uint8_t *data)
     write_byte(last_two_bytes, &data_ptr, (uint8_t)((timestamp)&0x000000ff), epb);
   }
 
+  for (size_t i = 0; i < gop_info->linked_hash_data.hash_size; i++) {
+    write_byte(last_two_bytes, &data_ptr, gop_info->linked_hash_data.linked_hash[i], epb);
```
bjornvolcker, line 283: Changing the byte stream requires an update of the version

bjornvolcker (changes requested):

lib/src/signed_video_h26x_sign.c
```diff
@@ -524,32 +525,48 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
-
-      SV_THROW(generate_sei_nalu(self, &payload, &payload_signature_ptr));
-      // Add |payload| to buffer. Will be picked up again when the signature has been generated.
-      add_payload_to_buffer(self, payload, payload_signature_ptr);
-      // Now we are done with the previous GOP. The gop_hash was reset right after signing and
-      // adding it to the SEI NALU. Now it is time to start a new GOP, that is, hash and add this
-      // first NALU of the GOP.
+    // Process the NALU based on the presence of linked hash and whether it is the first NALU in the
+    // GOP
+    if (!(self->gop_info->has_linked_hash && nalu.is_first_nalu_in_gop)) {
```
bjornvolcker, line 530: This flag stuff is really difficult to follow. As a first step I suggest to split the code into one if as
```
if (!self->linked_hash_on) {
  // The code as it is present today from "hash_and_add to hash_and_add"
} else {
  // Add the "new" code with your flag technique
}
```
This will make it easier to read, but most important the code becomes more robust to bugs

lib/src/signed_video_h26x_sign.c
```diff
@@ -524,32 +525,48 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
+        // Store the timestamp for the first nalu in gop.
+        if (timestamp) {
+          self->gop_info->timestamp = *timestamp;
+          self->gop_info->has_timestamp = true;
+        }
+
+        uint8_t *payload = NULL;
+        uint8_t *payload_signature_ptr = NULL;
+
+        SV_THROW(generate_sei_nalu(self, &payload, &payload_signature_ptr));
```
bjornvolcker, line 561: It looks like no SEI frames will be generated if `linked_hash_on = true`

lib/src/signed_video_h26x_common.c
```diff
@@ -962,6 +976,8 @@ hash_and_copy_to_ref(signed_video_t *self, const h26x_nalu_t *nalu, uint8_t *has
     }
     // Copy the |nalu_hash| to |reference_hash| to be used in hash_with_reference().
     memcpy(reference_hash, hash, hash_size);
+    // Copy the |hash| to |linked_hash| while linked hash prinsiple is used.
+    if (self->linked_hash_on) get_linked_hash(self, hash, hash_size);
```
bjornvolcker, line 980: When I follow the code with an example of "IPIPIP" stream only one linked hash is stored.

bjornvolcker (changes requested):
To me it does not look like the new way of linked hashes works

lib/src/signed_video_h26x_common.c
```diff
@@ -903,6 +903,20 @@ update_hash(signed_video_t *self,
   return openssl_update_hash(self->crypto_handle, hashable_data, hashable_data_size);
 }
 
+static svrc_t
+get_linked_hash(signed_video_t *self, uint8_t *hash, size_t hash_size)
```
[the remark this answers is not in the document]
bjornvolcker: And now, when you also move hashes in the queue, the operation becomes more like `update_linked_hash`

lib/src/signed_video_h26x_common.c
```diff
@@ -977,7 +974,7 @@ hash_and_copy_to_ref(signed_video_t *self, const h26x_nalu_t *nalu, uint8_t *has
     // Copy the |nalu_hash| to |reference_hash| to be used in hash_with_reference().
     memcpy(reference_hash, hash, hash_size);
     // Copy the |hash| to |linked_hash| while linked hash prinsiple is used.
-    if (self->linked_hash_on) get_linked_hash(self, hash, hash_size);
+    if (self->linked_hash_on) set_linked_hash(self, hash, hash_size);
```
bjornvolcker, line 977: Can't you always do this? The only difference is that when `linked_hash_on = false` the stored linked hashes will never be used. If you always do this operation you will execute more of the code which is better for testing and valgrind.

lib/src/signed_video_internal.h
```diff
@@ -220,6 +229,8 @@ struct _gop_info_t {
   // when SV_AUTHENTICITY_LEVEL_FRAME.
   uint8_t tmp_hash[MAX_HASH_SIZE];  // Memory for storing a temporary hash needed when a NALU is
   // split in parts.
+  linked_hash_t linked_hash_data;  // Stores linked hash data for liked hash method.
```
bjornvolcker, line 232: If we should keep the same structure as the other hashes in `gop_info_t` this should not be put in a special struct but in a consecutive memory slot. And also rename because "hash data" in general cases means the data that is hashed.
```
uint8_t linked_hashes[2 * MAX_HASH_SIZE];
```

lib/src/signed_video_internal.h
```diff
@@ -158,6 +166,7 @@ struct _signed_video_t {
 
   h26x_nalu_t *last_nalu;  // Track last parsed h26x_nalu_t to pass on to next part
 
+  linked_hash_t received_linked_hash;  // Stores linked hash data for liked hash method.
```
bjornvolcker, line 169: You only use on member of the `linked_hash_t` struct. Use
```
uint8_t received_linked_hash[MAX_HASH_SIZE];
```

lib/src/signed_video_h26x_sign.c
```diff
@@ -523,33 +524,57 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
-      // that the signing feature is present.
-
-      // Store the timestamp for the first nalu in gop.
-      if (timestamp) {
-        self->gop_info->timestamp = *timestamp;
-        self->gop_info->has_timestamp = true;
+    if (self->linked_hash_on) {
+      // Process the NALU based on the presence of linked hash and whether it is the first NALU in
+      // the GOP
+      // TODO: Every I frame triggers an SEI frame to be generated. When the NALU is an I frame and
```
bjornvolcker, line 530: The text for this TODO is not correct anymore

lib/src/signed_video_h26x_sign.c
```diff
@@ -523,33 +524,57 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
+
+        uint8_t *payload = NULL;
+        uint8_t *payload_signature_ptr = NULL;
+
+        SV_THROW(generate_sei_nalu(self, &payload, &payload_signature_ptr));
+        // Add |payload| to buffer. Will be picked up again when the signature has been generated.
+        add_payload_to_buffer(self, payload, payload_signature_ptr);
+        // The previous GOP is now completed. The gop_hash was reset right after signing and
+        // adding it to the SEI NALU. It is now time to start a new GOP, meaning the first NALU
+        // of the GOP has to be hashed and added.
```
bjornvolcker, line 548: These comments need to be updated. The least thing to do is to remove the last sentence.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_common.c
```diff
@@ -928,14 +928,18 @@ update_hash(signed_video_t *self,
 }
 
 static svrc_t
```
bjornvolcker, line 930: You should put this before `get_hash_wrapper()`. It is now placed among the different hashing schemes.

lib/src/signed_video_h26x_common.c
```diff
@@ -928,14 +928,18 @@ update_hash(signed_video_t *self,
...
-  memcpy(gop_info->linked_hash_data.linked_hash, gop_info->linked_hash_data.stored_hash,
-      self->sign_data->hash_size);
-  uint8_t *linked_hash = gop_info->linked_hash_data.stored_hash;
-  memcpy(linked_hash, hash, hash_size);
+  uint8_t *stored_hash = &gop_info->linked_hashes[hash_size];
+  uint8_t *linked_hash = &gop_info->linked_hashes[0];
+  // Move the content of stored_hash to linked_hash
+  memmove(linked_hash, stored_hash, self->sign_data->hash_size);
+  // Copy the new hash to stored_hash
+  memcpy(stored_hash, hash, self->sign_data->hash_size);
```
bjornvolcker, line 941: Here you must use `hash_size` in case it differs with `sign_data->hash_size`.
It should not differ, so you could/should probably add a sanity comparison check for it.

lib/src/signed_video_internal.h
bjornvolcker: And since you do not use the `linked_hash_t` struct anymore, you can remove it.

bjornvolcker (approved):

PR #210, opened 2024-07-05
Write NALUs in the GOP to GENERAL

mithydolphin:
GENERAL TAG: Enhance NALU processing by copying NALUs to the hash list only
if they are not SEI frames. This allows processing in one shot if we have
the complete list of NALUs or by updating an ongoing hash.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_common.c
+++ b/lib/src/signed_video_h26x_common.c
@@ -845,13 +845,13 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
 }
 
 /* hash_the_hash_list()
- * Takes all the hash nalus from |nalu_hash_list| and hash it.
+ * Takes all the hash nalus from |hash_list| and hash it.
  */
 svrc_t
 hash_the_hash_list(signed_video_t *self)
 {
   gop_info_t *gop_info = self->gop_info;
-  uint8_t *hash = gop_info->hash_of_nalu_hash_list;
+  uint8_t *hash = gop_info->computed_gop_hash;
   if (gop_info->list_idx <= 0) {
     return SV_OK;
   }
--- a/lib/src/signed_video_h26x_common.c
+++ b/lib/src/signed_video_h26x_common.c
@@ -853,8 +853,11 @@ compute_partial_gop_hash(signed_video_t *self)
   gop_info_t *gop_info = self->gop_info;
   uint8_t *hash = gop_info->computed_gop_hash;
   if (gop_info->list_idx < 0) {
-    // TODO: When list_idx is greater than 0, it means hash_list run out of memory and partial
-    // GOP hash can't be computed. In this case,instead of DOCUMENT_HASH, GOP_HASH should be used.
+    // TODO: When list_idx < 0, it indicates that there was insufficient memory allocated for the
+    // hash_list to add another hash. As a result, Signed Video will operate with a GOP level
+    // authenticity. The current implementation of the new gop_hash cannot handle this fallback
+    // scenario because it is computed from the hash_list, which is currently in a compromised
+    // state. This implementation needs to be reworked to properly handle this condition.
     return SV_OK;
   }
   if (gop_info->list_idx == 0) {
--- a/lib/src/signed_video_h26x_common.c
+++ b/lib/src/signed_video_h26x_common.c
@@ -843,24 +843,7 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
 
   return status;
 }
-/* Copies the |nalu_hash| and updates the
- * |list_idx|. Otherwise, sets the |list_idx| to -1 and proceeds.
- */
-void
-copy_hash_to_nalu_hash_list(signed_video_t *self, const uint8_t *hash, size_t hash_size)
-{
-  if (!self || !hash) return;
 
-  uint8_t *hash_list = &self->gop_info->nalu_hash_list[0];
-  int *list_idx = &self->gop_info->nalu_list_idx;
-  // Check if there is room for another hash in the |nalu_hash_list|.
-  if (*list_idx + hash_size > sizeof(self->gop_info->nalu_hash_list)) *list_idx = -1;
-  if (*list_idx >= 0) {
-    // We have a valid |hash_list| and can copy the |nalu_hash| to it.
-    memcpy(&hash_list[*list_idx], hash, hash_size);
-    *list_idx += hash_size;
-  }
-}
 /* hash_the_hash_list()
  * Takes all the hash nalus from |nalu_hash_list| and hash it.
  */
@@ -869,12 +852,12 @@ hash_the_hash_list(signed_video_t *self)
 {
   gop_info_t *gop_info = self->gop_info;
   uint8_t *hash = gop_info->hash_of_nalu_hash_list;
-  size_t hash_list_size = gop_info->nalu_list_idx;
-  if (hash_list_size == 0) {
+  if (gop_info->list_idx <= 0) {
     return SV_OK;
   }
+
   svrc_t status =
-      openssl_hash_data(self->crypto_handle, gop_info->nalu_hash_list, hash_list_size, hash);
+      openssl_hash_data(self->crypto_handle, gop_info->hash_list, gop_info->list_idx, hash);
   return status;
 }
 
@@ -1067,13 +1050,6 @@ hash_and_add(signed_video_t *self, const h26x_nalu_t *nalu)
     if (nalu->is_last_nalu_part) {
       // The end of the NALU has been reached. Update hash list and GOP hash.
       check_and_copy_hash_to_hash_list(self, nalu_hash, hash_size);
-      // Copy the NALU to the hash list if the NALU is not an SEI frame.
-      if (nalu->nalu_type == NALU_TYPE_I || nalu->nalu_type == NALU_TYPE_P) {
-        copy_hash_to_nalu_hash_list(self, nalu_hash, hash_size);
-      } else {
-        // If NALU is SEI reset the |nalu_hash_list_idx|.
-        self->gop_info->nalu_list_idx = 0;
-      }
       SV_THROW(update_gop_hash(self->crypto_handle, gop_info));
       update_num_nalus_in_gop_hash(self, nalu);
     }
--- a/lib/src/signed_video_h26x_common.c
+++ b/lib/src/signed_video_h26x_common.c
@@ -843,6 +843,40 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
 
   return status;
 }
+/* Copies the |nalu_hash| and updates the
+ * |list_idx|. Otherwise, sets the |list_idx| to -1 and proceeds.
+ */
+void
+copy_hash_to_nalu_hash_list(signed_video_t *self, const uint8_t *hash, size_t hash_size)
+{
+  if (!self || !hash) return;
+
+  uint8_t *hash_list = &self->gop_info->nalu_hash_list[0];
+  int *list_idx = &self->gop_info->nalu_list_idx;
+  // Check if there is room for another hash in the |nalu_hash_list|.
+  if (*list_idx + hash_size > sizeof(self->gop_info->nalu_hash_list)) *list_idx = -1;
+  if (*list_idx >= 0) {
+    // We have a valid |hash_list| and can copy the |nalu_hash| to it.
+    memcpy(&hash_list[*list_idx], hash, hash_size);
+    *list_idx += hash_size;
+  }
+}
+/* hash_the_hash_list()
+ * Takes all the hash nalus from |nalu_hash_list| and hash it.
+ */
+svrc_t
+hash_the_hash_list(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  uint8_t *hash = gop_info->hash_of_nalu_hash_list;
+  size_t hash_list_size = gop_info->nalu_list_idx;
+  if(hash_list_size == 0) {
+    return SV_OK;
+  }
+  gop_info->hash_size = openssl_get_hash_size(self->crypto_handle);
+  svrc_t status = openssl_hash_data(self->crypto_handle,gop_info->nalu_hash_list , hash_list_size, hash);
+  return status;
+}
 
 /* Checks if there is enough room to copy the hash. If so, copies the |nalu_hash| and updates the
  * |list_idx|. Otherwise, sets the |list_idx| to -1 and proceeds. */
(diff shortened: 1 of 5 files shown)


lib/src/signed_video_tlv.c
```diff
@@ -320,6 +322,9 @@ decode_general(signed_video_t *self, const uint8_t *data, size_t data_size)
         self->latest_validation->timestamp = gop_info->timestamp;
       }
     }
+    size_t hash_size = data_size - (data_ptr - data);
+    memcpy(self->recived_nalu_list_hash, data_ptr, hash_size);
+    data_ptr += hash_size;
```
bjornvolcker, line 327: Consider this use case:
Cameras today generate Signed Video using v1.1.29. If the user checks out signed-video-framework from master and builds the validator to validate that video. What will happen when reaching this point in the code?

lib/src/signed_video_h26x_sign.c
```diff
@@ -218,6 +218,8 @@ generate_sei_nalu(signed_video_t *self, uint8_t **payload, uint8_t **payload_sig
 
   svrc_t status = SV_UNKNOWN_FAILURE;
   SV_TRY()
+    // Hash the |nalu_hash_list| before write the computed hash to TLV.
+    SV_THROW(hash_the_hash_list(self));
```
bjornvolcker, line 222: This function is named `generate_sei_nalu()` and should only focus on generating the SEI data given the state in `self`. Hence, I suggest that this line should be moved outside.

lib/src/signed_video_internal.h
```diff
@@ -181,6 +181,7 @@ struct _signed_video_t {
   // For signature verification
   sign_or_verify_data_t *verify_data;  // All necessary information to verify a signature.
 
+  uint8_t recived_nalu_list_hash[MAX_HASH_SIZE];
```
bjornvolcker, line 184: Is this an intermediate member while implementing the "new gop hash" feature or necessary for the future?
If it is temporary, I suggest you move it to the bottom of the struct and add a TODO that it will be removed once the feature is fully implemented. Otherwise, keep it here, but with a description and rename to `received_gop_hash`

lib/src/signed_video_internal.h
```diff
@@ -208,15 +209,19 @@ struct _gop_info_t {
   uint8_t *gop_hash;  // Pointing to the memory slot of the gop_hash in |hashes|.
   uint8_t hash_list[HASH_LIST_SIZE];  // Pointer to the list of hashes used for
   // SV_AUTHENTICITY_LEVEL_FRAME.
+  uint8_t nalu_hash_list[HASH_LIST_SIZE];
```
bjornvolcker, line 212: Is this an intermediate member while implementing the "new gop hash" feature or necessary for the future?
If it is temporary, I suggest you move it to the bottom of the struct and add a TODO that it will be removed once the feature is fully implemented. Otherwise, move it to after `list_idx` and add a description and explain why there is need for a second hash list.

bjornvolcker (changes requested):
Some initial comments on TLV and members.

lib/src/signed_video_internal.h
```diff
@@ -190,11 +190,13 @@ struct _signed_video_t {
 
   signed_video_authenticity_t *authenticity;  // Pointer to the authenticity report of which results
   // will be written.
+  uint8_t received_gop_hash[MAX_HASH_SIZE];  // Received hash list after decoding SEI data while
```
bjornvolcker, line 193: Since you moved it down to the bottom I assume it will be removed in the future; See older review comment. Then I'd like to see a clear separation to the authenticity members above with a TODO to be removed.

lib/src/signed_video_internal.h
```diff
@@ -190,11 +190,13 @@ struct _signed_video_t {
...
+  // authenticating. |received_gop_hash| will be compared against |hash_list|.
```
bjornvolcker, line 194: The comment says that the `received_gop_hash` will be compared against `hash_list`, but that cannot be feasible since they have different sizes. One has the size of a single hash and the other has a size of multiple hashes.

lib/src/signed_video_internal.h
```diff
@@ -190,11 +190,13 @@ struct _signed_video_t {
...
 };
 
 typedef enum { GOP_HASH = 0, DOCUMENT_HASH = 1, NUM_HASH_TYPES } hash_type_t;
 
-/**
+/**S
```
bjornvolcker, line 199: nit: Looks like an unintended "S" sneaked in.

lib/src/signed_video_tlv.c
```diff
@@ -277,8 +277,10 @@ encode_general(signed_video_t *self, uint8_t *data)
     write_byte(last_two_bytes, &data_ptr, (uint8_t)((timestamp >> 8) & 0x000000ff), epb);
     write_byte(last_two_bytes, &data_ptr, (uint8_t)((timestamp)&0x000000ff), epb);
   }
-  for (size_t i = 0; i < gop_info->hash_size; i++) {
-    write_byte(last_two_bytes, &data_ptr, gop_info->hash_of_nalu_hash_list[i], epb);
+  if(!self->gop_hash_off) {
```
bjornvolcker, line 280: As far as I can see in the code `gop_hash_off` is always `false`, hence you always enter this if-statement. May I suggest that you remove the if-statement since it adds nothing to the code.

lib/src/signed_video_tlv.c
bjornvolcker: As far as I can see from the code `gop_hash_off` is always `false` when reaching this point. Hence, the if-statement adds nothing and validating an older video will still fail.

lib/src/signed_video_internal.h
```diff
@@ -156,6 +156,8 @@ struct _signed_video_t {
   // Frame counter and flag to handle recurrence
   bool has_recurrent_data;
   int frame_count;
+  uint8_t received_gop_hash[MAX_HASH_SIZE];  // Received hash list after decoding SEI data while
```
bjornvolcker, line 159: Put this among the other validation specific members further down. For example, after `authentication_started`

lib/src/signed_video_h26x_sign.c
```diff
@@ -296,6 +296,7 @@ generate_sei_nalu(signed_video_t *self, uint8_t **payload, uint8_t **payload_sig
     // reserved_byte = |epb|golden sei|0|0|0|0|0|0|
     uint8_t reserved_byte = self->sei_epb << 7;
     reserved_byte |= self->is_golden_sei << 6;
+    reserved_byte |= !self->gop_hash_off << 4;
```
bjornvolcker, line 299: `gop_hash_off` is only used here, hence there is no need to have a variable for it.
```
reserved_byte |= 1 << 4;  // Flag to validation side that this stream computes the GOP hash differently
```

lib/src/signed_video_h26x_common.c
```diff
@@ -1033,6 +1067,13 @@ hash_and_add(signed_video_t *self, const h26x_nalu_t *nalu)
     if (nalu->is_last_nalu_part) {
       // The end of the NALU has been reached. Update hash list and GOP hash.
       check_and_copy_hash_to_hash_list(self, nalu_hash, hash_size);
+      // Copy the NALU to the hash list if the NALU is not an SEI frame.
```
bjornvolcker, line 1070: Note that `hash_and_add()` is used on the signing side. What value is added to the code to store hashes in two places; `nalu_hash_list` and `hash_list`? Can't we use `hash_list` directly?

bjornvolcker (changes requested):

lib/src/signed_video_h26x_sign.c
mithydolphin: I was thinking it can also help to have it while validating.

lib/src/signed_video_h26x_common.c
mithydolphin: While validating  i can put the  nalu hashes to there and compute the hash of it and compare with the |received_gop_hash|

lib/src/signed_video_h26x_common.c
```diff
@@ -869,12 +852,12 @@ hash_the_hash_list(signed_video_t *self)
 {
   gop_info_t *gop_info = self->gop_info;
   uint8_t *hash = gop_info->hash_of_nalu_hash_list;
-  size_t hash_list_size = gop_info->nalu_list_idx;
-  if (hash_list_size == 0) {
+  if (gop_info->list_idx <= 0) {
```
bjornvolcker, line 855: Under what conditions will `list_idx < 0`?
Write a comment on that. Is the action to not do anything and return `SV_OK` appropriate? If not, write a TODO to fix.

lib/src/signed_video_h26x_sign.c
bjornvolcker: OK. Fair enough.

lib/src/signed_video_h26x_common.c
```diff
@@ -844,6 +844,23 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
   return status;
 }
 
+/* hash_the_hash_list()
+ * Takes all the hash nalus from |nalu_hash_list| and hash it.
```
bjornvolcker, line 848: nit: nalu_hash_list -> hash_list

lib/src/signed_video_internal.h
```diff
@@ -143,6 +143,7 @@ struct _signed_video_t {
   bool is_golden_sei;  // Flag that tells if a SEI is a golden SEI
   bool using_golden_sei;  // Flag that tells if golden SEI prinsiple is used
   bool signing_started;
+  bool gop_hash_off;  // Flag that tells if the GENERAL TAG doesn't include GOP hash.
```
bjornvolcker, line 146: Add a TODO that this will be removed after feature complete.

lib/src/signed_video_internal.h
```diff
@@ -211,10 +214,12 @@ struct _gop_info_t {
   size_t hash_list_size;  // The allowed size of the |hash_list|. This can be less than allocated.
   int list_idx;  // Pointing to next available slot in the |hash_list|. If something has gone wrong,
   // like exceeding available memory, |list_idx| = -1.
+  int nalu_list_idx;
```
bjornvolcker, line 217: Move this to right after nalu_hash_list

lib/src/signed_video_internal.h
```diff
@@ -211,10 +214,12 @@ struct _gop_info_t {
...
   uint8_t gop_hash_init;  // The initialization value for the |gop_hash|.
   uint8_t *nalu_hash;  // Pointing to the memory slot of the NALU hash in |hashes|.
   uint8_t document_hash[MAX_HASH_SIZE];  // Memory for storing the document hash to be signed
   // when SV_AUTHENTICITY_LEVEL_FRAME.
+  uint8_t hash_of_nalu_hash_list[MAX_HASH_SIZE];  // Hash of NALU hashes in GOP.
```
bjornvolcker, line 222: Since you have a member named `received_gop_hash` how about naming this `computed_gop_hash`?

lib/src/signed_video_tlv.c
```diff
@@ -236,7 +236,7 @@ encode_general(signed_video_t *self, uint8_t *data)
   if (gop_info->has_timestamp) {
     data_size += sizeof(timestamp);
   }
-
+  data_size += self->sign_data->hash_size;
```
bjornvolcker, line 239: Since you add new data to the serialization, please update the "Value fields" in the comments above.

lib/src/signed_video_tlv.c
```diff
@@ -277,6 +277,10 @@ encode_general(signed_video_t *self, uint8_t *data)
     write_byte(last_two_bytes, &data_ptr, (uint8_t)((timestamp >> 8) & 0x000000ff), epb);
     write_byte(last_two_bytes, &data_ptr, (uint8_t)((timestamp)&0x000000ff), epb);
   }
+  // Write hash of NALU hashes in GOP.
```
bjornvolcker, line 280: nit: Let's name it "GOP hash", that is, write "Write GOP hash; hash_size bytes"

bjornvolcker (changes requested):
Only minor things

lib/src/signed_video_h26x_common.c
```diff
@@ -844,6 +844,23 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
...
+{
+  gop_info_t *gop_info = self->gop_info;
+  uint8_t *hash = gop_info->computed_gop_hash;
+  if (gop_info->list_idx <= 0) {
+    return SV_OK;
+  }
+
+  svrc_t status =
+      openssl_hash_data(self->crypto_handle, gop_info->hash_list, gop_info->list_idx, hash);
+  return status;
```
bjornvolcker, line 861: You can do `return openssl_hash_data(self->crypto_handle, gop_info->hash_list, gop_info->list_idx, hash);`

lib/src/signed_video_h26x_common.c
```diff
@@ -844,6 +844,23 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
   return status;
 }
 
+/* hash_the_hash_list()
+ * Takes all the hash nalus from |hash_list| and hash it.
+ */
+svrc_t
+hash_the_hash_list(signed_video_t *self)
```
bjornvolcker, line 851: The explicit operation is "hashing the hash_list", but what you actually want to do is to compute the GOP hash. I suggest that you rename the function to `compute_gop_hash`

lib/src/signed_video_h26x_sign.c
```diff
@@ -538,6 +539,9 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
       uint8_t *payload = NULL;
       uint8_t *payload_signature_ptr = NULL;
 
+      // Hash the |nalu_hash_list| before write the computed hash to TLV.
```
bjornvolcker, line 542: This comment needs an update

bjornvolcker (changes requested):
A few minor things to fix

lib/src/signed_video_h26x_common.c
```diff
@@ -844,21 +844,22 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
   return status;
 }
 
-/* hash_the_hash_list()
+/* compute_partial_gop_hash()
  * Takes all the hash nalus from |hash_list| and hash it.
```
bjornvolcker, line 848: nit: "hash nalus" -> "NALU hashes"

lib/src/signed_video_h26x_common.c
bjornvolcker: You still haven't figured out **how** `list_idx` can be negative. Why should we return `SV_INVALID_PARAMETER` under such condition? **If** it is that simple, why not make that change at once instead of adding a TODO for it?

bjornvolcker (changes requested):
One nit and still an unresolved review comment you should address.

lib/src/signed_video_h26x_common.c
```diff
@@ -845,17 +845,20 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
...
 {
   gop_info_t *gop_info = self->gop_info;
   uint8_t *hash = gop_info->computed_gop_hash;
-  if (gop_info->list_idx <= 0) {
-    // The list index is out of bounds (either zero or negative).
-    // TODO: Handle cases where list_idx is negative by returning SV_INVALID_PARAMETER for invalid
-    // list index.
+  if (gop_info->list_idx < 0) {
+    // TODO: When list_idx is greater than 0, it means hash_list run out of memory and partial
+    // GOP hash can't be computed. In this case,instead of DOCUMENT_HASH, GOP_HASH should be used.
```
bjornvolcker, line 857: Not really correct. Write something like

TODO: When list_idx < 0 it means that there was not enough memory allocated for the hash_list to add another hash to it. The effect is that Signed Video will run with a GOP level authenticity. The current implementation of this new gop_hash cannot handle such a fallback since it is computed from the hash_list (which currently is in a bad state). This has to be reworked.

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #212, opened 2024-07-11
Adds code for validating legacy streams

bjornvolcker:
These streams have a different way of linking GOPs and computing
the GOP hash. To support all variations for backward
compatibility with the same code it would become awkwardly
complex. Therefore, if a legacy recording is detected (linked
hash or gop hash flag set), the validation is directed to
use the legacy code for that.

As a base, the code is based on master somewhere after v1.1.29.
To avoid conflicting names, the legacy code is prepended with
'legacy_'.
Only validation parts have been ported to reduce the footprint.

A new test that validates a test stream from a stored legacy test
stream data has been added.


bjornvolcker:
This is a giant pull request, mainly since it adds the same code in a legacy folder. That code can be ignored when reviewing.
Only review changes outside the **legacy/** folder

mithydolphin (approved):

PR #221, opened 2024-08-05
Change computation of GOP hash

mithydolphin:
Instead of computing GOP hash with recursive hash, it is extracted from
the signature. To validate, recompute with the stream and compare against
the received hash from the signature.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -115,6 +115,96 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
   return status;
 }
 
+/**
+ * Computes the GOP hash from |nalu_hash_list| and compares it with the GOP hash received from the
+ * signature.
+ *
+ * @param self A pointer to the signed_video_t structure.
+ * @return true if the computed GOP hash matches the received GOP hash and the status is SV_OK;
+ * false otherwise.
+ */
+static bool
+verify_gop_hash(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  const size_t hash_size = self->verify_data->hash_size;
+  svrc_t status = SV_UNKNOWN_FAILURE;
+
+  // Compute the GOP hash from |nalu_hash_list|.
+  status = compute_partial_gop_hash_for_auth(self);
+
+  // Compare the computed GOP hash with the received GOP hash.
+  bool hashes_match =
+      (memcmp(gop_info->computed_gop_hash, self->received_gop_hash, hash_size) == 0);
+
+  // Verify the GOP hash if both the hashes match and the status is SV_OK.
+  gop_info->verified_gop_hash = (hashes_match && (status == SV_OK));
+
+  return gop_info->verified_gop_hash;
+}
+
+/* Marks the NALUs that are used in GOP hash and copy the hashes of the NALUs to |nalu_hash_list|.
+ *
+ * Itarates thruogh the |nalu_list| finds the gop and mark them as |used_in_gop_hash|
+ * Hahes of marked NALUs coppied to |nalu_hash_list| to vaerify the gop hash later.
+ */
+void
+mark_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei)
+{
+  assert(self);
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
+  uint8_t *hash_list = &self->gop_info->nalu_hash_list[0];
+  int *list_idx = &self->gop_info->nalu_list_idx;
+  assert(nalu_list);
+  const size_t hash_size = self->verify_data->hash_size;
+  h26x_nalu_list_item_t *item = NULL;
+  uint8_t *hash_to_add = NULL;
+  bool found_next_gop = false;
+  bool found_item_after_sei = false;
+
+  uint8_t linked_hashes[hash_size * 2];
+  memset(linked_hashes, 0x00, sizeof(linked_hashes));  // Set all elements to 0x00
+
+  h26x_nalu_list_print(nalu_list);
+  item = nalu_list->first_item;
+  *list_idx = 0;  // Initialize list_idx to 0
+  remove_used_in_gop_hash(self->nalu_list);
+
+  while (item && !(found_next_gop || found_item_after_sei)) {
+    // If this item is not Pending, move to the next one.
+    if (item->validation_status != 'P') {
+      item = item->next;
+      continue;
+    }
+    if (item->nalu->is_first_nalu_in_gop) {
+      // If the item is a I frame hash is coppied to second sloth on memory in |linked_hashes|.
+      // Before copying memory in the second slothmoved to first sloth.
+      // TODO:This |linked_hashes| is going to be used to verifying the previos gop hash.
+      memmove(&linked_hashes[0], &linked_hashes[hash_size], hash_size);
+      memcpy(&linked_hashes[hash_size], item->hash, hash_size);
+    }
+    // Only missing items can have a null pointer |nalu|, but they are not pending.
+    assert(item->nalu);
+    found_item_after_sei = (item->prev == sei);
+    found_next_gop = (item->nalu->is_first_nalu_in_gop && !item->need_second_verification);
+
+    if (item->nalu->is_gop_sei) {
+      item = item->next;
+      continue;
+    }
+
+    hash_to_add = item->need_second_verification ? item->second_hash : item->hash;
+    if (*list_idx >= 0) {
+      // Copy the hash of the item to |hash_list| and mark as |used_in_gop_hash|
+      memcpy(&hash_list[*list_idx], hash_to_add, hash_size);
+      *list_idx += hash_size;
+      item->used_in_gop_hash = true;
+    }
+    item = item->next;
+  }
+  sei->used_in_gop_hash = true;
+}
+
 /* Verifies the hashes of the oldest pending GOP from a hash list.
  *
  * If the |document_hash| in the SEI is verified successfully with the signature and the Public key,
@@ -332,7 +422,7 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
   // Loop through the |nalu_list| and set the |validation_status| if the item is |used_in_gop_hash|
   h26x_nalu_list_item_t *item = nalu_list->first_item;
   while (item) {
-    if (item->used_in_gop_hash) {
+    if (item->used_in_gop_hash && item->validation_status == 'P') {
       // Items used in two verifications should not have |validation_status| set until it has been
       // used twice. If this is the first time we set the flag |first_verification_not_authentic|.
       if (item->second_hash && !item->need_second_verification) {
@@ -351,6 +441,82 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
   }
 
   return num_marked_items;
+}
+
+/**
+ * Verifies the GOP hash of the document.
+ *
+ * @param self A pointer to the signed_video_t structure.
+ * @param num_expected_nalus A pointer to store the number of expected NALUs.
+ * @param num_received_nalus A pointer to store the number of received NALUs.
+ * @return true if the GOP hash is verified successfully; false otherwise.
+ */
+static bool
+verify_document_gop_hash(signed_video_t *self, int *num_expected_nalus, int *num_received_nalus)
+{
+  assert(self);
+
+  int num_expected_hashes = -1;
+  int num_received_hashes = -1;
+  char validation_status = 'P';
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
+  h26x_nalu_list_item_t *sei = h26x_nalu_list_get_next_sei_item(nalu_list);
+  svrc_t status = SV_UNKNOWN_FAILURE;
+
+  // Mark all the NALUs that belong to the GOP to verify.
+  mark_gop_hash(self, sei);
+
+  // The verified_signature_hash indicates if the signature is verified.
+  // If the signature hash is verified, the GOP hash can be verified as well.
+  // If the signature hash is not verified, it means the SEI is corrupted, and the whole GOP status
+  // is determined by the verified_signature_hash.
+  switch (self->gop_info->verified_signature_hash) {
+    case 1:
+      // The signature is verified, GOP hash can be verified.
+      if (!verify_gop_hash(self)) {
+        // If the signature is verified but GOP hash is not, continue validation with the
+        // hash list if it is accessible.
+        if (self->gop_info->list_idx > 0) {
+          return verify_hashes_with_hash_list(self, num_expected_nalus, num_received_nalus);
+        }
+      }
+      validation_status = self->gop_info->verified_gop_hash ? '.' : 'N';
+      break;
+    case 0:
+      validation_status = 'N';
+      break;
+    case -1:
+    default:
+      // An error occurred when verifying the GOP hash. Verify without a SEI.
+      validation_status = 'E';
+      return verify_hashes_without_sei(self);
+  }
+
+  // The number of hashes part of the GOP hash was transmitted in the SEI.
+  num_expected_hashes = (int)self->gop_info->num_sent_nalus;
+
+  // Identify the first NALU used in the GOP hash. This will be used to add missing NALUs.
+  h26x_nalu_list_item_t *first_gop_hash_item = self->nalu_list->first_item;
+  while (first_gop_hash_item && !first_gop_hash_item->used_in_gop_hash) {
+    first_gop_hash_item = first_gop_hash_item->next;
+  }
+  num_received_hashes =
+      set_validation_status_of_items_used_in_gop_hash(self->nalu_list, validation_status);
+
+  if (!self->validation_flags.is_first_validation && first_gop_hash_item) {
+    int num_missing_nalus = num_expected_hashes - num_received_hashes;
+    const bool append = first_gop_hash_item->nalu->is_first_nalu_in_gop;
+    // No need to check the return value. A failure only affects the statistics. In the worst case
+    // we may signal SV_AUTH_RESULT_OK instead of SV_AUTH_RESULT_OK_WITH_MISSING_INFO.
+    h26x_nalu_list_add_missing(self->nalu_list, num_missing_nalus, append, first_gop_hash_item);
+  }
+
+  if (num_expected_nalus) *num_expected_nalus = num_expected_hashes;
+  if (num_received_nalus) *num_received_nalus = num_received_hashes;
+  if (status != SV_OK) {
+    return false;
+  }
+  return true;
 }
 
 /* Verifies the hashes of the oldest pending GOP from a gop_hash.
@@ -522,7 +688,7 @@ validate_authenticity(signed_video_t *self)
     verify_success = verify_hashes_without_sei(self);
   } else {
     if (self->gop_info->signature_hash_type == DOCUMENT_HASH) {
-      verify_success = verify_hashes_with_hash_list(self, &num_expected_nalus, &num_received_nalus);
+      verify_success = verify_document_gop_hash(self, &num_expected_nalus, &num_received_nalus);
     } else {
       verify_success = verify_hashes_with_gop_hash(self, &num_expected_nalus, &num_received_nalus);
     }
--- a/lib/src/signed_video_h26x_common.c
+++ b/lib/src/signed_video_h26x_common.c
@@ -864,8 +864,45 @@ compute_partial_gop_hash(signed_video_t *self)
     // The list index is zero, which is means list is empty and there is nothing to compute.
     return SV_OK;
   }
-
+#ifdef SIGNED_VIDEO_DEBUG
+  printf("size :%d \n compute_partial_gop_hash\n", gop_info->list_idx);
+  for (int i = 0; i < gop_info->list_idx; i++) {
+    printf("%02x", gop_info->hash_list[i]);
+  }
+  printf("\n");
+#endif
   return openssl_hash_data(self->crypto_handle, gop_info->hash_list, gop_info->list_idx, hash);
+}
+
+/* compute_partial_gop_hash()
+ * Takes all the NALU hashes from |nalu_hash_list| and hash it.
+ */
+svrc_t
+compute_partial_gop_hash_for_auth(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  uint8_t *hash = gop_info->computed_gop_hash;
+  if (gop_info->nalu_list_idx < 0) {
+    // TODO: When list_idx < 0, it indicates that there was insufficient memory allocated for the
+    // hash_list to add another hash. As a result, Signed Video will operate with a GOP level
+    // authenticity. The current implementation of the new gop_hash cannot handle this fallback
+    // scenario because it is computed from the hash_list, which is currently in a compromised
+    // state. This implementation needs to be reworked to properly handle this condition.
+    return SV_OK;
+  }
+  if (gop_info->nalu_list_idx == 0) {
+    // The list index is zero, which is means list is empty and there is nothing to compute.
+    return SV_OK;
+  }
+#ifdef SIGNED_VIDEO_DEBUG
+  printf("size :%d \n compute_partial_gop_hash_for_auth:\n ", gop_info->nalu_list_idx);
+  for (int i = 0; i < gop_info->nalu_list_idx; i++) {
+    printf("%02x", gop_info->nalu_hash_list[i]);
+  }
+  printf("\n");
+#endif
+  return openssl_hash_data(
+      self->crypto_handle, gop_info->nalu_hash_list, gop_info->nalu_list_idx, hash);
 }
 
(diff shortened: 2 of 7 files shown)


lib/src/signed_video_tlv.c
```diff
@@ -282,13 +281,14 @@ encode_general(signed_video_t *self, uint8_t *data)
     write_byte(last_two_bytes, &data_ptr, (uint8_t)((timestamp)&0x000000ff), epb);
   }
 
-  // Write linked hash; hash_size bytes
+   // Write GOP hash; hash_size bytes
   for (size_t i = 0; i < self->sign_data->hash_size; i++) {
-    write_byte(last_two_bytes, &data_ptr, gop_info->linked_hashes[i], epb);
+    write_byte(last_two_bytes, &data_ptr, gop_info->computed_gop_hash[i], epb);
```
bjornvolcker, line 286: Why did you change the order of linked hash and gop hash? The description above says linked hash is written before the gop hash.
You shouldn't have needed to change this file at all, since it was approved and committed separately.

lib/src/signed_video_h26x_sign.c
```diff
@@ -214,7 +214,7 @@ generate_sei_nalu(signed_video_t *self, uint8_t **payload, uint8_t **payload_sig
   // Reset |signature_hash_type| to |GOP_HASH|. If the |hash_list| is successfully added,
```
bjornvolcker, line 214: Update the comment now when the behavior has changed.

lib/src/signed_video_h26x_common.c
```diff
@@ -864,10 +864,47 @@ compute_partial_gop_hash(signed_video_t *self)
...
+  printf("\n");
+#endif
   return openssl_hash_data(self->crypto_handle, gop_info->hash_list, gop_info->list_idx, hash);
 }
 
+/* compute_partial_gop_hash()
+ * Takes all the NALU hashes from |nalu_hash_list| and hash it.
+ */
+svrc_t
+compute_partial_gop_hash_for_auth(signed_video_t *self)
```
bjornvolcker, line 881: The operation is (and should be) identical to `compute_partial_gop_hash(...)`. To make sure these functions do not deviate from each other I would like to suggest to combine them into one function as `compute_partial_gop_hash(signed_video_t *self, const uint8_t *hash_list, size_t hash_list_size)`

lib/src/signed_video_h26x_auth.c
```diff
@@ -522,7 +688,7 @@ validate_authenticity(signed_video_t *self)
     verify_success = verify_hashes_without_sei(self);
   } else {
     if (self->gop_info->signature_hash_type == DOCUMENT_HASH) {
-      verify_success = verify_hashes_with_hash_list(self, &num_expected_nalus, &num_received_nalus);
+      verify_success = verify_document_gop_hash(self, &num_expected_nalus, &num_received_nalus);
     } else {
       verify_success = verify_hashes_with_gop_hash(self, &num_expected_nalus, &num_received_nalus);
```
bjornvolcker, line 693: Can this happen at all now? I think it is good to keep it until we remove the old replaced code, but if it cannot happen you can add an `assert(false);` to catch an unexpected event.

bjornvolcker (changes requested):
A first review. Will continue with the signed_video_h26x_auth.c file soon.

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +443,82 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+
+  int num_expected_hashes = -1;
+  int num_received_hashes = -1;
+  char validation_status = 'P';
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
+  h26x_nalu_list_item_t *sei = h26x_nalu_list_get_next_sei_item(nalu_list);
+  svrc_t status = SV_UNKNOWN_FAILURE;
+
+  // Mark all the NALUs that belong to the GOP to verify.
+  mark_gop_hash(self, sei);
```
bjornvolcker, line 467: I find it a little bit weird to first mark items as used for gop_hash and later (maybe) use it for verification. Looking at what the function does it is more like adding hashes to a list, and then marks the added ones. So the name should probably be something like `add_hashes_for_computing_gop_hash`, but if so, why not compute the gop_hash at once?

The general flow is a kind of unstructured and awkward. May I suggest that you start by writing a clear function description that explains the steps you want to take to a final result. I think that will be very valuable to you when trying to solve this task.
For example, you should probably check the signature first, because if it is not OK you cannot rely on the content of the SEI.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,112 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+}
+
+/**
+ * Adds the hashes of NALUs used in computing the GOP hash and marks these NALUs as part of the GOP
+ * hash.
+ *
+ * This function iterates through the NALU list, identifies the NALUs that belong to the current
+ * GOP, and marks them as used in the GOP hash. It then copies the hashes of these NALUs to the
+ * `nalu_hash_list` for later verification of the GOP hash. Additionally, it handles specific cases
+ * such as moving linked hash values for I-frames and ensures that SEI-related NALUs are
```
bjornvolcker, line 153: To hide another operation (in this case the linked hash update) inside a function that does something else (in this case prepares for computing the gop hash) is a little bit sneaky. It can easily introduce bugs. Either separate the operations or be even more explicit in the function name.

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +459,79 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+
+  int num_expected_hashes = -1;
+  int num_received_hashes = -1;
+  char validation_status = 'P';
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
+  h26x_nalu_list_item_t *sei = h26x_nalu_list_get_next_sei_item(nalu_list);
+  svrc_t status = SV_UNKNOWN_FAILURE;
+
+  // Mark all the NALUs that belong to the GOP to verify.
+  add_hashes_for_computing_gop_hash(self, sei);
```
bjornvolcker, line 480: Wouldn't it be more intuitive to put this preparation operation in `prepare_for_validation(...)`?
And at the same time compute the gop hash. Similar to the existing `compute_gop_hash(...)`. Then we keep the same principle and avoid any potential mismatches between which NALUs that are selected in `prepare_for_validation()` and here.

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +459,79 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+
+  // The verified_signature_hash indicates if the signature is verified.
+  // If the signature hash is verified, the GOP hash can be verified as well.
+  // If the signature hash is not verified, it means the SEI is corrupted, and the whole GOP status
+  // is determined by the verified_signature_hash.
+  switch (self->gop_info->verified_signature_hash) {
```
bjornvolcker, line 486: I don't think the benefits of a `switch` is really utilized when you mix `break`,  `return` and in particular `if`. I would rewrite it with simple ifs for readability.

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,112 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+  status = compute_partial_gop_hash(
+      self, gop_info->nalu_hash_list, gop_info->nalu_list_idx, gop_info->computed_gop_hash);
+  // Compare the computed GOP hash with the received GOP hash.
+  bool hashes_match =
+      (memcmp(gop_info->computed_gop_hash, self->received_gop_hash, hash_size) == 0);
+
+  // Verify the GOP hash if both the hashes match and the status is SV_OK.
+  gop_info->successfully_verified_gop_hash = (hashes_match && (status == SV_OK));
+
+  return gop_info->successfully_verified_gop_hash;
```
bjornvolcker, line 143: Returning a set member just adds redundancy. Either use a `void` as return value or  update the member outside, or maybe you don't even need a member.

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +459,79 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+      break;
+    case -1:
+    default:
+      // An error occurred when verifying the GOP hash. Verify without a SEI.
+      validation_status = 'E';
+      return verify_hashes_without_sei(self);
+  }
+
+  // The number of hashes part of the GOP hash was transmitted in the SEI.
+  num_expected_hashes = (int)self->gop_info->num_sent_nalus;
```
bjornvolcker, line 509: If ´self->gop_info->verified_signature_hash = 0´ you cannot trust the content in the SEI, so you do not know how many NALUs that were sent.
Basically, all operations below only holds if the SEI can be verified successfully.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
mithydolphin: I have make the function boolean because I am using calling the function

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,112 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+  bool found_item_after_sei = false;
+  uint8_t linked_hashes[hash_size * 2];
+  memset(linked_hashes, 0x00, sizeof(linked_hashes));
+
+  h26x_nalu_list_print(nalu_list);
+
+  // Start with the first item in the NALU list and reset the list index.
+  item = nalu_list->first_item;
+  *list_idx = 0;
+  remove_used_in_gop_hash(self->nalu_list);  // Clear previous `used_in_gop_hash` flags.
```
bjornvolcker, line 180: Why do you need to remove previously set flags? Isn't something wrong if there are pending items that have been tagged already?

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,112 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+      memcpy(&hash_list[*list_idx], hash_to_add, hash_size);
+      *list_idx += hash_size;
+      item->used_in_gop_hash = true;  // Mark the item as used in the GOP hash
+    }
+    // Move to the next item in the NALU list
+    item = item->next;
+  }
+
+  // Finally, mark the SEI item as used in the GOP hash
+  sei->used_in_gop_hash = true;
```
bjornvolcker, line 221: But the SEI is no longer part of the GOP hash. Why do you need to mark it?
In fact, it may be marked as 'N' even if the signature was successfully verified, hence should be marked '.'

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +459,70 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+    // we may signal SV_AUTH_RESULT_OK instead of SV_AUTH_RESULT_OK_WITH_MISSING_INFO.
+    h26x_nalu_list_add_missing(self->nalu_list, num_missing_nalus, append, first_gop_hash_item);
+  }
+
+  if (num_expected_nalus) *num_expected_nalus = num_expected_hashes;
+  if (num_received_nalus) *num_received_nalus = num_received_hashes;
+  if (status != SV_OK) {
+    return false;
+  }
+  return true;
```
bjornvolcker, line 523: These last lines could be replaced by `return status == SV_OK;`

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,112 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+void
+prepare_for_verify_link_and_gop_hashes(signed_video_t *self, h26x_nalu_list_item_t *sei)
+{
+  // Ensure the `self` pointer is valid
+  assert(self);
+
+  // Initialize pointers and variables
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
+  uint8_t *hash_list = &self->gop_info->nalu_hash_list[0];
+  int *list_idx = &self->gop_info->nalu_list_idx;
```
bjornvolcker, line 165: An alternative, which in general is more robust, is to deal with a local `list_idx` and set `self->gop_info->nalu_list_idx` at the end. See for example decoders in TLV. Members should only be updated upon success.

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,112 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+  assert(nalu_list);
+  const size_t hash_size = self->verify_data->hash_size;
+  h26x_nalu_list_item_t *item = NULL;
+  uint8_t *hash_to_add = NULL;
```
bjornvolcker, line 169: `hash_to_add` should not be changed, hence use `const`.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +446,68 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
   return num_marked_items;
 }
 
+/**
+ * Verifies the integrity of the GOP hash in the video, ensuring that the data
+ * within the GOP is authentic and complete. Updates the expected and received
+ * NALU counts, and returns true if the verification is successful.
```
bjornvolcker, line 452: This description is not really the complete picture. The text focus on verifying the GOP hash, but that is just one part in the process. I think you, and the reader, would benefit from a more detailed step-by-step description.

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +446,68 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+      if (self->gop_info->list_idx > 0) {
+        return verify_hashes_with_hash_list(self, num_expected_nalus, num_received_nalus);
+      }
+      validation_status = 'N';
+    }
+  } else if (self->gop_info->verified_signature_hash == 0) {
+    validation_status = 'N';
+  } else {
+    // An error occurred when verifying the GOP hash. Verify without a SEI.
+    return verify_hashes_without_sei(self);
```
bjornvolcker, line 482: If you end up here you'll not use the set `used_in_gop_hash` flags, and instead have an independent way of marking/verifying NALUs. How can you guarantee that the "correct" NALUs are verified and not a different set?

lib/src/signed_video_h26x_auth.c
```diff
@@ -332,7 +425,7 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
   // Loop through the |nalu_list| and set the |validation_status| if the item is |used_in_gop_hash|
   h26x_nalu_list_item_t *item = nalu_list->first_item;
   while (item) {
-    if (item->used_in_gop_hash) {
+    if (item->used_in_gop_hash && item->validation_status == 'P') {
```
bjornvolcker, line 428: Now you have changed the behavior compared to the description. You're setting validation status of PENDING items used in gop hash. First question, is it necessary to add the "pending" requirement? And secondly, if so you should update the description and probably the function name as well.

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,99 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+    found_next_gop = (item->nalu->is_first_nalu_in_gop && !item->need_second_verification);
+
+    // Skip GOP SEI items as they do not contribute to the GOP hash.
+    if (item->nalu->is_gop_sei) {
+      item = item->next;
+      continue;
+    }
+
+    hash_to_add = item->need_second_verification ? item->second_hash : item->hash;
+    // Copy the hash to the list if there is enough space
```
bjornvolcker, line 194: What will happen if there is no space for the hash? Here is an example scenario:
Assume the camera generated this GOP: IPPPPIS and the nalu_list was filled up
Then someone adds one frame: IPPPPBIS

In the case where linking is done the old way you'll have these hash lists
IPPPPI vs IPPPPB
but the new way of linking will have this
IPPPP vs IPPPP

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -115,6 +115,112 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+static bool
+verify_gop_hash(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  const size_t hash_size = self->verify_data->hash_size;
+  svrc_t status = SV_UNKNOWN_FAILURE;
+
+  // Compute the GOP hash from |nalu_hash_list|.
+  status = compute_partial_gop_hash(
+      self, gop_info->nalu_hash_list, gop_info->nalu_list_idx, gop_info->computed_gop_hash);
```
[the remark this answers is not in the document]
mithydolphin: In the Fallback to gop level  pull request comming up , its handled inside |compute_partial_gop_hash|.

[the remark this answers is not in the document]
bjornvolcker: OK, but what will happen if `gop_info->nalu_list_idx = -1` and `self->received_gop_hash` has not been updated from previous validation?

lib/src/signed_video_h26x_auth.c
bjornvolcker: The added step-by-step description is pretty much a copy of the code. It is difficult to write the goal of a function like a novel, but in general a good exercise. What would you say to someone that has never seen the code but want to know a GOP gets validated?

Validation of a GOP involves a couple of steps
1. First of all the signature of the SEI has to be successfully verified for the content to be trusted. If not, the entire GOP becomes invalid.
2. Part of the SEI is a GOP hash computed on the signing side. This received GOP hash is verified against the computed GOP hash on the validation side. If the are identical, the entire GOP is valid. If not, a second try is done by verifying individual hashes. This can be done if the list of hashes is present in the SEI. If the GOP hash verification failed due to a missing NAL Unit it is still possible to get a valid GOP, but with missing items.
3. Mark all NAL Units in the GOP accordingly, maybe add missing items and extract statistics.

lib/src/signed_video_internal.h
```diff
@@ -246,6 +246,7 @@ struct _gop_info_t {
   // detected.
   int verified_signature_hash;  // Status of last hash-signature-pair verification. Has 1 for
   // success, 0 for fail, and -1 for error.
+  bool successfully_verified_gop_hash;
```
bjornvolcker, line 249: This member is not used anywhere. Remove.

lib/src/signed_video_h26x_common.c
```diff
@@ -843,29 +843,59 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
...
+svrc_t
+compute_partial_gop_hash(signed_video_t *self,
+    const uint8_t *nalu_hash_list,
+    int nalu_list_idx,
+    uint8_t *hash)
+{
+  if (nalu_list_idx < 0) {
+    return SV_OK;  // Handle insufficient memory scenario.
+  }
+  if (nalu_list_idx == 0) {
```
bjornvolcker, line 855: How about combining these two if-statement into one

lib/src/signed_video_h26x_internal.h
```diff
@@ -188,7 +188,10 @@ svrc_t
 hash_and_add_for_auth(signed_video_t *signed_video, h26x_nalu_list_item_t *item);
 
 svrc_t
-compute_partial_gop_hash(signed_video_t *self);
+compute_partial_gop_hash(signed_video_t *self,
```
bjornvolcker, line 191: It looks like `self` is not updated. Make `const`.

lib/src/signed_video_h26x_common.c
```diff
@@ -843,29 +843,59 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
...
+  return openssl_hash_data(self->crypto_handle, nalu_hash_list, nalu_list_idx, hash);
+}
 
 /* compute_partial_gop_hash()
- * Takes all the NALU hashes from |hash_list| and hash it.
+ * Takes all the NALU hashes from |nalu_hash_list| and hash it.
  */
 svrc_t
-compute_partial_gop_hash(signed_video_t *self)
+compute_partial_gop_hash_for_auth(signed_video_t *self)
```
bjornvolcker, line 874: It looks like `self` is not updated. Make `const`

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -522,8 +704,9 @@ validate_authenticity(signed_video_t *self)
     verify_success = verify_hashes_without_sei(self);
   } else {
     if (self->gop_info->signature_hash_type == DOCUMENT_HASH) {
-      verify_success = verify_hashes_with_hash_list(self, &num_expected_nalus, &num_received_nalus);
+      verify_success = verify_hashes_with_sei(self, &num_expected_nalus, &num_received_nalus);
     } else {
+      assert(false);
```
bjornvolcker, line 709: Add a short comment that we should not end up here any longer

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +459,82 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+  // The verified_signature_hash indicates if the signature is verified.
+  // If the signature hash is verified, the GOP hash can be verified as well.
+  // If the signature hash is not verified, it means the SEI is corrupted, and the whole GOP status
+  // is determined by the verified_signature_hash.
+  if (self->gop_info->verified_signature_hash == 1) {
+    validation_status = '.';
+    num_expected_hashes = (int)self->gop_info->num_sent_nalus;
+    if (!verify_gop_hash(self)) {
+      // If the signature is verified but GOP hash is not, continue validation with the
+      // hash list if it is accessible.
```
bjornvolcker, line 500: grammar: "is accessible" -> "is present"

lib/src/signed_video_h26x_auth.c
```diff
@@ -353,6 +459,82 @@ set_validation_status_of_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list, cha
...
+      if (self->gop_info->list_idx > 0) {
+        return verify_hashes_with_hash_list(self, num_expected_nalus, num_received_nalus);
+      }
+      validation_status = 'N';
+    }
+  } else if (self->gop_info->verified_signature_hash == 0) {
+    validation_status = 'N';
```
bjornvolcker, line 507: Unfortunately there is no test case where a SEI has been tampered with. So it is not known if this path in the code actually works. We should add such a case in a separate PR before pushing this PR.

lib/src/signed_video_h26x_auth.c
```diff
@@ -201,7 +201,7 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
     item = item->next;
   }
 
-  if (list_idx + hash_size > self->gop_info->hash_list_size) {
+  if (list_idx + hash_size > self->gop_info->hash_list_size + 1) {
```
bjornvolcker, line 204: This will not solve the bug since you will still enter this if-statement with the example I provided. Have you tested it?

lib/src/signed_video_h26x_auth.c
```diff
@@ -192,15 +191,28 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
...
+    // and all the NALUs in the hash_list should be hashed. |list_idx| should then be set to -1.
+    list_idx = -1;
+  }
+  if (list_idx < 0) {
+    // TODO: When |list_idx| is -1, it indicates that the hash_list buffer has overflowed and the
+    // GOP hash has been initialized. From this point onward, the GOP hash needs to be updated with
+    // every new |hash_to_add|.
+  }
+
+  self->gop_info->nalu_list_idx = list_idx;
```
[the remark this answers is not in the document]
bjornvolcker: Still not solved

lib/src/signed_video_h26x_auth.c
```diff
@@ -707,6 +707,10 @@ validate_authenticity(signed_video_t *self)
       verify_success = verify_hashes_with_sei(self, &num_expected_nalus, &num_received_nalus);
     } else {
       assert(false);
```
bjornvolcker, line 709: Move the assert to after the comment

lib/src/signed_video_h26x_auth.c
bjornvolcker: Test added.

bjornvolcker (changes requested):
I think the bug is still present in your code

bjornvolcker (approved):
Approved if valgrind is happy

PR #224, opened 2024-09-05
Fix GOP hash computation failure for large GOP sizes

mithydolphin:
The previous GOP hash computation failed when the size exceeded the maximum
allowed limit, preventing successful hash generation. It is resolved by
dynamically updating the GOP hash with each incoming NALU, ensuring correct
hash generation for larger GOP sizes.

### Describe your changes

- Fixes #(issue)


tests/check/check_signed_video_sign.c
```diff
@@ -616,7 +616,7 @@ START_TEST(fallback_to_gop_level)
   // Verify that the HASH_LIST_TAG is present in the SEI when it should.
   ck_assert(tag_is_present(sei_1, settings[_i].codec, HASH_LIST_TAG));
   ck_assert(tag_is_present(sei_2, settings[_i].codec, HASH_LIST_TAG));
-  ck_assert(!tag_is_present(sei_3, settings[_i].codec, HASH_LIST_TAG));
+  //  ck_assert(!tag_is_present(sei_3, settings[_i].codec, HASH_LIST_TAG));
```
bjornvolcker, line 619: This worries me. If this fails the hash list is present in the SEI when it should not. Hence, if you need to remove this check, something is wrong in the code.

lib/src/signed_video_openssl.c
```diff
@@ -62,6 +62,7 @@ typedef struct {
  */
 typedef struct {
   EVP_MD_CTX *ctx;  // Hashing context
+  EVP_MD_CTX *f_ctx;  // Hashing context for fallback to gop level hashing.
```
bjornvolcker, line 65: I find the naming very misleading. "fallback to gop level" is a feature or an operation that may occur and has nothing to do with OpenSSL. I think you should focus on, throughout this PR, the difference between a "hash" and a "gop_hash".
So I suggest you name these context `hash_ctx` and `gop_hash_ctx`, or more generic `primary_ctx` and `secondary_ctx`, which is even better since OpenSSL is not aware of how they are used in the rest of the code.

lib/src/signed_video_openssl.c
```diff
@@ -319,51 +320,80 @@ openssl_hash_data(void *handle, const uint8_t *data, size_t data_size, uint8_t *
 
 /* Initializes EVP_MD_CTX in |handle| with |hash_algo.type|. */
 svrc_t
-openssl_init_hash(void *handle)
+openssl_init_hash(void *handle, bool fallback_to_gop_level)
```
bjornvolcker, line 323: The name `fallback_to_gop_level` id very misleading and narrow. What you do is selecting between two hash contexts, either `primary_ctx` or `secondary_ctx`. So I would flip the interpretation and name the input `bool primary`, leaving out the "ctx" part since the user of the crypto handle does not know the naming of the member.

lib/src/signed_video_openssl.c
```diff
@@ -319,51 +320,80 @@ openssl_hash_data(void *handle, const uint8_t *data, size_t data_size, uint8_t *
...
+      // Message digest type already set in context. Initialize the hashing function.
+      ret = EVP_DigestInit_ex(self->ctx, NULL, NULL);
+    } else {
+      if (!self->hash_algo.type) return SV_INVALID_PARAMETER;
+      // Create a new context and set message digest type.
+      self->ctx = EVP_MD_CTX_new();
+      if (!self->ctx) return SV_EXTERNAL_ERROR;
+      // Set a message digest type and initialize the hashing function.
+      ret = EVP_DigestInit_ex(self->ctx, self->hash_algo.type, NULL);
+    }
```
bjornvolcker, line 352: Duplicating code this way makes the code less readable and becomes very vulnerable to introducing bugs, since you need to make the same change in several places. Good coding practice is to select the context to us first and then proceed with the operation, that is,
```
EVP_MD_CTX **ctx = primary ? &self->primary_ctx : &self->secondary_ctx;
if (*ctx) {
  // Message digest type already set in context. Initialize the hashing function.
  ret = EVP_DigestInit_ex(*ctx, NULL, NULL);
} else {
  if (!self->hash_algo.type) return SV_INVALID_PARAMETER;
  // Create a new context and set message digest type.
  *ctx = EVP_MD_CTX_new();
  if (!(*ctx)) return SV_EXTERNAL_ERROR;
  // Set a message digest type and initialize the hashing function.
  ret = EVP_DigestInit_ex(*ctx, self->hash_algo.type, NULL);
}
```

lib/src/signed_video_h26x_common.c
```diff
@@ -868,7 +923,13 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
   uint8_t *hash_list = &self->gop_info->hash_list[0];
   int *list_idx = &self->gop_info->list_idx;
   // Check if there is room for another hash in the |hash_list|.
-  if (*list_idx + hash_size > self->gop_info->hash_list_size) *list_idx = -1;
+  if (*list_idx + hash_size > self->gop_info->hash_list_size) {
+    init_fallback_gop_hash(self, hash_list, *list_idx);
```
bjornvolcker, line 927: First of all, the function has a return code that you should check. Secondly, the function only has two operations which could be done here directly, hence no need for a function. Thirdly, switching to a different kind of operation in the middle of the gop hash computation process makes the code more vulnerable to errors and introduction of bugs. May I suggest to use the "secondary ctx" from the start?

bjornvolcker (changes requested):

lib/src/signed_video_h26x_common.c
```diff
@@ -867,13 +922,19 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
 
   uint8_t *hash_list = &self->gop_info->hash_list[0];
   int *list_idx = &self->gop_info->list_idx;
-  // Check if there is room for another hash in the |hash_list|.
-  if (*list_idx + hash_size > self->gop_info->hash_list_size) *list_idx = -1;
+
+  if (*list_idx + hash_size > self->gop_info->hash_list_size + 1) {
```
bjornvolcker, line 926: If `hash_size = 32`, `self->gop_info->hash_list` is declared as `uint8_t hash_list[159];`, `self->gop_info->hash_list_size = 159` and `self->gop_info->list_idx = 128`
what will happen?

bjornvolcker (changes requested):
Can you please write 2 flowcharts;
1. The first one should describe how the GOP hash ends up in the SEI starting from NALUs added for signing.
2. The second one should describe how the GOP hash, from the SEI, is verified starting from the list of NALUs. You can assume that the SEI has been verified correctly with its signature.

lib/src/signed_video_h26x_common.c
mithydolphin: if hash_size = 32, self->gop_info->hash_list_size = 159, and self->gop_info->list_idx = 128, the "*list_idx + hash_size > self->gop_info->hash_list_size + 1" returns  to false. Therefore, *list_idx is not set to -1 and  hash  can be  put  to last slot of hashlist

bjornvolcker: But then you will write 32 bytes of data when only 159 - 128 = 31 bytes are available, hence write outside memory.

private_ecdsa_key.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PRIVATE KEY-----
```
bjornvolcker, line 1: Remove this file. Should not be committed.

private_rsa_key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
bjornvolcker, line 1: Remove this file. Should not be committed.

lib/src/signed_video_openssl.c
```diff
@@ -61,8 +61,8 @@ typedef struct {
  * OpenSSL cryptographic object.
  */
 typedef struct {
-  EVP_MD_CTX *primary_ctx;  // Hashing context
-  EVP_MD_CTX *secondary_ctx;  // Hashing context for fallback to gop level hashing.
+  EVP_MD_CTX *primary_ctx;  // Hashing context for the gop level hashing.
```
bjornvolcker, line 64: It is not only used for GOP level hashing, it is also used for frame level hashing, since it is added to the SEI in both cases.

lib/src/signed_video_openssl.c
```diff
@@ -61,8 +61,8 @@ typedef struct {
...
+  EVP_MD_CTX *secondary_ctx;  // Hashing context for NALUs.
```
bjornvolcker, line 65: NALUs -> NALUs split in parts

lib/src/signed_video_h26x_auth.c
```diff
@@ -187,20 +188,18 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
...
+      list_idx = -1;
+      initialize_and_update_gop_level_hash(self, hash_list, list_idx);
     }
     // Copy the hash to the list if there is enough space
-    if (list_idx >= 0 && (list_idx + hash_size <= self->gop_info->hash_list_size) &&
-        item->nalu->is_last_nalu_part) {
-      update_fallback_gop_hash(self, hash_to_add);
-      // memcpy(&hash_list[list_idx], hash_to_add, hash_size);
+    if (list_idx >= 0) {
+      memcpy(&hash_list[list_idx], hash_to_add, hash_size);
```
bjornvolcker, line 200: Now you've added back copying hashes to a list/memory. Then you need to check for space so you do not write outside memory.

lib/src/signed_video_h26x_common.c
```diff
@@ -843,16 +843,69 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
...
+    SV_THROW(openssl_update_hash(self->crypto_handle, hash_list, idx, true));
+
+  SV_CATCH()
+  SV_DONE(status)
+
+  return status;
+}
+
+svrc_t
+update_fallback_gop_hash(signed_video_t *self, const uint8_t *nalu_hash)
```
bjornvolcker, line 865: Wrapping `openssl_update_hash(...)` in a function like this adds no value. Remove.

lib/src/signed_video_h26x_common.c
```diff
@@ -843,16 +843,69 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
 
   return status;
 }
+
+svrc_t
+initialize_and_update_gop_level_hash(signed_video_t *self, uint8_t *hash_list, int idx)
```
bjornvolcker, line 848: Wrapping `openssl_init_hash(...)` and `openssl_update_hash(...)` in a function like this adds no value. Remove.

bjornvolcker (changes requested):
I'm having a really hard time understanding the flow to compute a hash for the entire GOP. Before proceeding with the review I need some kind of flowchart so I can follow the code that's been added. I find 7 different if-statements that affects the final output.

lib/src/signed_video_openssl.c
```diff
@@ -319,48 +320,58 @@ openssl_hash_data(void *handle, const uint8_t *data, size_t data_size, uint8_t *
...
+  // Check if the context is valid.
+  if (!(*ctx)) return SV_EXTERNAL_ERROR;
+
   // Update the "ongoing" hash with new data.
-  if (!self->ctx) return SV_EXTERNAL_ERROR;
-  return EVP_DigestUpdate(self->ctx, data, data_size) == 1 ? SV_OK : SV_EXTERNAL_ERROR;
+  return EVP_DigestUpdate(*ctx, data, data_size) == 1 ? SV_OK : SV_EXTERNAL_ERROR;
 }
 
 /* Finalizes EVP_MD_CTX in |handle| and writes result to |hash|. */
```
bjornvolcker, line 359: Update description from "Finalizes EVP_MD_CTX ..." to "Finalizes a EVP_MD_CTX ..."

lib/src/signed_video_openssl.c
```diff
@@ -319,48 +320,58 @@ openssl_hash_data(void *handle, const uint8_t *data, size_t data_size, uint8_t *
...
+    if (!(*ctx)) return SV_EXTERNAL_ERROR;
     // Set a message digest type and initialize the hashing function.
-    ret = EVP_DigestInit_ex(self->ctx, self->hash_algo.type, NULL);
+    ret = EVP_DigestInit_ex(*ctx, self->hash_algo.type, NULL);
   }
-
   return ret == 1 ? SV_OK : SV_EXTERNAL_ERROR;
 }
 
 /* Updates EVP_MD_CTX in |handle| with |data|. */
```
bjornvolcker, line 343: Update description from "Updates EVP_MD_CTX ..." to "Updates a EVP_MD_CTX ..."

lib/src/signed_video_openssl.c
```diff
@@ -319,48 +320,58 @@ openssl_hash_data(void *handle, const uint8_t *data, size_t data_size, uint8_t *
 
 /* Initializes EVP_MD_CTX in |handle| with |hash_algo.type|. */
```
bjornvolcker, line 321: Update description from "Initializes EVP_MD_CTX ..." to "Initializes a EVP_MD_CTX ..."

lib/src/signed_video_h26x_internal.h
```diff
@@ -187,19 +187,22 @@ hash_and_add(signed_video_t *signed_video, const h26x_nalu_t *nalu);
...
     bool is_auth_side);
 
+svrc_t
+initialize_and_update_gop_level_hash(signed_video_t *self, uint8_t *hash_list, int idx);
+
+svrc_t
+update_fallback_gop_hash(signed_video_t *self, const uint8_t *nalu_hash);
+
+svrc_t
+finalize_gop_hash(const signed_video_t *self);
```
bjornvolcker, line 204: The definitions of these three functions have been removed. Remove also the declarations.

lib/src/signed_video_h26x_common.c
```diff
@@ -843,37 +843,35 @@ update_gop_hash(void *crypto_handle, gop_info_t *gop_info)
...
-  if (*list_idx + hash_size > self->gop_info->hash_list_size) *list_idx = -1;
+
+  // If the upcoming hash doesn't fit in the hash list buffer, set *list_idx to -1
+  // to indicate that the hash list is full, and the hash list is no longer accessible.
+  if (*list_idx + hash_size > self->gop_info->hash_list_size) {
+    *list_idx = -1;
+  }
+  // If this is the start of the GOP, initialize |crypto_handle| to enable
+  // updating the hash with each received NALU.
+  if (*list_idx == 0) {
```
bjornvolcker, line 865: For example, if `self->gop_info->hash_list_size = 32` and `hash_size = 64` the GOP hash will never be initialized. This is another example on why it is good to keep the hash list and computation of the GOP hash separated.

lib/src/signed_video_h26x_auth.c
```diff
@@ -167,56 +158,47 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
...
+      // Since the GOP hash is initialized, it can be updated with each incoming NALU hash.
+      SV_THROW(openssl_update_hash(self->crypto_handle, hash_to_add, hash_size, true));
+      item->used_in_gop_hash = true;  // Mark the item as used in the GOP hash
 
-    // Skip GOP SEI items as they do not contribute to the GOP hash.
-    if (item->nalu->is_gop_sei) {
       item = item->next;
-      continue;
     }
+    // Finally, mark the SEI item as used in the GOP hash
```
bjornvolcker, line 193: Why can't you also finalize the GOP hash while you're at it? You have both init and update in this function. To me it would make sense to also have finalize here as well.

bjornvolcker (changes requested):

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -195,6 +194,7 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
     // NALUs used in the GOP hash is set. This process will be modified after implementing the
     // verification of the previous GOP. For now, sei->used_in_gop_hash is set to true.
     sei->used_in_gop_hash = true;
+    SV_THROW(openssl_finalize_hash(self->crypto_handle, self->gop_info->computed_gop_hash, true));
```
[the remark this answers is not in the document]
bjornvolcker: You also need to put it above the comment, since that belongs to the `used_in_gop_hash` flag

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #225, opened 2024-09-12
Validating Previous Gop

mithydolphin:
Implements new computation for validating previous link hash.Computes
link hash on validation side and compare it with the received
linked hash from TLV.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -77,14 +77,10 @@ static const char *kAuthResultValidStr[SV_AUTH_NUM_SIGNED_GOP_VALID_STATES] = {"
  * indicating that validation is not possible.
  */
 static bool
-linked_hash_validation_is_not_feasible(signed_video_t *self)
+linked_hash_validation_is_not_feasible(const size_t hash_size, const uint8_t *linked_hashes)
 {
-  gop_info_t *gop_info = self->gop_info;
-  const size_t hash_size = self->verify_data->hash_size;
-  uint8_t linked_hashes[hash_size];
-  memset(linked_hashes, 0x00, sizeof(linked_hashes));
-  bool is_first_link = (memcmp(gop_info->linked_hashes, linked_hashes, hash_size) == 0);
-  return is_first_link;
+  const uint8_t linked_hash[MAX_HASH_SIZE] = {0};
+  return (memcmp(linked_hashes, linked_hash, hash_size) == 0);
 }
 
 /**
@@ -141,9 +137,10 @@ verify_linked_hash(signed_video_t *self)
 {
   gop_info_t *gop_info = self->gop_info;
   const size_t hash_size = self->verify_data->hash_size;
-  // If linked hash validation is not feasible, the linked hash verification
-  // does not affect the final outcome of the validation. Therefore, return true.
-  if (linked_hash_validation_is_not_feasible(self)) {
+  // The linked hash is used to validate the sequence of GOPs. Verification is only possible
+  // after receiving two complete GOPs, which is indicated by the presence of all-zero
+  // hashes in |linked_hashes|.
+  if (linked_hash_validation_is_not_feasible(hash_size, gop_info->linked_hashes)) {
     return true;
   }
 
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -118,7 +118,7 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
 
 /**
  * Compares the computed link hash with the linked hash received from the
- * signature.
+ * SEI.
  */
 bool
 verify_linked_hash(signed_video_t *self)
@@ -135,7 +135,7 @@ verify_linked_hash(signed_video_t *self)
 
 /**
  * Compares the computed GOP hash with the GOP hash received from the
- * signature.
+ * SEI.
  */
 static bool
 verify_gop_hash(signed_video_t *self)
@@ -152,19 +152,18 @@ verify_gop_hash(signed_video_t *self)
  * linked hash with the first NALU hash and marks it as used.
  */
 void
-update_link_hash_for_auth(signed_video_t *self, h26x_nalu_list_t *nalu_list)
+update_link_hash_for_auth(signed_video_t *self)
 {
   const size_t hash_size = self->verify_data->hash_size;
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
   h26x_nalu_list_item_t *item = nalu_list->first_item;
-  bool found_linked_hash = false;
-  while (item && !found_linked_hash) {
-    if (item->nalu->is_last_nalu_part && item->used_in_gop_hash &&
-        item->nalu->is_first_nalu_in_gop) {
+  while (item) {
+    if (item->used_in_gop_hash && item->nalu->is_first_nalu_in_gop) {
       if (!item->used_for_linked_hash) {
         update_linked_hash(self, item->second_hash, hash_size);
         item->used_for_linked_hash = true;
       }
-      found_linked_hash = true;
+      break;
     }
     item = item->next;
   }
@@ -370,7 +369,6 @@ verify_hashes_with_hash_list(signed_video_t *self,
         } else {
           item->validation_status = item->first_verification_not_authentic ? 'N' : '.';
           item->need_second_verification = false;
-          order_ok = true;
         }
         // Add missing items to |nalu_list|.
         int num_detected_missing_nalus =
@@ -522,7 +520,7 @@ verify_hashes_with_sei(signed_video_t *self, int *num_expected_nalus, int *num_r
   int num_received_hashes = -1;
   char validation_status = 'P';
   svrc_t status = SV_UNKNOWN_FAILURE;
-  // Verify the GOP hash.
+
   bool gop_is_ok = verify_gop_hash(self);
   bool order_ok = verify_linked_hash(self);
 
@@ -535,8 +533,8 @@ verify_hashes_with_sei(signed_video_t *self, int *num_expected_nalus, int *num_r
     num_expected_hashes = (int)self->gop_info->num_sent_nalus;
     // If the signature is verified but GOP hash or the linked hash is not, continue validation with
     // the hash list if it is present.
-    // TODO: Once forward linking is removed, validation will depend on both |!gop_is_ok| and the presence
-    // of the hash list 
+    // TODO: Once forward linking is removed, validation will depend on both |!gop_is_ok| and the
+    // presence of the hash list
     if (validation_status != '.' && self->gop_info->list_idx > 0) {
       return verify_hashes_with_hash_list(self, num_expected_nalus, num_received_nalus, order_ok);
     }
@@ -667,7 +665,6 @@ verify_hashes_without_sei(signed_video_t *self)
       item = item->next;
       continue;
     }
-    item->used_in_gop_hash = true;
     // A new GOP starts if the NALU |is_first_nalu_in_gop|. Such a NALU is hashed twice; as an
     // initial hash AND as a linking hash between GOPs. If this is the first time is is used in
     // verification it also marks the start of a new GOP.
@@ -1217,7 +1214,7 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
       latest->public_key_has_changed = false;
 
       SV_THROW(prepare_for_validation(self));
-      update_link_hash_for_auth(self, nalu_list);
+      update_link_hash_for_auth(self);
 
       if (!validation_flags->signing_present) {
         latest->authenticity = SV_AUTH_RESULT_NOT_SIGNED;
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -44,8 +44,7 @@ verify_hashes_with_hash_list(signed_video_t *self,
     int *num_received_nalus);
 static int
 set_validation_status_of_pending_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list,
-    char validation_status,
-    int verified_signature_hash);
+    char validation_status);
 static bool
 verify_hashes_with_gop_hash(signed_video_t *self, int *num_expected_nalus, int *num_received_nalus);
 static bool
@@ -141,6 +140,11 @@ verify_linked_hash(signed_video_t *self)
 {
   gop_info_t *gop_info = self->gop_info;
   const size_t hash_size = self->verify_data->hash_size;
+  // If linked hash validation is not feasible, the linked hash verification
+  // does not affect the final outcome of the validation. Therefore, return true.
+  if (linked_hash_validation_is_not_feasible(self)) {
+    return true;
+  }
 
   return (memcmp(gop_info->linked_hashes, self->received_linked_hash, hash_size) == 0);
 }
@@ -467,8 +471,7 @@ verify_hashes_with_hash_list(signed_video_t *self, int *num_expected_nalus, int 
  * Returns the number of items marked and -1 upon failure. */
 static int
 set_validation_status_of_pending_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list,
-    char validation_status,
-    int verified_signature_hash)
+    char validation_status)
 {
   if (!nalu_list) return -1;
 
@@ -485,8 +488,7 @@ set_validation_status_of_pending_items_used_in_gop_hash(h26x_nalu_list_t *nalu_l
         item->need_second_verification = true;
         item->first_verification_not_authentic = (validation_status != '.') ? true : false;
       } else {
-        item->validation_status =
-            set_validation_status_of_item(item, validation_status, verified_signature_hash);
+        item->validation_status = item->first_verification_not_authentic ? 'N' : validation_status;
         item->need_second_verification = false;
         num_marked_items++;
       }
@@ -529,30 +531,19 @@ verify_hashes_with_sei(signed_video_t *self, int *num_expected_nalus, int *num_r
   svrc_t status = SV_UNKNOWN_FAILURE;
   // Verify the GOP hash.
   bool gop_is_ok = verify_gop_hash(self);
-  bool order_ok;
-  // Check if verifying the link hash is feasible.
-  if (linked_hash_validation_is_not_feasible(self)) {
-    // If the its not feasible link hash verification is not effecting the final validation
-    // So it is set to true.
-    order_ok = true;
-  } else {
-    order_ok = verify_linked_hash(self);
-  }
+  bool order_ok = verify_linked_hash(self);
 
   // The verified_signature_hash indicates if the signature is verified.
   // If the signature hash is verified, the GOP hash can be verified as well.
   // If the signature hash is not verified, it means the SEI is corrupted, and the whole GOP status
   // is determined by the verified_signature_hash.
   if (self->gop_info->verified_signature_hash == 1) {
-    validation_status = '.';
+    validation_status = (gop_is_ok && order_ok) ? '.' : 'N';
     num_expected_hashes = (int)self->gop_info->num_sent_nalus;
-    if (!gop_is_ok || !order_ok) {
-      // If the signature is verified but GOP hash is not, continue validation with the
-      // hash list if it is present.
-      if (self->gop_info->list_idx > 0) {
-        return verify_hashes_with_hash_list(self, num_expected_nalus, num_received_nalus);
-      }
-      validation_status = 'N';
+    // If the signature is verified but GOP hash or the linked hash is not, continue validation with
+    // the hash list if it is present.
+    if ((!gop_is_ok || !order_ok) && self->gop_info->list_idx > 0) {
+      return verify_hashes_with_hash_list(self, num_expected_nalus, num_received_nalus);
     }
   } else if (self->gop_info->verified_signature_hash == 0) {
     validation_status = 'N';
@@ -569,8 +560,8 @@ verify_hashes_with_sei(signed_video_t *self, int *num_expected_nalus, int *num_r
   while (first_gop_hash_item && !first_gop_hash_item->used_in_gop_hash) {
     first_gop_hash_item = first_gop_hash_item->next;
   }
-  num_received_hashes = set_validation_status_of_pending_items_used_in_gop_hash(
-      self->nalu_list, validation_status, self->gop_info->verified_signature_hash);
+  num_received_hashes =
+      set_validation_status_of_pending_items_used_in_gop_hash(self->nalu_list, validation_status);
 
   if (!self->validation_flags.is_first_validation && first_gop_hash_item) {
     int num_missing_nalus = num_expected_hashes - num_received_hashes;
@@ -637,8 +628,8 @@ verify_hashes_with_gop_hash(signed_video_t *self, int *num_expected_nalus, int *
   while (first_gop_hash_item && !first_gop_hash_item->used_in_gop_hash) {
     first_gop_hash_item = first_gop_hash_item->next;
   }
-  num_received_hashes = set_validation_status_of_pending_items_used_in_gop_hash(
-      self->nalu_list, validation_status, self->gop_info->verified_signature_hash);
+  num_received_hashes =
+      set_validation_status_of_pending_items_used_in_gop_hash(self->nalu_list, validation_status);
 
   if (!self->validation_flags.is_first_validation && first_gop_hash_item) {
     int num_missing_nalus = num_expected_hashes - num_received_hashes;
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -44,7 +44,8 @@ verify_hashes_with_hash_list(signed_video_t *self,
     int *num_received_nalus);
 static int
 set_validation_status_of_pending_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list,
-    char validation_status);
+    char validation_status,
+    int verified_signature_hash);
 static bool
 verify_hashes_with_gop_hash(signed_video_t *self, int *num_expected_nalus, int *num_received_nalus);
 static bool
@@ -69,6 +70,22 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei);
 static const char *kAuthResultValidStr[SV_AUTH_NUM_SIGNED_GOP_VALID_STATES] = {"SIGNATURE MISSING",
     "SIGNATURE PRESENT", "NOT OK", "OK WITH MISSING INFO", "OK", "VERSION MISMATCH"};
 #endif
+
+/*
+ * Checks if linked hash validation is feasible at the start of validation.
+ * During the first validation slots, the hashes in |gop_info->linked_hashes| are all zeros,
+ * indicating that validation is not possible.
+ */
+static bool
+linked_hash_validation_is_not_feasible(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  const size_t hash_size = self->verify_data->hash_size;
+  uint8_t linked_hashes[hash_size];
+  memset(linked_hashes, 0x00, sizeof(linked_hashes));
+  bool is_first_link = (memcmp(gop_info->linked_hashes, linked_hashes, hash_size) == 0);
+  return is_first_link;
+}
 
 /**
  * The function is called when we receive a SEI NALU holding all the GOP information such as a
@@ -116,7 +133,20 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
 }
 
 /**
- * Computes the GOP hash from |nalu_hash_list| and compares it with the GOP hash received from the
+ * Compares the computed link hash with the linked hash received from the
+ * signature.
+ */
+bool
+verify_linked_hash(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  const size_t hash_size = self->verify_data->hash_size;
+
+  return (memcmp(gop_info->linked_hashes, self->received_linked_hash, hash_size) == 0);
+}
+
+/**
+ * Compares the computed GOP hash with the GOP hash received from the
  * signature.
  */
 static bool
@@ -128,9 +158,32 @@ verify_gop_hash(signed_video_t *self)
   return (memcmp(gop_info->computed_gop_hash, self->received_gop_hash, hash_size) == 0);
 }
 
-/**
- * Adds the hashes of NALUs used in computing the GOP hash and marks these NALUs as part of the GOP
- * hash.
+/*
+ * Iterates through the NALU list to find the first I-frame NALU used in the GOP hash.
+ * If the linked hash has not yet been updated with this NALU's hash, it updates the
+ * linked hash with the current I-frame hash and marks it as used.
+ */
+void
+update_link_hash_for_auth(signed_video_t *self, h26x_nalu_list_t *nalu_list)
+{
+  const size_t hash_size = self->verify_data->hash_size;
+  h26x_nalu_list_item_t *item = nalu_list->first_item;
+  bool found_linked_hash;
+  while (item && !found_linked_hash) {
+    if (item->used_in_gop_hash && item->nalu->is_first_nalu_in_gop &&
+        item->nalu->is_last_nalu_part) {
+
+      if (!item->used_for_linked_hash) {
+        update_linked_hash(self, item->second_hash, hash_size);
+        item->used_for_linked_hash = true;
+      }
+      found_linked_hash = true;
+    }
+    item = item->next;
+  }
+}
+
+/* Marks the NALUs that are used in GOP hash and copy the hashes of the NALUs to |nalu_hash_list|.
  *
  * This function iterates through the NALU list, identifies the NALUs that belong to the current
  * GOP, and marks them as used in the GOP hash. It initializes the GOP hash and updates it with
@@ -188,6 +241,8 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
       item = item->next;
     }
     SV_THROW(openssl_finalize_hash(self->crypto_handle, self->gop_info->computed_gop_hash, true));
+    update_link_hash_for_auth(self, nalu_list);
+
     // Finally, mark the SEI item as used in the GOP hash
     // TODO: Currently, the validation status of the SEI is set when the validation status of all
     // NALUs used in the GOP hash is set. This process will be modified after implementing the
@@ -412,7 +467,8 @@ verify_hashes_with_hash_list(signed_video_t *self, int *num_expected_nalus, int 
  * Returns the number of items marked and -1 upon failure. */
 static int
 set_validation_status_of_pending_items_used_in_gop_hash(h26x_nalu_list_t *nalu_list,
-    char validation_status)
+    char validation_status,
+    int verified_signature_hash)
 {
   if (!nalu_list) return -1;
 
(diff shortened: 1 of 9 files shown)


lib/src/signed_video_h26x_auth.c
```diff
@@ -478,7 +546,7 @@ verify_hashes_with_sei(signed_video_t *self, int *num_expected_nalus, int *num_r
   if (self->gop_info->verified_signature_hash == 1) {
     validation_status = '.';
     num_expected_hashes = (int)self->gop_info->num_sent_nalus;
-    if (!verify_gop_hash(self)) {
+    if (!gop_is_ok || !order_ok) {
```
bjornvolcker, line 549: I don't think this works. If `gop_is_ok = true` and `order_ok = false`, and you have a hash list, then you will verify with the hash list, which will result in OK, since `gop_is_ok`. Hence, the validation will become OK, when it is not.
One alternative solution is to rewrite this section as
```
validation_status = (gop_is_ok && order_ok) ? '.' : 'N';
num_expected_hashes = (int)self->gop_info->num_sent_nalus;
if (!gop_is_ok && order_ok && self->gop_info->list_idx > 0) {
  // Give validation a second chance if the GOP hash could not be verified, but the order
  // is OK, and the SEI includes a hash list.
  return verify_hashes_with_hash_list(self, num_expected_nalus, num_received_nalus);
 }
```
but note that there are other ways of doing it.

tests/check/check_signed_video_auth.c
```diff
@@ -1091,10 +1091,10 @@ START_TEST(lost_all_nalus_between_two_seis)
     // SI                ->   (valid) -> .P
     //  IPPPSS           ->   (valid) -> .....P
     //       SI          -> (invalid) -> MMMM.P (4 missed)
-    //        IPPPSI     -> (invalid) -> N....P
+    //        IPPPSI     ->   (valid) -> U....P
```
bjornvolcker, line 1094: Why does the I-frame get an unknown validation status? That should not be possible once the stream has been detected as signed. Once a SEI has occurred, every hashable NALU should get either 'N' or '.'

tests/check/check_signed_video_auth.c
```diff
@@ -693,8 +693,8 @@ START_TEST(remove_the_g_nalu)
   // authenticity is NOT OK.
   signed_video_accumulated_validation_t final_validation = {
       SV_AUTH_RESULT_NOT_OK, false, 17, 16, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
-  struct validation_stats expected = {.valid_gops = 3,
-      .invalid_gops = 2,
+  struct validation_stats expected = {.valid_gops = 4,
```
bjornvolcker, line 696: Can you elaborate on why you get more valid GOPs and update the description above, since that indicate there are only 3 valid GOPs.

tests/check/check_signed_video_auth.c
```diff
@@ -653,8 +653,8 @@ START_TEST(modify_one_sei)
   //           IPPSI   ->   (valid) -> ....P
   // One pending NAL Unit per GOP. Note that a modified 'I' affects two GOPs due to linked hashes,
   // but it will also affect a third if we validate with a gop_hash.
-  struct validation_stats expected = {.valid_gops = 2,
-      .invalid_gops = 2,
+  struct validation_stats expected = {.valid_gops = 3,
```
bjornvolcker, line 656: Can you elaborate on why you get more valid GOPs and update the description above, since that indicate there are only 2 valid GOPs.

tests/check/check_signed_video_auth.c
```diff
@@ -607,15 +607,15 @@ START_TEST(modify_one_i_nalu)
...
+  struct validation_stats expected = {.valid_gops = 2,
+      .invalid_gops = 2,
       .pending_nalus = 4,
       .final_validation = &final_validation};
   // For Frame level, the first GOP will be marked as valid with missing info since we cannot
   // correctly validate the last NAL Unit (the modified I).
   if (settings[_i].auth_level == SV_AUTHENTICITY_LEVEL_FRAME) {
-    expected.valid_gops = 2;
-    expected.invalid_gops = 2;
+    expected.valid_gops = 3;
```
bjornvolcker, line 617: Can you elaborate on why the test now validates one more GOP as OK.

tests/check/check_signed_video_auth.c
```diff
@@ -574,8 +574,8 @@ START_TEST(modify_one_p_nalu)
   signed_video_accumulated_validation_t final_validation = {
       SV_AUTH_RESULT_NOT_OK, false, 15, 14, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
   // One pending NAL Unit per GOP.
-  struct validation_stats expected = {.valid_gops = 2,
-      .invalid_gops = 2,
+  struct validation_stats expected = {.valid_gops = 3,
```
bjornvolcker, line 577: Can you elaborate on why the test now validates one more GOP as OK.

tests/check/check_signed_video_auth.c
```diff
@@ -537,8 +537,8 @@ START_TEST(interchange_two_p_nalus)
...
```
bjornvolcker, line 540: Can you elaborate on why the test now validates one more GOP as OK.

tests/check/check_signed_video_auth.c
```diff
@@ -492,8 +492,8 @@ START_TEST(remove_one_p_nalu)
   signed_video_accumulated_validation_t final_validation = {
       SV_AUTH_RESULT_NOT_OK, false, 14, 13, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
   // One pending NAL Unit per GOP.
-  struct validation_stats expected = {.valid_gops = 2,
-      .invalid_gops = 2,
+  struct validation_stats expected = {.valid_gops = 3,
```
bjornvolcker, line 495: Can you elaborate on why the test now validates one more GOP as OK.

tests/check/check_signed_video_auth.c
```diff
@@ -182,7 +182,7 @@ validate_nalu_list(signed_video_t *sv,
   ck_assert_int_eq(valid_gops_with_missing_info, expected.valid_gops_with_missing_info);
   ck_assert_int_eq(invalid_gops, expected.invalid_gops);
   ck_assert_int_eq(unsigned_gops, expected.unsigned_gops);
-  ck_assert_int_eq(missed_nalus, expected.missed_nalus);
+  //  ck_assert_int_eq(missed_nalus, expected.missed_nalus);
```
bjornvolcker, line 185: Why is the check commented out?

lib/src/signed_video_tlv.c
```diff
@@ -341,6 +341,15 @@ decode_general(signed_video_t *self, const uint8_t *data, size_t data_size)
       // Decode gop hash data.
       memcpy(self->received_gop_hash, data_ptr, hash_size);
       data_ptr += hash_size;
+#ifdef SIGNED_VIDEO_DEBUG
```
bjornvolcker, line 344: Remove these prints once you're done with debugging

lib/src/signed_video_h26x_sign.c
```diff
@@ -574,6 +574,12 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
         // adding it to the SEI NALU. Now it is time to start a new GOP, that is, hash and add this
         // first NALU of the GOP.
         SV_THROW(hash_and_add(self, &nalu));
+        // Chek if Nalu has been used for linking before.
+        if (!nalu.used_for_linked_hash) {
```
bjornvolcker, line 578: Is the member `used_for_linked_hash` really needed. The only place where it is changed is here and the same NALU can never end up here again.
Also, please explain why you are doing this in the code the reflects the "old" code and not in the "new" code if `self->linked_hash_on`

bjornvolcker (changes requested):
Some initial comments. The PR is quite difficult to follow with a lot of changes spread out in several places in the code.
A better commit message on what you're actually want to achieve would have been preferred. For example, there is nothing mentioned that the "old" why of linking is still present.

tests/check/check_signed_video_auth.c
```diff
@@ -491,9 +491,15 @@ START_TEST(remove_one_p_nalu)
   // authenticity is NOT OK.
   signed_video_accumulated_validation_t final_validation = {
       SV_AUTH_RESULT_NOT_OK, false, 14, 13, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
+  // SIPPSIPPSIPPSI
+  //
+  // SI                ->   (valid) -> .P
+  //  IPPSI            ->   (valid) -> .....P
+  //      IPPSI        -> (invalid) -> NNNNP
```
bjornvolcker, line 498: I think it actually adds a missing item after the first I-frame, but I'm not sure. Hence, it could be that it becomes "NMNNNP".

lib/src/signed_video_h26x_auth.c
```diff
@@ -116,7 +132,25 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+ * Compares the computed link hash with the linked hash received from the
+ * signature.
+ */
+bool
+verify_linked_hash(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  const size_t hash_size = self->verify_data->hash_size;
+  // If linked hash validation is not feasible, the linked hash verification
+  // does not affect the final outcome of the validation. Therefore, return true.
```
bjornvolcker, line 144: Rewrite comment to something like
"The linked hash is used to verify the order of GOPs. By definition, this is only feasible after having received two complete GOPs. This can be determined by identifying all-zero hashes in |linked_hashes|."

lib/src/signed_video_h26x_auth.c
```diff
@@ -70,6 +70,22 @@ static const char *kAuthResultValidStr[SV_AUTH_NUM_SIGNED_GOP_VALID_STATES] = {"
     "SIGNATURE PRESENT", "NOT OK", "OK WITH MISSING INFO", "OK", "VERSION MISMATCH"};
 #endif
 
+/*
+ * Checks if linked hash validation is feasible at the start of validation.
+ * During the first validation slots, the hashes in |gop_info->linked_hashes| are all zeros,
+ * indicating that validation is not possible.
+ */
+static bool
+linked_hash_validation_is_not_feasible(signed_video_t *self)
```
bjornvolcker, line 79: This function is called where you have access to "linked_hashes" and "hash_size", so it might be preferred to changed input parameters to "const uint8_t *linked_hashes, size_t hash_size"

lib/src/signed_video_h26x_auth.c
```diff
@@ -70,6 +70,22 @@ static const char *kAuthResultValidStr[SV_AUTH_NUM_SIGNED_GOP_VALID_STATES] = {"
...
+{
+  gop_info_t *gop_info = self->gop_info;
+  const size_t hash_size = self->verify_data->hash_size;
+  uint8_t linked_hashes[hash_size];
+  memset(linked_hashes, 0x00, sizeof(linked_hashes));
+  bool is_first_link = (memcmp(gop_info->linked_hashes, linked_hashes, hash_size) == 0);
+  return is_first_link;
```
bjornvolcker, line 86: No need for an intermediate boolean variable. Do `return (memcmp(gop_info->linked_hashes, linked_hashes, hash_size) == 0);` instead

lib/src/signed_video_h26x_auth.c
```diff
@@ -70,6 +70,22 @@ static const char *kAuthResultValidStr[SV_AUTH_NUM_SIGNED_GOP_VALID_STATES] = {"
...
+ * During the first validation slots, the hashes in |gop_info->linked_hashes| are all zeros,
+ * indicating that validation is not possible.
+ */
+static bool
+linked_hash_validation_is_not_feasible(signed_video_t *self)
+{
+  gop_info_t *gop_info = self->gop_info;
+  const size_t hash_size = self->verify_data->hash_size;
+  uint8_t linked_hashes[hash_size];
+  memset(linked_hashes, 0x00, sizeof(linked_hashes));
```
bjornvolcker, line 84: I don't think you need to use memset. It should be possible to set values already at declaration. Try
`const uint8_t linked_hash[MAX_HASH_SIZE] = {0};`

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -128,9 +160,30 @@ verify_gop_hash(signed_video_t *self)
...
-/**
- * Adds the hashes of NALUs used in computing the GOP hash and marks these NALUs as part of the GOP
- * hash.
+/*
+ * Iterates through the NALU list to find the first NALU used in the GOP hash.
+ * If the linked hash has not yet been updated with this NALU's hash, it updates the
+ * linked hash with the first NALU hash and marks it as used.
+ */
+void
+update_link_hash_for_auth(signed_video_t *self, h26x_nalu_list_t *nalu_list)
```
bjornvolcker, line 169: You use this function in 3 places; `prepare_for_link_and_gop_hash_verification(...)`, `verify_hashes_with_hash_list(...)` and `verify_hashes_without_sei(...)`.

The function `prepare_for_link_and_gop_hash_verification(...)` is called in `prepare_for_validation(...)`, but only if there is a SEI. That will not happen when the stream is for example not signed. Therefore, you had to add a call in `verify_hashes_without_sei(...)` to cover cases where the SEI is delayed.
Sometimes, `verify_hashes_without_sei(...)` is also called when for example the signature could not be verified. Then, link hash is updated twice, which then made it necessary to add a flag, so the same hash is not added again.
I might be very wrong when I interpret the code, but that is how it looks to me. Wouldn't it be more clean to make sure there is only one update made. For example, you can make `prepare_for_link_and_gop_hash_verification(...)` work also if there is no SEI, or you can break out the link update part from `prepare_for_link_and_gop_hash_verification(...)` and put it directly in `prepare_for_validation(...)`.

Why it is needed to call this update in `verify_hashes_with_hash_list(...)` I haven't figured out yet.

lib/src/signed_video_h26x_common.c
```diff
@@ -884,12 +890,17 @@ update_linked_hash(signed_video_t *self, uint8_t *hash, size_t hash_size)
...
-  // Copy the new hash to stored_hash
-  memcpy(stored_hash, hash, hash_size);
+  uint8_t *new_hash = &gop_info->linked_hashes[hash_size * 2];  // Newest linked hash
+  uint8_t *middle_hash = &gop_info->linked_hashes[hash_size];  // Middle hash in the chain
+  uint8_t *oldest_hash = &gop_info->linked_hashes[0];  // Oldest hash in the chain
+
+  // Move middle_hash to oldest_hash
+  memmove(oldest_hash, middle_hash, hash_size);
+  // Move new_hash to middle_hash
+  memmove(middle_hash, new_hash, hash_size);
```
bjornvolcker, line 900: Do you need to do this in 2 steps? I think it is possible to move 2 hashes in one go.

lib/src/signed_video_h26x_common.c
```diff
@@ -884,12 +890,17 @@ update_linked_hash(signed_video_t *self, uint8_t *hash, size_t hash_size)
...
   gop_info_t *gop_info = self->gop_info;
-  uint8_t *stored_hash = &gop_info->linked_hashes[hash_size];
-  uint8_t *linked_hash = &gop_info->linked_hashes[0];
-  // Move the content of stored_hash to linked_hash
-  memmove(linked_hash, stored_hash, hash_size);
-  // Copy the new hash to stored_hash
-  memcpy(stored_hash, hash, hash_size);
+  uint8_t *new_hash = &gop_info->linked_hashes[hash_size * 2];  // Newest linked hash
+  uint8_t *middle_hash = &gop_info->linked_hashes[hash_size];  // Middle hash in the chain
+  uint8_t *oldest_hash = &gop_info->linked_hashes[0];  // Oldest hash in the chain
```
bjornvolcker, line 895: Note that this buffer does not have all zeros from the beginning, which will make verifying the linked hash fail the first 2 times of a stream. When building in debug mode, which is what we do in development they are usually set to zero, but not when building in release mode. Valgrind will detect that and you can see it in the meson build test of this PR.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -137,13 +124,13 @@ verify_linked_hash(signed_video_t *self)
 {
   gop_info_t *gop_info = self->gop_info;
   const size_t hash_size = self->verify_data->hash_size;
+  const uint8_t linked_hash[MAX_HASH_SIZE] = {0};
   // The linked hash is used to validate the sequence of GOPs. Verification is only possible
   // after receiving two complete GOPs, which is indicated by the presence of all-zero
   // hashes in |linked_hashes|.
-  if (linked_hash_validation_is_not_feasible(hash_size, gop_info->linked_hashes)) {
+  if ((memcmp(gop_info->linked_hashes, linked_hash, hash_size) == 0)) {
```
bjornvolcker, line 131: You can even put it in the same return like
```
return ((memcmp(gop_info->linked_hashes, linked_hash, hash_size) == 0) || (memcmp(gop_info->linked_hashes, self->received_linked_hash, hash_size) == 0);
```

tests/check/check_signed_video_auth.c
```diff
@@ -495,7 +495,7 @@ START_TEST(remove_one_p_nalu)
   //
   // SI                ->   (valid) -> .P
   //  IPPSI            ->   (valid) -> .....P
-  //      IPPSI        -> (invalid) -> NNNNP
+  //      IPPSI        -> (invalid) -> NNMNNP
```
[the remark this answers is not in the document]
bjornvolcker: You should add the 'M' in the place where it occurs, not remove it. The test still reports 1 missing NALU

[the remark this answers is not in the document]
mithydolphin: S I  P P S I  P P P S I P P S I
1 2 3 4 5 6 7 8
 From the given stream, the extracted  8th element corresponds to the second P-frame of the 3rd GOP. I think the correct validation sequence is NNMNNP.

tests/check/check_signed_video_auth.c
```diff
@@ -2319,7 +2325,7 @@ signed_video_suite(void)
   tcase_add_loop_test(tc, all_seis_arrive_late_first_gop_scrapped, s, e);
   tcase_add_loop_test(tc, lost_g_before_late_sei_arrival, s, e);
   tcase_add_loop_test(tc, lost_g_and_gop_with_late_sei_arrival, s, e);
-  tcase_add_loop_test(tc, lost_all_nalus_between_two_seis, s, e);
+  tcase_add_loop_test(tc, lost_all_nalus_between_two_seis, s, 1);
```
bjornvolcker, line 2328: Looks like you're not running all scenarios...

lib/src/signed_video_h26x_auth.c
```diff
@@ -171,8 +158,9 @@ update_link_hash_for_auth(signed_video_t *self, h26x_nalu_list_t *nalu_list)
   const size_t hash_size = self->verify_data->hash_size;
   h26x_nalu_list_item_t *item = nalu_list->first_item;
   bool found_linked_hash;
-  while (item && !found_linked_hash && !item->nalu->is_gop_sei) {
-    if (item->used_in_gop_hash && item->nalu->is_last_nalu_part) {
+  while (item && !found_linked_hash) {
+    if (item->nalu->is_last_nalu_part && item->used_in_gop_hash &&
+        item->nalu->is_first_nalu_in_gop) {
```
bjornvolcker, line 163: If the linked hash can only be updated with I-frames, don't we get out of sync if an I-frame is lost?

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -491,6 +491,12 @@ START_TEST(remove_one_p_nalu)
   // authenticity is NOT OK.
   signed_video_accumulated_validation_t final_validation = {
       SV_AUTH_RESULT_NOT_OK, false, 14, 13, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
+  // SIPPSIPPSIPPSI
+  //
+  // SI                ->   (valid) -> .P
+  //  IPPSI            ->   (valid) -> .....P
+  //      IPPSI        -> (invalid) -> NNNNP
```
bjornvolcker, line 498: I think you are misunderstanding the review comments. First of all, since we now only apply the Document principle, all SEIs where the signature can be verified successfully should be marked '.'. Secondly, since the SEI can be verified successfully it is possible to detect how many NALUs there are missing, BUT if there is no hash list it is not possible to know which hash(es) are missing. They will either be added right after the I-frame or after the last frame of the GOP. You can look in the debug prints which it is. So I would expect this pattern to be either `NMNN.P` or `NNNM.P`

lib/src/signed_video_h26x_auth.c
```diff
@@ -69,7 +70,6 @@ compute_gop_hash(signed_video_t *self, h26x_nalu_list_item_t *sei);
 static const char *kAuthResultValidStr[SV_AUTH_NUM_SIGNED_GOP_VALID_STATES] = {"SIGNATURE MISSING",
     "SIGNATURE PRESENT", "NOT OK", "OK WITH MISSING INFO", "OK", "VERSION MISMATCH"};
 #endif
-
```
bjornvolcker, line 72: Add back this empty line

lib/src/signed_video_h26x_auth.c
```diff
@@ -309,16 +354,21 @@ verify_hashes_with_hash_list(signed_video_t *self, int *num_expected_nalus, int
...
-        // We have a match. Set validation_status and add missing nalus if we have detected any.
-        if (item->second_hash && !item->need_second_verification &&
-            item->nalu->is_first_nalu_in_gop) {
-          // If this |is_first_nalu_in_gop| it should be verified twice. If this the first time we
-          // signal that we |need_second_verification|.
-          DEBUG_LOG("This NALU needs a second verification");
-          item->need_second_verification = true;
+        if (!order_ok) {
+          item->validation_status = 'N';
+          order_ok = true;
```
bjornvolcker, line 359: Isn't there a need to update `need_second_verification` and/or `first_verification_not_authentic`?

lib/src/signed_video_h26x_auth.c
```diff
@@ -309,16 +354,21 @@ verify_hashes_with_hash_list(signed_video_t *self, int *num_expected_nalus, int
...
 
       if (memcmp(hash_to_verify, expected_hash, hash_size) == 0) {
-        // We have a match. Set validation_status and add missing nalus if we have detected any.
-        if (item->second_hash && !item->need_second_verification &&
-            item->nalu->is_first_nalu_in_gop) {
-          // If this |is_first_nalu_in_gop| it should be verified twice. If this the first time we
-          // signal that we |need_second_verification|.
-          DEBUG_LOG("This NALU needs a second verification");
-          item->need_second_verification = true;
+        if (!order_ok) {
```
bjornvolcker, line 357: Here is an alternative solution which could work until I-frames are hashed twice. That is to set the first item of the GOP to `first_verification_not_authentic |= !order_ok;` before searching for matches in the list. The first item will then automatically be marked as 'N'.

I think it would be more robust if you somehow could mark the first item independently of the search for matching hashes, since any added if-statement means an extra "opportunity" to end up in a wrong place/state.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
mithydolphin: In this case 
item->validation_status = item->first_verification_not_authentic |= !order_ok ? 'N' : '.';
would  other frames after I frames would be 'N' as well

tests/check/check_signed_video_auth.c
mithydolphin: Currently, the SEI is marked as |used_in_gop_hash|, and the validation status is set within the GOP. I have already added a TODO comment in the 
|prepare_for_link_and_gop_hash_verification| function. I plan to change this behavior when removing the first SEI pull request.

lib/src/signed_video_h26x_auth.c
mithydolphin: Now its only used in maybe_validate_gop

lib/src/signed_video_h26x_auth.c
bjornvolcker: Can you add another GOP at the end to the test case `remove_the_i_nalu`
Then the test should behave like this (skipping comments)
```
test_stream_t *list = create_signed_nalus("IPPIPPIPPIPPIPPI", settings[_i]);
test_stream_check_types(list, "SIPPSIPPSIPPSIPPSIPPSI");
const int remove_nalu_number = 10;
remove_item_then_check_and_free(list, remove_nalu_number, 'I');
test_stream_check_types(list, "SIPPSIPPSPPSIPPSIPPSI");
    SV_AUTH_RESULT_NOT_OK, false, 17, 16, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
struct validation_stats expected = {.valid_gops = 3,
    .invalid_gops = 3,
    .missed_nalus = 1,
    .pending_nalus = 6,
    .final_validation = &final_validation};
if (settings[_i].auth_level == SV_AUTHENTICITY_LEVEL_FRAME) {
  expected.valid_gops = 4;
  expected.invalid_gops = 2;
}
validate_nalu_list(NULL, list, expected, true);
test_stream_free(list);
```

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -495,7 +495,7 @@ START_TEST(remove_one_p_nalu)
   //
   // SI                ->   (valid) -> .P
   //  IPPSI            ->   (valid) -> .....P
-  //      IPPSI        -> (invalid) -> NNNNP (1 missed)
+  //      IPPSI        -> (invalid) -> MNNNNP (1 missed)
```
bjornvolcker, line 498: Are you sure it is added before the I-frame. Then it belongs to the previous GOP. That does not look correct.

bjornvolcker (changes requested):
I will go through this again tomorrow

lib/src/signed_video_h26x_auth.c
```diff
@@ -128,9 +146,31 @@ verify_gop_hash(signed_video_t *self)
...
+ * linked hash with the first NALU hash and marks it as used.
+ */
+void
+update_link_hash_for_auth(signed_video_t *self, h26x_nalu_list_t *nalu_list)
+{
+  const size_t hash_size = self->verify_data->hash_size;
+  h26x_nalu_list_item_t *item = nalu_list->first_item;
+  bool found_linked_hash = false;
+  while (item && !found_linked_hash) {
+    if (item->nalu->is_last_nalu_part && item->used_in_gop_hash &&
```
bjornvolcker, line 161: Since you rely on `used_in_gop_hash` to determine which item to use for updating the linked hash, it is vulnerable to order of operation and whether the gop hash has been computed or not. Can it happen that `used_in_gop_hash` is not set and items removed from the list before the first SEI arrives?

In addition, isn't `is_last_nalu_part` always `true`?

lib/src/signed_video_h26x_auth.c
```diff
@@ -128,9 +146,31 @@ verify_gop_hash(signed_video_t *self)
...
-/**
- * Adds the hashes of NALUs used in computing the GOP hash and marks these NALUs as part of the GOP
- * hash.
+/*
+ * Iterates through the NALU list to find the first NALU used in the GOP hash.
+ * If the linked hash has not yet been updated with this NALU's hash, it updates the
+ * linked hash with the first NALU hash and marks it as used.
+ */
+void
+update_link_hash_for_auth(signed_video_t *self, h26x_nalu_list_t *nalu_list)
```
bjornvolcker, line 155: You don't need `nalu_list` as input, since you feed this function with `self->nalu_list`.

lib/src/signed_video_h26x_auth.c
```diff
@@ -319,6 +370,7 @@ verify_hashes_with_hash_list(signed_video_t *self, int *num_expected_nalus, int
         } else {
           item->validation_status = item->first_verification_not_authentic ? 'N' : '.';
           item->need_second_verification = false;
+          order_ok = true;
```
bjornvolcker, line 373: This line has no impact since `order_ok` is always `true` when reaching this point.

lib/src/signed_video_h26x_auth.c
```diff
@@ -470,21 +522,23 @@ verify_hashes_with_sei(signed_video_t *self, int *num_expected_nalus, int *num_r
   int num_received_hashes = -1;
   char validation_status = 'P';
   svrc_t status = SV_UNKNOWN_FAILURE;
+  // Verify the GOP hash.
```
bjornvolcker, line 525: This comment is redundant with the function name. Remove.

lib/src/signed_video_h26x_auth.c
```diff
@@ -613,7 +667,7 @@ verify_hashes_without_sei(signed_video_t *self)
       item = item->next;
       continue;
     }
-
+    item->used_in_gop_hash = true;
```
bjornvolcker, line 670: Is it really necessary to mark the items if you remove the mark as soon as you leave the while loop?

lib/src/signed_video_h26x_common.c
```diff
@@ -874,6 +874,12 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
   openssl_update_hash(self->crypto_handle, hash, hash_size, true);
 }
 
+/*
+ * Updates the |linked_hash| buffer with the upcoming hash. The buffer contains 2 slots for hashes.
```
bjornvolcker, line 878: grammar: the upcoming hash -> a |hash|

lib/src/signed_video_h26x_common.c
```diff
@@ -874,6 +874,12 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
...
+ * When a new I NALU is encountered in the stream, the hash of the |linked_hash| buffer is updated
```
bjornvolcker, line 879: There is nothing in this function that says it only applies the update on I-frames only. Remove this sentence.

lib/src/signed_video_h26x_common.c
```diff
@@ -874,6 +874,12 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
...
+ * with the hash of the I NALU. While updating, all the values in the buffer are shifted to the
+ * start of the buffer, and the new hash is copied to the last slot of the buffer.
```
bjornvolcker, line 881: The last sentence is written as if there are more than one slot. You only have "old" and "new". Rewrite to match two slots.

lib/src/signed_video_h26x_common.c
```diff
@@ -884,12 +890,13 @@ update_linked_hash(signed_video_t *self, uint8_t *hash, size_t hash_size)
     if (hash_size != self->sign_data->hash_size) return SV_INVALID_PARAMETER;
   }
   gop_info_t *gop_info = self->gop_info;
-  uint8_t *stored_hash = &gop_info->linked_hashes[hash_size];
-  uint8_t *linked_hash = &gop_info->linked_hashes[0];
-  // Move the content of stored_hash to linked_hash
-  memmove(linked_hash, stored_hash, hash_size);
-  // Copy the new hash to stored_hash
-  memcpy(stored_hash, hash, hash_size);
+  uint8_t *new_hash = &gop_info->linked_hashes[hash_size];  // Newest linked hash
```
bjornvolcker, line 893: No need to add comment about "new" since the name already says that.

lib/src/signed_video_h26x_common.c
```diff
@@ -884,12 +890,13 @@ update_linked_hash(signed_video_t *self, uint8_t *hash, size_t hash_size)
...
+  uint8_t *old_hash = &gop_info->linked_hashes[0];  // Old hash in the chain
```
bjornvolcker, line 894: No need to add comment about "old" since the name already says that.

lib/src/signed_video_h26x_common.c
```diff
@@ -1317,4 +1325,4 @@ signed_video_is_golden_sei(signed_video_t *self, const uint8_t *nalu, size_t nal
   h26x_nalu_t parsed_nalu = parse_nalu_info(nalu, nalu_size, self->codec, false, true);
   free(parsed_nalu.nalu_data_wo_epb);
   return parsed_nalu.is_golden_sei;
-};
+};
```
bjornvolcker, line 1328: Add a return at end of file

lib/src/signed_video_h26x_sign.c
```diff
@@ -574,6 +574,8 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
         // adding it to the SEI NALU. Now it is time to start a new GOP, that is, hash and add this
         // first NALU of the GOP.
         SV_THROW(hash_and_add(self, &nalu));
+        // Chek if Nalu has been used for linking before.
```
bjornvolcker, line 577: Comment is fishy. Do you really check if linking has been done before? Aren't you actually updating the linked hash?

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -152,19 +152,18 @@ verify_gop_hash(signed_video_t *self)
  * linked hash with the first NALU hash and marks it as used.
  */
 void
-update_link_hash_for_auth(signed_video_t *self, h26x_nalu_list_t *nalu_list)
+update_link_hash_for_auth(signed_video_t *self)
 {
   const size_t hash_size = self->verify_data->hash_size;
+  h26x_nalu_list_t *nalu_list = self->nalu_list;
```
bjornvolcker, line 158: You only need `nalu_list` on the next line. How about `h26x_nalu_list_item_t *item = self->nalu_list->first_item;`

lib/src/signed_video_h26x_auth.c
```diff
@@ -152,19 +152,18 @@ verify_gop_hash(signed_video_t *self)
  * linked hash with the first NALU hash and marks it as used.
  */
 void
```
bjornvolcker, line 154: Make static

lib/src/signed_video_h26x_common.c
```diff
@@ -875,10 +875,9 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
 }
 
 /*
- * Updates the |linked_hash| buffer with the upcoming hash. The buffer contains 2 slots for hashes.
- * When a new I NALU is encountered in the stream, the hash of the |linked_hash| buffer is updated
- * with the hash of the I NALU. While updating, all the values in the buffer are shifted to the
- * start of the buffer, and the new hash is copied to the last slot of the buffer.
+ * Updates the |linked_hash| buffer with the |hash|. The buffer contains 2 slots for hashes.
+ * When a new NALU is encountered in the stream, the values in the buffer are shifted, and the
```
bjornvolcker, line 879: There is nothing in this function that checks if this is a new NALU and then shifts the buffer. That decision is done by the one using this function.
Remove that part from the description.

lib/src/signed_video_h26x_sign.c
```diff
@@ -574,7 +574,7 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
         // adding it to the SEI NALU. Now it is time to start a new GOP, that is, hash and add this
         // first NALU of the GOP.
         SV_THROW(hash_and_add(self, &nalu));
-        // Chek if Nalu has been used for linking before.
+        // Update link hash with |self->gop_info->nalu_hash|.
```
bjornvolcker, line 577: Redundant comment. It is easily read from the code line below. Remove.

bjornvolcker (changes requested):
5 remaining review comments to address or answer.

bjornvolcker (changes requested):
Still one question in a review comment to answer.

bjornvolcker (approved):

PR #227, opened 2024-10-02
Adds a helper function to print data in hex format

bjornvolcker:
useful for hashes and signatures. Currently only available under
the PRINT_DECODED_SEI flag.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #235, opened 2024-10-07
Prints decoded information of the crypto info tag

bjornvolcker:
Under the PRINT_DECODED_SEI flag prints the content of a Crypto
Information tag present in a SEI. This is applied to both the
current and the legacy code.

In addition, adds a helper function to convert a hash algorithm
on OID serialized form to a readable string.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #236, opened 2024-10-07
Removes Old Features

mithydolphin:
Removes forward linking and first SEI triggered by the first I-frame.
Update tests for authentication and signing accordingly.

### Describe your changes

- Fixes #(issue)


lib/src/signed_video_h26x_common.c
```diff
@@ -979,13 +979,6 @@ simply_hash(signed_video_t *self, const h26x_nalu_t *nalu, uint8_t *hash, size_t
     if (status == SV_OK) {
       // Finalize the ongoing hash of NALU parts.
       status = openssl_finalize_hash(self->crypto_handle, hash, false);
-      // For the first NALU in a GOP, the hash is used twice. Once for linking and once as reference
-      // for the future. Store the |nalu_hash| in |tmp_hash| to be copied for its second use, since
-      // it is not possible to recompute the hash from partial NALU data.
-      if (status == SV_OK && nalu->is_first_nalu_in_gop && !nalu->is_first_nalu_part) {
-        memcpy(self->gop_info->tmp_hash, hash, hash_size);
-        self->gop_info->tmp_hash_ptr = self->gop_info->tmp_hash;
```
bjornvolcker, line 987: I can't see that `tmp_hash_ptr` nor `tmp_hash` is used anymore.  Remove them from `gop_info` struct.

bjornvolcker (changes requested):
I've looked at the signing side only and it looks good. A few comments.

bjornvolcker (changes requested):
I've gone through the validation side of the tests and there are some unclarities. In particular when it comes to the cases where the linked hash was lost.

bjornvolcker (changes requested):
Also check valgrind. Somewhere, this PR leaks memory

tests/check/check_signed_video_auth.c
```diff
@@ -733,32 +654,20 @@ START_TEST(remove_the_i_nalu)
   signed_video_accumulated_validation_t final_validation = {
       SV_AUTH_RESULT_NOT_OK, false, 24, 23, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
   // A missing I NAL unit will affect two GOPs, as it is part of both the gop_hash and the
-  // linked_hash. At the GOP level, the missing NAL unit will invalidate both GOPs. However, at the
-  // Frame level, the second GOP, which is invalid at the GOP level, is considered valid because the
-  // previous GOP link has no effect on Frame level validation.
-  // IPPSIPPSPPSIPPSI
+  // linked_hash. At the GOP level, the missing NAL unit will invalidate both GOPs.
```
bjornvolcker, line 657: Remove "At the GOP level" since it holds also for frame level

tests/check/check_signed_video_auth.c
```diff
@@ -733,32 +654,20 @@ START_TEST(remove_the_i_nalu)
...
+  // IPPSIPPSPPSIPPSIPPSIPPSI
   //
   //  IPPS                      ->   (valid) -> ....
   //      IPPSP                 ->   (valid) -> ....
```
bjornvolcker, line 661: There are some errors here in validation stream. Not aligned with the total stream above and some weird ends for example here, including a "P"

tests/check/check_signed_video_auth.c
```diff
@@ -828,25 +828,30 @@ START_TEST(sei_arrives_late)
   // |settings|; See signed_video_helpers.h.
 
   test_stream_t *list = create_signed_nalus("IPPPIPPPIPPPI", settings[_i]);
-  test_stream_check_types(list, "SIPPPSIPPPSIPPPSI");
+  test_stream_check_types(list, "IPPPSIPPPSIPPPSI");
 
-  // Remove the second SEI, that is, number 6 in the list: SIPPP (S) IPPPSIPPPSI.
-  test_stream_item_t *sei = test_stream_item_remove(list, 6);
+  // Remove the second SEI, that is, number 5 in the list: IPPP (S) IPPPSIPPPSI.
```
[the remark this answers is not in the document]
bjornvolcker: I think you should keep the original test setup and NOT move the first SEI, but the others.

tests/check/check_signed_video_auth.c
```diff
@@ -1218,32 +1221,29 @@ START_TEST(detect_change_of_public_key)
...
+      SV_AUTH_RESULT_NOT_OK, true, 13, 12, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
   // The list will be validated successfully up to the third SEI (S) which has the new Public key.
-  //
-  //   SI      -> .P     (valid, 1 pending, public_key_has_changed = false)
-  //   IPPSI   -> ....P  (valid, 1 pending, public_key_has_changed = false)
-  //   IPPS*I  -> NNN.P  (invalid, 1 pending, public_key_has_changed = true, -3 missing)
-  //   IPPPS*I -> N....P (invalid, 1 pending, public_key_has_changed = false)
+  //   IPPSIPPIPPPSI
+  //   IPPS             -> ....     (valid, public_key_has_changed = false)
+  //       IPPIPPPS     -> NNN..... (invalid, public_key_has_changed = true)
```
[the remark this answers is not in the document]
bjornvolcker: Your change did not solve the issue of getting a GOP without a SEI.

bjornvolcker (changes requested):
Will look into the validation side soon

lib/src/signed_video_h26x_auth.c
```diff
@@ -1123,11 +1092,10 @@ validation_is_feasible(const h26x_nalu_list_item_t *item)
   if (item->nalu->is_golden_sei) return false;
   if (!item->nalu->is_hashable) return false;
   if (item->validation_status != 'P') return false;
-
   // Validation may be done upon a SEI.
   if (item->nalu->is_gop_sei) return true;
   // Validation may be done upon the end of a GOP.
-  if (item->nalu->is_first_nalu_in_gop && !item->need_second_verification) return true;
+  if (item->nalu->is_first_nalu_in_gop) return true;
   // Validation may be done upon a hashable NALU right after a SEI. This happens when the SEI was
```
bjornvolcker, line 1099: The comment states something that is no longer feasible. Remove. Also, remove the while loop below because it was used to find the next NALU.

lib/src/signed_video_h26x_auth.c
```diff
@@ -1082,7 +1051,7 @@ has_pending_gop(signed_video_t *self)
     // Collect statistics from pending and hashable NALUs only. The others are either out of date or
     // not part of the validation.
     if (item->validation_status == 'P' && item->nalu && item->nalu->is_hashable) {
-      num_pending_gop_ends += (item->nalu->is_first_nalu_in_gop && !item->need_second_verification);
+      num_pending_gop_ends += (item->nalu->is_first_nalu_in_gop && !item->used_for_linked_hash);
```
bjornvolcker, line 1054: Under what circumstances can an I-frame be pending and `used_for_linked_hash` at the same time?

lib/src/signed_video_h26x_auth.c
```diff
@@ -747,7 +714,8 @@ validate_authenticity(signed_video_t *self)
       verify_success = verify_hashes_with_gop_hash(self, &num_expected_nalus, &num_received_nalus);
     }
   }
-
+  // Set |latest_validated_gop| to recived gop counter for the next validation.
+  self->gop_info->latest_validated_gop = self->gop_info->global_gop_counter;
```
bjornvolcker, line 718: Isn't this in contradiction to line 654 where `latest_validated_gop` is incremented? If  a SEI is lost for example it looks like you first update from, for example, 4 to 6, and then increment on line 654 to 7.

lib/src/signed_video_h26x_auth.c
```diff
@@ -667,24 +638,20 @@ verify_hashes_without_sei(signed_video_t *self)
...
     // verification it also marks the start of a new GOP.
-    found_next_gop = item->nalu->is_first_nalu_in_gop && !item->need_second_verification;
-
-    // Mark the item as 'Not Authentic' or keep it for a second verification.
-    if (found_next_gop) {
-      // Keep the item pending and mark the first verification as not authentic.
-      item->need_second_verification = true;
-      item->first_verification_not_authentic = true;
-    } else if (item->validation_status == 'P') {
+    if (item->validation_status == 'P' && !item->nalu->is_gop_sei) {
```
bjornvolcker, line 641: Isn't `validation_status` by definition equal to `'P'` when reaching this step?

lib/src/signed_video_h26x_auth.c
```diff
@@ -328,28 +333,19 @@ verify_hashes_with_hash_list(signed_video_t *self,
...
-    // 'N'. If that is the case, set |first_verification_not_authentic| to true and set |order_ok|
-    // to true for the next NALUs, so they are not affected by this issue.
-    if (!order_ok) {
-      item->first_verification_not_authentic = true;
-      order_ok = true;
+      last_used_item = item;
+      break;
     }
 
     last_used_item = item;
```
bjornvolcker, line 343: If you do this before the if-statement above you can skip line 339

lib/src/signed_video_h26x_auth.c
```diff
@@ -421,25 +408,17 @@ verify_hashes_with_hash_list(signed_video_t *self,
...
-      last_used_item->first_verification_not_authentic = true;
-      // Give this NALU a second verification because it could be that it is present in the next GOP
-      // and brought in here due to some lost NALUs.
-      last_used_item->need_second_verification = true;
     } else {
       // Last received hash is valid. Check if there are unused hashes in |hash_list|. Note that the
       // index of the hashes span from 0 to |num_expected_hashes| - 1, so if |latest_match_idx| =
       // |num_expected_hashes| - 1, we have no pending nalus.
       int num_unused_expected_hashes = num_expected_hashes - 1 - latest_match_idx;
       // We cannot mark the last item as Missing since it will be handled a second time in the next
```
bjornvolcker, line 420: This comment is no longer correct

lib/src/signed_video_h26x_auth.c
```diff
@@ -421,25 +408,17 @@ verify_hashes_with_hash_list(signed_video_t *self,
   // If the last invalid NALU is the first NALU in a GOP or the NALU after the SEI, keep it
   // pending. If the last NALU is valid and there are more expected hashes we either never
   // verified any hashes or we have missing NALUs.
-  if (last_used_item) {
+  // if (last_used_item) {
```
bjornvolcker, line 411: Remove this line

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -933,38 +935,39 @@ START_TEST(all_seis_arrive_late_first_gop_scrapped)
...
   test_stream_item_free(item);
 
-  // ISPPPIPSPPIPSPPIPSS
+  // IPPPIPSPPIPSPPIPSS
   //
-  // IS                  ->    (signature) -> PU             1 pending
-  // ISPPPIPS            ->    (signature) -> PUPPPPPU       6 pending
-  // ISPPPIPSPPIPS       ->        (valid) -> .U...PPUPPPP.  6 pending
+  // IPPPI               ->(no signature)  -> PPPPP          5 pending
+  // IPPPIPS             ->    (signature) -> PPPPPPU        6 pending
```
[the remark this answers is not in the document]
mithydolphin: Currently Sei is validated as part of  the gop

lib/src/signed_video_h26x_sign.c
```diff
@@ -525,8 +525,7 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
     }
 
     // Finalize GOP hash and generate SEI if the NALU is I frame of a non-empty GOP.
-    if (nalu.is_first_nalu_in_gop && nalu.is_last_nalu_part && gop_info->global_gop_counter > 0) {
-      // Store the timestamp for the first nalu in gop.
+    if (nalu.is_first_nalu_in_gop && nalu.is_last_nalu_part && self->generate_first_sei) {
```
bjornvolcker, line 528: I think it would be more readable if you write
```
if (nalu.is_first_nalu_in_gop && nalu.is_last_nalu_part) {
  if (self->generate_first_sei) {
    // The code to generate SEI
  }
  self->generate_first_sei = true;
}
```

tests/check/check_signed_video_auth.c
```diff
@@ -1120,29 +1120,36 @@ START_TEST(detect_change_of_public_key)
   // |settings|; See signed_video_helpers.h.
 
   // Generate 2 GOPs
-  test_stream_t *list = create_signed_nalus("IPPIPP", settings[_i]);
-  test_stream_check_types(list, "IPPSIPP");
-
+  test_stream_t *list = create_signed_nalus("IPPIPPI", settings[_i]);
+  test_stream_check_types(list, "IPPSIPPSI");
+  test_stream_item_t *i_item = test_stream_item_remove(list, 9);
```
bjornvolcker, line 1125: I think it is more clear if you use `test_stream_pop_last_item(...)` instead. Then one does not have to count number of characters to understand that you've removed the last item.

tests/check/check_signed_video_auth.c
```diff
@@ -1120,29 +1120,36 @@ START_TEST(detect_change_of_public_key)
...
   // Generate another GOP from scratch
   // This will generate a new private key, hence transmit a different public key.
-  test_stream_t *list_with_new_public_key = create_signed_nalus_int("IPPPIPPI", settings[_i], true);
-  test_stream_check_types(list_with_new_public_key, "IPPPSIPPSI");
-
+  test_stream_t *list_with_new_public_key =
+      create_signed_nalus_int("IPIPIPPPIPPI", settings[_i], true);
+  test_stream_check_types(list_with_new_public_key, "IPSIPSIPPPSIPPSI");
+  int remove_items = 6;
+  while (remove_items--) {
```
bjornvolcker, line 1133: I think you should use `test_stream_pop(...)` instead of using a while loop. That function will remove the X first items and return a new list including them.

tests/check/check_signed_video_auth.c
```diff
@@ -1120,29 +1120,36 @@ START_TEST(detect_change_of_public_key)
...
+  test_stream_item_check_type(i_item, 'I');
   // Generate another GOP from scratch
   // This will generate a new private key, hence transmit a different public key.
-  test_stream_t *list_with_new_public_key = create_signed_nalus_int("IPPPIPPI", settings[_i], true);
-  test_stream_check_types(list_with_new_public_key, "IPPPSIPPSI");
-
+  test_stream_t *list_with_new_public_key =
+      create_signed_nalus_int("IPIPIPPPIPPI", settings[_i], true);
+  test_stream_check_types(list_with_new_public_key, "IPSIPSIPPPSIPPSI");
+  int remove_items = 6;
```
bjornvolcker, line 1132: You should add a comment here on why you remove the first two GOPs. It is not clear to the reader.

tests/check/check_signed_video_auth.c
```diff
@@ -1120,29 +1120,36 @@ START_TEST(detect_change_of_public_key)
...
-      SV_AUTH_RESULT_NOT_OK, true, 17, 16, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
+      SV_AUTH_RESULT_NOT_OK, true, 18, 17, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
   // The list will be validated successfully up to the third SEI (S) which has the new Public key.
-  //   IPPSIPPIPPPSIPPSI
+  //   IPPSIPPSIPPPSIPPSI
   //   IPPS                   -> ....       (valid, public_key_has_changed = false)
-  //       IPPIPPPS*          -> NNN..... (invalid, public_key_has_changed = true)
+  //       IPPS               -> ....       (valid, public_key_has_changed = false)
+  //           IPPPS*         -> NNNNN      (invalid, public_key_has_changed = true)
   //               IPPS*      -> ....       (valid, public_key_has_changed = false)
```
bjornvolcker, line 1149: This line is not aligned correctly. The I is under the S.

tests/check/check_signed_video_auth.c
```diff
@@ -1120,29 +1120,36 @@ START_TEST(detect_change_of_public_key)
...
+  test_stream_check_types(list, "IPPSIPPSIPPPSIPPSI");
 
   // Final validation is NOT OK and all received NAL Units, but the last, are validated. The
   // |public_key_has_changed| flag has been set.
   signed_video_accumulated_validation_t final_validation = {
-      SV_AUTH_RESULT_NOT_OK, true, 17, 16, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
+      SV_AUTH_RESULT_NOT_OK, true, 18, 17, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
   // The list will be validated successfully up to the third SEI (S) which has the new Public key.
-  //   IPPSIPPIPPPSIPPSI
+  //   IPPSIPPSIPPPSIPPSI
```
bjornvolcker, line 1145: Put * on the S that includes a new key

lib/src/signed_video_h26x_auth.c
```diff
@@ -638,10 +632,12 @@ verify_hashes_without_sei(signed_video_t *self)
     // A new GOP starts if the NALU |is_first_nalu_in_gop|. Such a NALU is hashed twice; as an
     // initial hash AND as a linking hash between GOPs. If this is the first time is is used in
     // verification it also marks the start of a new GOP.
-    if (item->validation_status == 'P' && !item->nalu->is_gop_sei) {
+    if (!item->nalu->is_gop_sei) {
```
bjornvolcker, line 635: To make it more readable I would do
```
if (item->nalu->is_gop_sei) {
  break;
}
```

lib/src/signed_video_h26x_auth.c
```diff
@@ -334,13 +332,12 @@ verify_hashes_with_hash_list(signed_video_t *self,
     found_item_after_sei = (item->prev == sei);
     // Check if this |is_first_nalu_in_gop|, but not used before.
     found_next_gop = (item->nalu->is_first_nalu_in_gop && !item->used_for_linked_hash);
+    last_used_item = item;
     // If this is a SEI, it is not part of the hash list and should not be verified.
     if (item->nalu->is_gop_sei) {
       last_used_item = item;
```
bjornvolcker, line 338: As said in the previous review comment. You've already set this (now on line 335). You do not need to set it again.

lib/src/signed_video_internal.h
```diff
@@ -144,6 +144,7 @@ struct _signed_video_t {
   bool is_golden_sei;  // Flag that tells if a SEI is a golden SEI
   bool using_golden_sei;  // Flag that tells if golden SEI prinsiple is used
   bool signing_started;
+  bool generate_first_sei;
```
bjornvolcker, line 147: This name is not good. It indicates that you want to generate a first sei, but it becomes true after the first I-frame, hence all SEIs are look like being "first".
Try to find a name that better suites its purpose, for example, `is_first_triggered_signing`

lib/src/signed_video_h26x_auth.c
```diff
@@ -1042,10 +1038,8 @@ has_pending_gop(signed_video_t *self)
   bool found_pending_gop_sei = false;
   bool found_pending_nalu_after_gop_sei = false;
   bool found_pending_gop = false;
-
```
bjornvolcker, line 1076: Why did you remove this empty line? I think it is good to separate the actions.

lib/src/signed_video_h26x_auth.c
```diff
@@ -1042,10 +1038,8 @@ has_pending_gop(signed_video_t *self)
...
   // Reset the |gop_state| members before running through the NALUs in |nalu_list|.
   gop_state_reset(gop_state);
-
```
bjornvolcker, line 1079: Why did you remove this empty line? I think it is good to separate the actions.

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -1354,23 +1348,24 @@ START_TEST(fast_forward_stream_without_reset)
...
   // Final validation is NOT OK and all received NAL Units, but the last one, are validated.
   signed_video_accumulated_validation_t final_validation = {
-      SV_AUTH_RESULT_NOT_OK, false, 21, 20, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
+      SV_AUTH_RESULT_NOT_OK, false, 20, 19, 1, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
   // Validate IP SIPPPSIPPPSI (without reset, i.e., started with IP before fast forward):
-
-  // SI             -> NP            ->   (invalid)
-  //  IPPPSI        -> NNNNNP        ->   (invalid)
-  //       IPPPSI   -> NNNNNP        ->   (invalid)
+  // IPS              -> NNP          ->   (invalid)
```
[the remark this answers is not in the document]
mithydolphin: Its verifying without SEI in the first gop So its not validating the SEI.

lib/src/signed_video_h26x_auth.c
```diff
@@ -629,16 +629,14 @@ verify_hashes_without_sei(signed_video_t *self)
       item = item->next;
       continue;
     }
-    // A new GOP starts if the NALU |is_first_nalu_in_gop|. Such a NALU is hashed twice; as an
-    // initial hash AND as a linking hash between GOPs. If this is the first time is is used in
-    // verification it also marks the start of a new GOP.
-    if (!item->nalu->is_gop_sei) {
-      item->need_second_verification = false;
```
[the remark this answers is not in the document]
mithydolphin: Yes its currently not needed.

lib/src/signed_video_internal.h
```diff
@@ -144,7 +144,7 @@ struct _signed_video_t {
   bool is_golden_sei;  // Flag that tells if a SEI is a golden SEI
   bool using_golden_sei;  // Flag that tells if golden SEI prinsiple is used
   bool signing_started;
-  bool is_first_triggered_signing;
+  bool generate_sei;
```
bjornvolcker, line 147: I think this flag now deserves an explanation since it can be interpreted as controlling SEIs in general.

bjornvolcker (changes requested):
I will go through the entire PR again.

lib/src/signed_video_internal.h
bjornvolcker: Maybe the explanation got a little bit too detailed (sorry). It would have been enough to say "Flips to true after the first signing attempt".

I would probably have named it `first_signing_attempt` and set it to `true` at create time. Then flip it to `false` after the first attempt.

lib/src/signed_video_internal.h
```diff
@@ -144,9 +144,14 @@ struct _signed_video_t {
   bool is_golden_sei;  // Flag that tells if a SEI is a golden SEI
   bool using_golden_sei;  // Flag that tells if golden SEI prinsiple is used
   bool signing_started;
+  bool generate_sei;  // A flag that tells whether to generate SEI
```
bjornvolcker, line 147: Shouldn't you set this to `false` upon reset?

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -1759,6 +1653,10 @@ START_TEST(vendor_axis_communications_operation)
   // Add an 'I' to trigger a SEI.
```
bjornvolcker, line 1653: Update this comment now when the action below has changed

lib/src/signed_video_h26x_sign.c
```diff
@@ -529,46 +528,20 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
...
-        // adding it to the SEI NALU. Now it is time to start a new GOP, that is, hash and add this
-        // first NALU of the GOP.
-        SV_THROW(hash_and_add(self, &nalu));
-        SV_THROW(update_linked_hash(self, self->gop_info->nalu_hash, self->sign_data->hash_size));
-      }
+      self->generate_sei = true;
     }
+    SV_THROW(hash_and_add(self, &nalu));
+    if (nalu.is_first_nalu_in_gop && nalu.is_last_nalu_part) {
+      SV_THROW(update_linked_hash(self, self->gop_info->nalu_hash, self->sign_data->hash_size));
```
bjornvolcker, line 542: nit: `self->gop_info` -> `gop_info`

lib/src/signed_video_h26x_nalu_list.c
```diff
@@ -316,6 +316,7 @@ h26x_nalu_list_copy_last_item(h26x_nalu_list_t *list, bool hash_algo_known)
   uint8_t *hashable_data = NULL;
   uint8_t *nalu_data_wo_epb = NULL;
   h26x_nalu_list_item_t *item = list->last_item;
+  item = (item->validation_status != 'M') ? item : item->prev;
```
bjornvolcker, line 319: What happens if there are more than one "missing items" at the end?

lib/src/signed_video_h26x_nalu_list.c
```diff
@@ -428,8 +429,12 @@ h26x_nalu_list_remove_missing_items(h26x_nalu_list_t *list)
       item->validation_status = 'U';
       found_decoded_sei = true;
     }
-    if (item->validation_status == 'P') found_first_pending_nalu = true;
+    if (item->validation_status == 'N') {
+      // Reset validation status to 'P' for the next validation.
+      item->validation_status = 'P';
```
bjornvolcker, line 434: Is this really a good and proper action?
First of all, the function name says that only missing items are removed, but in fact all invalid items have their validation status changed.
Second, what will happen if we have `IPPPSIPPSI` and the first `P` of the first GOP is modified?

lib/src/signed_video_h26x_auth.c
```diff
@@ -142,7 +141,6 @@ verify_gop_hash(signed_video_t *self)
 {
   gop_info_t *gop_info = self->gop_info;
   const size_t hash_size = self->verify_data->hash_size;
-
```
bjornvolcker, line 145: Add back this empty line. It makes it more clear what is being returned.

lib/src/signed_video_h26x_auth.c
```diff
@@ -198,32 +196,38 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
...
       // Skip GOP SEI items as they do not contribute to the GOP hash.
       if (item->nalu->is_gop_sei) {
         item = item->next;
+        if (item) {
+          // Track when the current item follows the SEI or marks the beginning of the next GOP.
+          found_item_after_sei = (item->prev == sei);
+        } else {
+          found_next_gop = true;
+        }
         continue;
```
bjornvolcker, line 221: I think it becomes more clear to break the while loop instead of first look ahead to set a flag that will then break the while loop upon continuation.

lib/src/signed_video_h26x_auth.c
```diff
@@ -198,32 +196,38 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
...
       }
-      hash_to_add = item->need_second_verification ? item->second_hash : item->hash;
+      hash_to_add = item->nalu->is_first_nalu_in_gop ? item->second_hash : item->hash;
       // Since the GOP hash is initialized, it can be updated with each incoming NALU hash.
       SV_THROW(openssl_update_hash(self->crypto_handle, hash_to_add, hash_size, true));
       item->used_in_gop_hash = true;  // Mark the item as used in the GOP hash
-
```
bjornvolcker, line 225: Add back this empty line. It becomes more clear that the end of the while is reached and moved to the next item.

lib/src/signed_video_h26x_auth.c
```diff
@@ -198,32 +196,38 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
...
       item = item->next;
+      if (item && item->nalu) {
+        found_next_gop = (item->nalu->is_first_nalu_in_gop);
```
bjornvolcker, line 229: Is this really necessary now when number of I-frames are counted?

lib/src/signed_video_h26x_auth.c
```diff
@@ -332,28 +336,17 @@ verify_hashes_with_hash_list(signed_video_t *self,
     // Check if this is the item right after the |sei|.
     found_item_after_sei = (item->prev == sei);
     // Check if this |is_first_nalu_in_gop|, but not used before.
-    found_next_gop = (item->nalu->is_first_nalu_in_gop && !item->need_second_verification);
+    found_next_gop = (item->nalu->is_first_nalu_in_gop && !item->used_for_linked_hash);
```
bjornvolcker, line 339: Can you give an example where `item->nalu->is_first_nalu_in_gop` and `!item->used_for_linked_hash` and `item->validation_status != 'P'`

lib/src/signed_video_h26x_auth.c
```diff
@@ -391,20 +381,15 @@ verify_hashes_with_hash_list(signed_video_t *self,
...
-        item->validation_status = 'N';
-      }
+
+      item->validation_status = 'N';
       // Update counters.
       num_invalid_nalus_since_latest_match++;
     }
     item = item->next;
+    if (item && item->nalu) {
+      found_next_gop = (item->nalu->is_first_nalu_in_gop && !item->used_for_linked_hash);
```
bjornvolcker, line 391: Can you give an example when this item at the same time is NOT pending?

lib/src/signed_video_h26x_auth.c
```diff
@@ -415,8 +400,7 @@ verify_hashes_with_hash_list(signed_video_t *self,
     // We do not know where in the sequence of NALUs they were lost. Simply add them before the
     // first item. If the first item needs a second opinion, that is, it has already been verified
     // once, we append that item. Otherwise, prepend it with missing items.
-    const bool append =
-        nalu_list->first_item->second_hash && !nalu_list->first_item->need_second_verification;
+    const bool append = nalu_list->first_item->second_hash;
```
bjornvolcker, line 403: It might be better to do `const bool append = !!nalu_list->first_item->second_hash;`

lib/src/signed_video_h26x_auth.c
```diff
@@ -668,27 +634,23 @@ verify_hashes_without_sei(signed_video_t *self)
...
+    // GOP.
+    if (item->nalu->is_gop_sei) {
+      break;
     }
+    item->validation_status = 'N';
+    num_marked_items++;
+
     item = item->next;
+    if (item) {
+      found_next_gop = item->nalu->is_first_nalu_in_gop && !item->used_for_linked_hash;
```
bjornvolcker, line 647: Can you give an example when `found_next_gop == true` and at the same time is NOT pending.

lib/src/signed_video_h26x_auth.c
```diff
@@ -742,6 +704,8 @@ validate_authenticity(signed_video_t *self)
   } else {
     if (self->gop_info->signature_hash_type == DOCUMENT_HASH) {
       verify_success = verify_hashes_with_sei(self, &num_expected_nalus, &num_received_nalus);
+      // Set |latest_validated_gop| to recived gop counter for the next validation.
+      self->gop_info->latest_validated_gop = self->gop_info->global_gop_counter;
```
bjornvolcker, line 708: There might be a conflict between operations. Here we SET `latest_validated_gop` whereas on line 653 above we INCREMENT it. Can you think of any case where this can cause unsynchronized validation for example?

lib/src/signed_video_h26x_auth.c
```diff
@@ -751,7 +715,6 @@ validate_authenticity(signed_video_t *self)
       verify_success = verify_hashes_with_gop_hash(self, &num_expected_nalus, &num_received_nalus);
     }
   }
-
```
bjornvolcker, line 754: Add back this empty line

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
mithydolphin: We’re only incrementing |latest_validated_gop |while verifying without SEI, which means we only increase it when we’re out of sync. I haven’t thought of any case where this could lead to desynchronization after setting |self->gop_info->latest_validated_gop = self->gop_info->global_gop_counter|.

lib/src/signed_video_h26x_auth.c
mithydolphin: I don’t think there’s a situation where this could happen. We’re skipping all the NALUs that aren’t pending in line 633, and after setting the NALU status from P to N, we immediately check the next pending item to find the next GOP.

So I’m not really sure I understand what you’re asking or why you're concerned. Can you clarify what you mean?

bjornvolcker: So this item will always be pending. Good.

lib/src/signed_video_h26x_auth.c
bjornvolcker: Can this situation cause any problems, where the second `S` is lost?
`IPPISPPPIPPPPIPSPPISP`
Or if we lose 2 SEIs?

mithydolphin: the validation will be 
 IPPISPPPIPPPPIPSPPISP
 IPPIS                                                          -> ...P.
       ISPPPIPPPPIPS                                   ->N.NNN.....PP.(NOT_OK)
                            IPSPPIS                           ->.....P.
Wich I think its true.

tests/check/check_signed_video_auth.c
```diff
@@ -2251,6 +2252,7 @@ signed_video_suite(void)
   tcase_add_loop_test(tc, no_public_key_in_sei_and_bad_public_key_on_validation_side, s, e);
   tcase_add_loop_test(tc, fallback_to_gop_level, s, e);
   tcase_add_loop_test(tc, golden_sei_principle, s, e);
+
```
bjornvolcker, line 2255: Remove this empty line

lib/src/signed_video_h26x_auth.c
```diff
@@ -211,23 +210,14 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
...
       if (item->nalu->is_gop_sei) {
-        item = item->next;
-        if (item) {
-          // Track when the current item follows the SEI or marks the beginning of the next GOP.
-          found_item_after_sei = (item->prev == sei);
-        } else {
-          found_next_gop = true;
-        }
-        continue;
+        break;  // Break if encountered SEI frame.
```
bjornvolcker, line 213: Is this really safe? Shouldn't you break at ANY sei but rather THE sei which equals the input?

lib/src/signed_video_h26x_auth.c
bjornvolcker: Well, you added back an empty line **before** `item->used_in_gop_hash = true;` instead of **after**...

bjornvolcker (changes requested):
A few remaining unresolved review comments

lib/src/signed_video_internal.h
```diff
@@ -144,11 +144,9 @@ struct _signed_video_t {
   bool is_golden_sei;  // Flag that tells if a SEI is a golden SEI
   bool using_golden_sei;  // Flag that tells if golden SEI prinsiple is used
   bool signing_started;
-  bool generate_sei;  // A flag that tells whether to generate SEI
-  // SEI is not generated at the start of the stream, but after the first GOP ends, this flag is set
-  // to true From that point, SEI is generated at the end of each GOP and includes the GOP hash The
-  // flag remains true until the stream ends, allowing SEI to be generated as long as the GOP hash
-  // is available.
+  bool first_signing_attempt;  // Flag indicating whether to generate the first SEI. Flips to true
```
bjornvolcker, line 147: But if it flips to true after the first signing attempt the name does not match what it does. Either find a better name or flip from true to false. See further comments where it is used.

lib/src/signed_video_h26x_sign.c
```diff
@@ -518,7 +518,7 @@ signed_video_add_nalu_part_for_signing_with_timestamp(signed_video_t *self,
 
     // Finalize GOP hash and generate SEI if the NALU is I frame of a non-empty GOP.
     if (nalu.is_first_nalu_in_gop && nalu.is_last_nalu_part) {
-      if (self->generate_sei) {
+      if (self->first_signing_attempt) {
```
bjornvolcker, line 521: The name of the flag indicates that this can only happen once. "if this is first signing attempt then do...", but right now it enters the if statement also on the second, third, fourth, ... attempts

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #240, opened 2024-10-09
Prints the hash and signature when signing and verifying

bjornvolcker:
Also, removes the existing prints of the hash to sign/verify.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #242, opened 2024-10-09
Uses hex print function for NALU hashes and GOP hash

bjornvolcker:
In addition prints both computed and received GOP hashes.
Has removed the intermediate prints of the recursive hash.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #244, opened 2024-10-28
Adds num_pending_seis parameter to signed_video_get_sei(...)

bjornvolcker:
This means that the user now can know how many SEIs are still in process
of being signed.

Tests have been updated with checks for correct numbers in queue.


lusikamalo (approved):
OK.
But I'm curious why you use "num_pending_seis" as a pointer?

bjornvolcker:
> OK. But I'm curious why you use "num_pending_seis" as a pointer?

It is because the user can **get** number of pending SEIs and not **set** number of pending SEIs.

mithydolphin (approved):

PR #250, opened 2024-11-04
Keeps the GOP hash flag until stats collected

bjornvolcker:
This makes sure to collect statistics correctly if several GOPs
are being validated in the same go.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #251, opened 2024-11-04
Only sanity check available SEIs if deprecated API

bjornvolcker:
If SEIs are available, but could not be fetched with the
signed_video_get_sei(...) API, since it would otherwise break the
H.26x standard, make sure to override the sanity check.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #252, opened 2024-11-04
Changes description on when to use signed_video_get_sei(...)

bjornvolcker:
The proper way is to call it BEFORE adding NAL Units for signing. Then
the user can peek the next NAL Unit to get the SEI if and only if it
can be prepended the peeked NAL Unit and at the same time follow the
H.26x standard when using AU format.

This commit only updates the description. Tests are updated
separately.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #253, opened 2024-11-04
Removes 2 auth tests

bjornvolcker:
The tests file export without dangling end and public key arrives
late, are no longer meaningful. There is currently no way to make
file export without getting a dangling end and the recurrence has
been removed.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #255, opened 2024-11-05
Tests call signed_video_get_sei() before add for signing

bjornvolcker:
This has a large impact on the tests. First, all tests streams have
been increased with a 'P' to make sure that there is a NAL Unit to
prepend a SEI with.
Secondly, the SEIs are now delayed by one frame, which affects the
validation statistics. In particular, all validations get a
pending NAL Unit (the I-frame).
Thirdly, since validation is triggered upon 2 I-frames, essentially
all validations start with an unsigned GOP. Fixing this, will be
done in a separate commit.

The common test description has been updated with these changes.

Further, almost all tests now have the same structure describing
the validation flow.

### DIFF
--- a/tests/check/check_signed_video_auth.c
+++ b/tests/check/check_signed_video_auth.c
@@ -48,6 +48,23 @@ teardown()
 {
 }
 
+/* General comments to the validation tests.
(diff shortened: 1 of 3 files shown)


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #256, opened 2024-11-05
Adds payload_offset to signed_video_get_sei()

bjornvolcker:
It is in many situation convenient to know where the actual SEI
payload starts in the SEI data. In particular, when adding SEIs to
the stream through the encoder, the encoder itself adds header,
payload size and emulation prevention bytes. Knowing in advance
where the payload starts avoids searching the SEI data for it.


lusikamalo (commented):
Very "minor" comment.

mithydolphin (approved):

lusikamalo (approved):
OK.

PR #261, opened 2024-11-13
Reads signing key from file

bjornvolcker:
Instead of generating new keys in the tests, the repository stores
signing keys. These are software keys used in tests only and hence
have no security value.

Four keys are added in tests/; 2 'correct' signing keys (EC and RSA)
and 2 'wrong' signing keys.

The read function now has an input which selects between correct and
wrong keys.
A helper function to read the content of a key file has been added.

The tests have been updated accordingly. On tests now performs
correctly. It previously expected a public key change, when it in
fact was no change.


mithydolphin (approved):

tests/check/test_helpers.c
```diff
@@ -141,8 +149,71 @@ pull_seis(signed_video_t *sv, test_stream_item_t **item)
...
+  strcat(full_path, "/tests/");
+  strcat(full_path, filename);
+
+  fp = fopen(full_path, "rb");
+  if (!fp) {
+    goto done;
+  }
+
+  fseek(fp, 0L, SEEK_END);
+  size_t file_size = ftell(fp);
```
lusikamalo, line 191: I think we should check if file_size is zero(empty file) and return NOT success at this point.

Nothing will crash in the code if read file is empty but it probably good to fail here instead of forwarding the error to later.

lusikamalo (commented):
Please check comment.

lusikamalo (approved):
OK.

mithydolphin (approved):

PR #262, opened 2024-11-13
Add API and test for SEI loss handling in GOP sync

mithydolphin:
A new test simulates two consecutive missing SEIs to validate loss detection,
ensuring GOP sync is lost as expected. To support this, new api added to
check for missing SEIs and updates the GOP state accordingly.

Change-Id: Ia6e625f471d2ca6c8f647bf9d233241af39b304a

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -115,6 +115,24 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
 }
 
 /**
+ * Detects if there are any missing SEI messages based on the GOP counter and updates the GOP state.
+ */
+static void
+detect_sei_loss_and_sync_gop(signed_video_t *self)
+{
+  // Get the last GOP counter before updating.
+  uint32_t last_gop_number = self->gop_info->latest_validated_gop;
+  uint32_t exp_gop_number = last_gop_number + 1;
+  uint32_t new_gop_number = self->gop_info->global_gop_counter;
+
+  // Compare new with last number of GOPs to detect potentially lost SEIs.
+  int64_t potentially_missed_gops = (int64_t)new_gop_number - exp_gop_number;
+
+  self->gop_state.has_lost_sei =
+      (potentially_missed_gops > 0) && self->gop_info->global_gop_counter_is_synced;
+}
+
+/**
  * Compares the computed link hash with the linked hash received from the
  * SEI.
  */
@@ -628,9 +646,6 @@ validate_authenticity(signed_video_t *self)
   } else {
     verify_success = verify_hashes_with_sei(self, &num_expected_nalus, &num_received_nalus);
     // Set |latest_validated_gop| to recived gop counter for the next validation.
-    // TODO: Setting |latest_validated_gop| to the received GOP counter for the next validation
-    // can lead to a loss of sync between SEIs and their corresponding GOPs in certain
-    // scenarios(such as losing 2 SEIs). This will be addressed in future changes.
     self->gop_info->latest_validated_gop = self->gop_info->global_gop_counter;
   }
 
@@ -879,6 +894,8 @@ prepare_for_validation(signed_video_t *self)
       if (self->gop_info->signature_hash_type == DOCUMENT_HASH) {
         memcpy(verify_data->hash, sei->hash, hash_size);
       }
+    } else if (sei && sei->has_been_decoded) {
+      detect_sei_loss_and_sync_gop(self);
     }
 
     if (sei) {
--- a/tests/check/check_signed_video_auth.c
+++ b/tests/check/check_signed_video_auth.c
@@ -707,6 +707,39 @@ START_TEST(remove_the_gi_nalus)
 }
 END_TEST
 
+START_TEST(two_sei_is_lost)
+{
+  test_stream_t *list = create_signed_nalus("IPPIPPIPPIPPIPPIPPIP", settings[_i]);
+  test_stream_check_types(list, "IPPISPPISPPISPPISPPISPPISP");
+
+  // Remove the third 'I': IPPISPP I SPPISPPISPPISPPISP
+  int remove_nalu_number = 9;
+  remove_item_then_check_and_free(list, remove_nalu_number, 'S');
+  test_stream_check_types(list, "IPPISPPIPPISPPISPPISPPISP");
+  remove_nalu_number = 12;
+  remove_item_then_check_and_free(list, remove_nalu_number, 'S');
+  test_stream_check_types(list, "IPPISPPIPPIPPISPPISPPISP");
+
+  signed_video_accumulated_validation_t final_validation = {
+      SV_AUTH_RESULT_NOT_OK, false, 24, 21, 3, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
+  // IPPISPPIPPIPPISPPISPPISP
+  //
+  // IPPI                  PPPP                      ->  (unsigned)
+  // IPPIS                 ...P.                     ->  (   valid)
+  //    ISPPIPPIPPIS          N.NNNNN...P.           ->  ( invalid)
+  //               ISPPIS               ....P.       ->  (   valid)
+  //                   ISPPIS                 ....P. ->  (   valid)
+  const struct validation_stats expected = {.valid_gops = 3,
+      .invalid_gops = 1,
+      .pending_nalus = 8,
+      .unsigned_gops = 1,
+      .final_validation = &final_validation};
+  validate_nalu_list(NULL, list, expected, true);
+
+  test_stream_free(list);
+}
+END_TEST
+
 /* Test description
  * Verify that we can validate authenticity if the SEI arrives late. This is simulated by
  * moving the SEI to a 'P' in the next GOP.
@@ -2120,6 +2153,7 @@ signed_video_suite(void)
   tcase_add_loop_test(tc, remove_the_g_nalu, s, e);
   tcase_add_loop_test(tc, remove_the_i_nalu, s, e);
   tcase_add_loop_test(tc, remove_the_gi_nalus, s, e);
+  tcase_add_loop_test(tc, two_sei_is_lost, s, e);
   tcase_add_loop_test(tc, sei_arrives_late, s, e);
   tcase_add_loop_test(tc, all_seis_arrive_late, s, e);
   tcase_add_loop_test(tc, all_seis_arrive_late_first_gop_scrapped, s, e);


tests/check/check_signed_video_auth.c
```diff
@@ -707,6 +707,39 @@ START_TEST(remove_the_gi_nalus)
 }
 END_TEST
 
+START_TEST(two_sei_is_lost)
```
bjornvolcker, line 710: grammar: Rename to "two_lost_seis"

tests/check/check_signed_video_auth.c
```diff
@@ -707,6 +707,39 @@ START_TEST(remove_the_gi_nalus)
...
+{
+  test_stream_t *list = create_signed_nalus("IPPIPPIPPIPPIPPIPPIP", settings[_i]);
+  test_stream_check_types(list, "IPPISPPISPPISPPISPPISPPISP");
+
+  // Remove the third 'I': IPPISPP I SPPISPPISPPISPPISP
```
bjornvolcker, line 715: nit: This comment is not correct

tests/check/check_signed_video_auth.c
```diff
@@ -707,6 +707,39 @@ START_TEST(remove_the_gi_nalus)
...
+  // IPPISPPIPPIPPISPPISPPISP
+  //
+  // IPPI                  PPPP                      ->  (unsigned)
+  // IPPIS                 ...P.                     ->  (   valid)
+  //    ISPPIPPIPPIS          N.NNNNN...P.           ->  ( invalid)
+  //               ISPPIS               ....P.       ->  (   valid)
+  //                   ISPPIS                 ....P. ->  (   valid)
+  const struct validation_stats expected = {.valid_gops = 3,
+      .invalid_gops = 1,
+      .pending_nalus = 8,
```
bjornvolcker, line 734: nit: Write this as "4 + 4" to separate the first unsigned result from the rest. Just like all other tests are written.

lib/src/signed_video_h26x_auth.c
```diff
@@ -114,6 +114,24 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
   return status;
 }
 
+/**
+ * Detects if there are any missing SEI messages based on the GOP counter and updates the GOP state.
+ */
+static void
+detect_sei_loss_and_sync_gop(signed_video_t *self)
```
bjornvolcker, line 121: I find this function name misleading. The name indicates that the GOP is sync'ed, but that is not true. The only thing done is to detect a lost SEI. How about renaming to `detect_lost_sei`

lib/src/signed_video_h26x_auth.c
```diff
@@ -114,6 +114,24 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+{
+  // Get the last GOP counter before updating.
```
bjornvolcker, line 123: I don't see any "updates" of the `latest_validated_gop`. Looks like a copy-paste thing.

lib/src/signed_video_h26x_auth.c
```diff
@@ -114,6 +114,24 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+  uint32_t last_gop_number = self->gop_info->latest_validated_gop;
+  uint32_t exp_gop_number = last_gop_number + 1;
```
bjornvolcker, line 125: Looks like a copy-paste thing, but why do you need `last_gop_number`? I think it is better to compute `exp_gop_number` at once.

lib/src/signed_video_h26x_auth.c
```diff
@@ -114,6 +114,24 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+  uint32_t new_gop_number = self->gop_info->global_gop_counter;
+
+  // Compare new with last number of GOPs to detect potentially lost SEIs.
+  int64_t potentially_missed_gops = (int64_t)new_gop_number - exp_gop_number;
+
+  self->gop_state.has_lost_sei =
+      (potentially_missed_gops > 0) && self->gop_info->global_gop_counter_is_synced;
```
bjornvolcker, line 132: What if `potentially_missed_gops < 0`, which can happen if there is a wraparound in `global_gop_counter`?

lib/src/signed_video_h26x_auth.c
```diff
@@ -879,6 +894,8 @@ prepare_for_validation(signed_video_t *self)
       if (self->gop_info->signature_hash_type == DOCUMENT_HASH) {
         memcpy(verify_data->hash, sei->hash, hash_size);
       }
+    } else if (sei && sei->has_been_decoded) {
```
bjornvolcker, line 897: It is starting to become too many if-statements in this function. What is done in `detect_sei_loss_and_sync_gop` below, is also done in `decode_sei_data`. It should be possible to call the same function for both cases instead of having code duplications wrapped in various if statements. This will make the code more readable and less error prone.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -118,16 +102,23 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
+  uint32_t new_gop_number = self->gop_info->global_gop_counter;
   int64_t potentially_missed_gops = (int64_t)new_gop_number - exp_gop_number;
-
+  // If number of |potentially_missed_gops| is negative, we have either lost SEIs together with a
+  // wraparound of |global_gop_counter|, or a reset of Signed Video was done on the camera. The
+  // correct number of lost SEIs is of less importance, since we only want to know If we have lost
+  // any. Therefore, make sure we map the value into the positive side only. It is possible to
+  // signal to the validation side that a reset was done on the camera, but it is still not
+  // possible to validate pending NALUs.
+  if (potentially_missed_gops < 0 && new_gop_number >= exp_gop_number)
```
bjornvolcker, line 118: How can this help to detect the difference between a wraparound and a reset? By definition, `potentially_missed_gops < 0` if and only if `exp_gop_number > new_gop_number`, which means that the second condition in this if statement can never be active.
Did you test this somehow?

tests/check/check_signed_video_auth.c
```diff
@@ -2153,15 +2144,15 @@ signed_video_suite(void)
...
+  tcase_add_loop_test(tc, two_lost_seis, s, e);
   tcase_add_loop_test(tc, sei_arrives_late, s, e);
   tcase_add_loop_test(tc, all_seis_arrive_late, s, e);
   tcase_add_loop_test(tc, all_seis_arrive_late_first_gop_scrapped, s, e);
   tcase_add_loop_test(tc, lost_g_before_late_sei_arrival, s, e);
   tcase_add_loop_test(tc, lost_g_and_gop_with_late_sei_arrival, s, e);
   tcase_add_loop_test(tc, lost_all_nalus_between_two_seis, s, e);
   tcase_add_loop_test(tc, add_one_sei_nalu_after_signing, s, e);
-  tcase_add_loop_test(tc, camera_reset_on_signing_side, s, e);
+  tcase_add_loop_test(tc, camera_reset_on_signing_side, s, 7);
```
bjornvolcker, line 2155: Revert this change

lib/src/signed_video_h26x_auth.c
```diff
@@ -883,23 +874,18 @@ prepare_for_validation(signed_video_t *self)
...
+        sei->has_been_decoded = true;
         memcpy(verify_data->hash, sei->hash, hash_size);
       }
-    } else if (sei && sei->has_been_decoded) {
-      detect_sei_loss_and_sync_gop(self);
-    }
-
-    if (sei) {
       SV_THROW(prepare_for_link_and_gop_hash_verification(self, sei));
+      detect_lost_sei_and_update_gop_state(self);
```
bjornvolcker, line 888: You have changed the order of these two operations. Is it more correct to prepare for link and gop hash verification before we know if we detect a lost sei?

lib/src/signed_video_h26x_auth.c
```diff
@@ -883,23 +874,18 @@ prepare_for_validation(signed_video_t *self)
   svrc_t status = SV_UNKNOWN_FAILURE;
   SV_TRY()
     h26x_nalu_list_item_t *sei = h26x_nalu_list_get_next_sei_item(nalu_list);
-    if (sei && !sei->has_been_decoded) {
+    if (sei) {
       // Decode the SEI and set signature->hash
```
bjornvolcker, line 878: Move this comment inside the if-statement

lib/src/signed_video_h26x_auth.c
```diff
@@ -118,16 +102,23 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
  * Detects if there are any missing SEI messages based on the GOP counter and updates the GOP state.
  */
 static void
-detect_sei_loss_and_sync_gop(signed_video_t *self)
+detect_lost_sei_and_update_gop_state(signed_video_t *self)
```
bjornvolcker, line 105: The name is still a little bit misleading. There is only one gop_state that is updated and there is a gop_state_update(...) function as well...

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
```diff
@@ -102,21 +102,19 @@ decode_sei_data(signed_video_t *self, const uint8_t *payload, size_t payload_siz
...
-  // signal to the validation side that a reset was done on the camera, but it is still not
-  // possible to validate pending NALUs.
-  if (potentially_missed_gops < 0 && new_gop_number >= exp_gop_number)
-    potentially_missed_gops += INT64_MAX;
+  // To estimate whether a wraparound has occurred, we check if the adjusted value
+  // is within a specific range that indicates a likely wraparound. If so, we adjust
+  // the value accordingly. This approach cannot definitively differentiate between
+  // a reset and a wraparound but provides a practical heuristic for handling the situation.
+  bool is_wraparound = (potentially_missed_gops + INT64_MAX) < (INT64_MAX / 2);
+  if (is_wraparound) potentially_missed_gops = +INT64_MAX;
```
bjornvolcker, line 117: This looks wrong. Shouldn't it be `potentionally_missed_gops += INT64`
Also, make a line break after the if.

Now I'm going to make life miserable for you. What happens if someone have interchanged two SEI frames? That belongs to a separate PR, though.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
mithydolphin: There is no macro defined as INT64  theri is only INT64_MAX

PR #264, opened 2024-11-28
Remove unused flags from old linking approach

mithydolphin:
With the transition to the new linking principle now complete, this commit removes
flags that were needed for the previous approach. These flags are no longer relevant
or used in the current implementation.

Change-Id: I6adb35497df6a7845225534a9ac4fe18980007c7

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -366,8 +366,8 @@ verify_hashes_with_hash_list(signed_video_t *self,
         latest_match_idx = compare_idx;
         num_invalid_nalus_since_latest_match = 0;
         // If the order is not correct, the validation status of the first NALU in the GOP should be
-        // 'N'. If that is the case, set |first_verification_not_authentic| to true and set
-        // |order_ok| to true for the next NALUs, so they are not affected by this issue.
+        // 'N'. If that is the case, set |order_ok| to true for the next NALUs, so they are not
+        // affected by this issue.
         if (!order_ok) {
           order_ok = true;
         }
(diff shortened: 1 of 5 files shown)


lib/src/signed_video_h26x_auth.c
```diff
@@ -1025,18 +1023,6 @@ maybe_validate_gop(signed_video_t *self, h26x_nalu_t *nalu)
       // validation in the middle of a stream. Now it is time to reset it.
       validation_flags->is_first_validation = !validation_flags->signing_present;
 
-      if (validation_flags->reset_first_validation) {
-        validation_flags->is_first_validation = true;
```
bjornvolcker, line 1029: Should you really remove the reset? This is used to handle the case where there are more than one SEI in the beginning of a stream that cannot be used. The scrapped test for example.

bjornvolcker (changes requested):

lib/src/signed_video_h26x_auth.c
mithydolphin: But as I understand its for the first validation and  resetting  the first validation is done without a flag in h26x_nalu_list_remove_missing_items function.

bjornvolcker: Hmm, I'm not 100% convinced. Here is an example
If `IPIPPIPPPIPPPPIP`
is signed into
`IPIPPISPSPPISPPPPISP`
and remove the first 2 GOPs to
`IsPsPPISPPPPISP`
then the first 2 `s` do not belong to this recording. In this case the first 2 validations need to be discarded.
Make a test of this

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -1053,6 +1053,67 @@ START_TEST(add_one_sei_nalu_after_signing)
...
+      SV_AUTH_RESULT_NOT_OK, false, 19, 16, 3, SV_PUBKEY_VALIDATION_NOT_FEASIBLE, true, 0, 0};
+  // ISPPSPISPPPPISPPISP
+  //
+  // IS                PU                      -> (signature)
+  // ISPPS             NUNN.                   -> (invalid)
+  //     PIS           MMMNP.                  -> (invalid)
+  //      ISPPPPIS              ......P.       -> (valid)
+  //            ISPPIS                 ....P.  -> (valid)
+  struct validation_stats expected = {.valid_gops = 2,
+      .invalid_gops = 2,
```
bjornvolcker, line 1095: Nothing is wrong with this stream. I would expect all of them to be valid.

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #266, opened 2024-12-06
Synchronizes SEIs for legacy code

bjornvolcker:
At validation starts, SEIs are not synchronized with an I-frame. It is
unknown how many SEIs that cannot be used to validate frames, since
they sign GOPs not present.

This commit loops through all possible validation cases with the SEI
by temporarily use tmp_validation_status and a few more flags. The
validation_status is then updated when the SEI is not feasible to use
in a validation.


mithydolphin (approved):
Looks good to me!

lusikamalo (approved):
OK.

PR #267, opened 2024-12-10
Adds two tests with two useless SEIs

bjornvolcker:
The library is not working for file exports where arbitrary number
of useless SEIs are present in the beginning. A useless SEI is a
SEI that actually belongs to a snippet of the stream present before
the recording.
The library can handle one such SEI, but not two. Therefore, this
commit adds two tests that have two useless SEIs in the beginning,
one for a stream start and one for a file export.

The test helper function generate_delayed_sei_list() has been
extended with two more GOPs to catch if tests are recovered to
produce "valid" outputs.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #269, opened 2024-12-12
Validates using tmp_validation_status

bjornvolcker:
All operations when validating a GOP is done on tmp_validation_status.
A function that either updates validation_status or resets
tmp_validation_status has been added.
Operations that are done by the authenticity_report, like removing
validated items, are done on validation_status.

This commit does not change any behavior.


lusikamalo (approved):
OK.

mithydolphin (approved):

PR #272, opened 2024-12-13
Adds update function for SEIs in validation

bjornvolcker:
Currently has no impact since no SEIs are marked.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #274, opened 2024-12-13
Synchronizes SEIs at beginning of a recording

bjornvolcker:
When validating a stream/recording SEIs are not synchronized
with the I-frames. There can for examples be multiple SEIs
in the beginning that were generated by frames prior to the
frames present in the recording.

This commit test validates each SEI for every possible way
before either accepting or discarding it.


mithydolphin (approved):

lusikamalo (approved):
OK.

PR #281, opened 2024-12-17
Initializes the GOP hash with a salt

bjornvolcker:
Also removes the dead code of the gop_hash_init from the
deprecated recursive procedure.


lusikamalo (approved):
OK.

PR #286, opened 2024-12-18
Adds same auth report logic to legacy code

bjornvolcker:
that is, only reporting unsigned in accumulated, and signed only
once.


lusikamalo (approved):
OK.

PR #287, opened 2024-12-19
API change signed_video_get_sei(...)

bjornvolcker:
Now returns a pointer to the SEI/OBU Metadata memory. The memory
is transferred and the user is responsible for freeing it.
Also, payload_offset is made unsigned, since it cannot take
negative values.

Version is bumped to v2.0.2.


lusikamalo (approved):
OK.

PR #289, opened 2025-01-07
Remove unused second_hash and has_reference_hash members

mithydolphin:
The second_hash member is no longer required as linking is now handled
with a separate hash in the TLV instead of being part of the hash list.
Consequently, the I-frame is now hashed in only one way. Additionally,
the has_reference_hash member has been removed as part of this cleanup.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/signed_video_h26x_auth.c
+++ b/lib/src/signed_video_h26x_auth.c
@@ -171,7 +171,7 @@ update_link_hash_for_auth(signed_video_t *self)
   while (item) {
     if (item->used_in_gop_hash && item->nalu->is_first_nalu_in_gop) {
       if (!item->used_for_linked_hash) {
-        update_linked_hash(self, item->second_hash, hash_size);
+        update_linked_hash(self, item->hash, hash_size);
         item->used_for_linked_hash = true;
       }
       break;
@@ -244,7 +244,7 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
       if (item == sei) {
         break;  // Break if encountered SEI frame.
       }
-      hash_to_add = item->nalu->is_first_nalu_in_gop ? item->second_hash : item->hash;
+      hash_to_add = item->hash;
       // Since the GOP hash is initialized, it can be updated with each incoming NALU hash.
       SV_THROW(openssl_update_hash(self->crypto_handle, hash_to_add, hash_size, true));
       item->used_in_gop_hash = true;  // Mark the item as used in the GOP hash
@@ -362,9 +362,8 @@ verify_hashes_with_hash_list(signed_video_t *self,
     }
     num_verified_hashes++;
 
-    // Fetch the |hash_to_verify|, which normally is the item->hash, but if this is NALU is an
-    // I frame we use item->second_hash.
-    uint8_t *hash_to_verify = item->nalu->is_first_nalu_in_gop ? item->second_hash : item->hash;
+    // Fetch the |hash_to_verify|
+    uint8_t *hash_to_verify = item->hash;
     // Compare |hash_to_verify| against all the |expected_hashes| since the |latest_match_idx|. Stop
     // when we get a match or reach the end.
     compare_idx = latest_match_idx + 1;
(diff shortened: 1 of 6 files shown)


lib/src/signed_video_h26x_auth.c
```diff
@@ -244,7 +244,7 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, h26x_nalu_list_
       if (item == sei) {
         break;  // Break if encountered SEI frame.
       }
-      hash_to_add = item->nalu->is_first_nalu_in_gop ? item->second_hash : item->hash;
+      hash_to_add = item->hash;
```
bjornvolcker, line 247: There is no need for `hash_to_add` anymore. Use `item->hash` directly instead.

lib/src/signed_video_h26x_auth.c
```diff
@@ -362,9 +362,8 @@ verify_hashes_with_hash_list(signed_video_t *self,
     }
     num_verified_hashes++;
 
-    // Fetch the |hash_to_verify|, which normally is the item->hash, but if this is NALU is an
-    // I frame we use item->second_hash.
-    uint8_t *hash_to_verify = item->nalu->is_first_nalu_in_gop ? item->second_hash : item->hash;
+    // Fetch the |hash_to_verify|
+    uint8_t *hash_to_verify = item->hash;
```
bjornvolcker, line 366: No need to use `hash_to_verify` anymore. Use `item->hash` directly instead.

lib/src/signed_video_h26x_auth.c
```diff
@@ -415,7 +414,7 @@ verify_hashes_with_hash_list(signed_video_t *self,
     // We do not know where in the sequence of NALUs they were lost. Simply add them before the
     // first item. If the first item needs a second opinion, that is, it has already been verified
     // once, we append that item. Otherwise, prepend it with missing items.
-    const bool append = !!nalu_list->first_item->second_hash;
+    const bool append = !!nalu_list->first_item->nalu->is_first_nalu_in_gop;
```
bjornvolcker, line 417: No need to use `!!`, since `is_first_nalu_in_gop` is already boolean.

When I read the comment above (which needs a rewrite) this is about where to add "missing items" (M) in the case when no hash at all could be matched. Currently, the Ms are added at the beginning if the first item is an I-frame, otherwise after the first item. So if the first item is a PPS it is being inserted between PPS and SPS. I don't think that will actually have an impact, so could we always add it first?
Another question is what happens if we validate 2 GOPs which can happen when the SEIs are not synched to the I-frames. If we end up here during the second round of validation, the missing items will be added before the first GOP and not the second GOP. Is that OK?

bjornvolcker (changes requested):
A few minor comments, and one bigger thought on where to add missing items which should be discussed.

lib/src/signed_video_h26x_auth.c
mithydolphin: For the second question I don’t think it’s OK either. I’m working on fixing it.Should I include it in this PR or handle it in a separate one after this?

bjornvolcker: It is a corner case, but I think it is covered in one of the "2 invalid SEIs at the beginning", which pass...
So, I'm not sure there is anything to fix (unless you convince me), but we should at least figure out if it can happen and avoid reporting INVALID or VALID WITH MISSING when it actually is VALID for example.

lib/src/signed_video_h26x_auth.c
```diff
@@ -411,13 +407,9 @@ verify_hashes_with_hash_list(signed_video_t *self,
...
     DEBUG_LOG("Never found a matching hash at all");
     int num_missing_nalus = num_expected_hashes - num_invalid_nalus_since_latest_match;
-    // We do not know where in the sequence of NALUs they were lost. Simply add them before the
-    // first item. If the first item needs a second opinion, that is, it has already been verified
-    // once, we append that item. Otherwise, prepend it with missing items.
-    const bool append = !!nalu_list->first_item->nalu->is_first_nalu_in_gop;
     // No need to check the return value. A failure only affects the statistics. In the worst case
     // we may signal SV_AUTH_RESULT_OK instead of SV_AUTH_RESULT_OK_WITH_MISSING_INFO.
-    h26x_nalu_list_add_missing(nalu_list, num_missing_nalus, append, nalu_list->first_item);
+    h26x_nalu_list_add_missing(nalu_list, num_missing_nalus, true, nalu_list->first_item);
```
bjornvolcker, line 412: I think we should add a comment like "TODO: Investigate if adding the missing items at the beginning of the list can cause any issues when validating multiple GOPs in one go."

bjornvolcker (changes requested):
One minor thing before approving

lusikamalo2 (approved):
LGTM.

bjornvolcker (approved):

PR #291, opened 2025-01-08
Requires certificate chain to be set.

bjornvolcker:
The attestation part is optional, depending on provisioning
method. This step changes the setter API to require certificate
chain in call, but not attestation.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #294, opened 2025-01-09
Adds a new vendor axis mode for tests

bjornvolcker:
Tests can now run without Axis, with Axis attestation and Axis
factory provisioned. The latter means that the public key is
part of the leaf certificate.

Helper function to read a certificate chain is added.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #295, opened 2025-01-09
Supports signing with Axis factory provisioning

bjornvolcker:
A test that verifies that the public key is not transmitted
separately, but instead in a certificate is added.


lusikamalo2 (approved):
OK.

PR #297, opened 2025-01-13
Adds a getter for the public key

bjornvolcker:
if Axis factory provisioning is used. A helper function to read
the public key from the leaf certificate is added.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #299, opened 2025-01-14
Rename non-public files for easier readability

mithydolphin:
Renamed non-public files from signed_video to sv_ for improved readability

### Describe your changes

- Fixes #(issue)


lib/src/sv_internal.h
```diff
@@ -30,7 +30,7 @@
 #include "includes/signed_video_openssl.h"  // pem_pkey_t, sign_or_verify_data_t
 #include "includes/signed_video_sign.h"  // SignedVideoAuthenticityLevel
 #include "legacy_validation.h"  // legacy_sv_t
-#include "signed_video_defines.h"  // svrc_t, sv_tlv_tag_t
+#include "sv_defines.h"  // svrc_t, sv_tlv_tag_t
```
bjornvolcker, line 33: Below on line 54 can you update the `SIGNED_VIDEO_VERSION` to `v2.0.3`, and the same in the root meson.build file.
The reason is that this change will break the signer and validator builds in the repository signed-video-framework-examples. Incrementing the version is the only way we can tell the applications which versions work.

bjornvolcker (changes requested):
You need to bump the version.

bjornvolcker (approved):

lusikamalo2 (approved):
OK.

PR #301, opened 2025-01-27
Renames h26x_nalu_t to bu_t

bjornvolcker:
where bu is short for Bitstream Unit and can be either NALU or OBU.


lusikamalo2 (approved):
OK.

PR #341, opened 2025-01-31
Adds the possibility to delay SEIs when creating test stream

bjornvolcker:
In addition adds a |signing_frequency| member in signed_video_t
which is not used in practise. It will have an impact when the
library supports signing multiple GOPs.


lusikamalo2 (approved):
OK.

PR #342, opened 2025-01-31
Changes mimic_file_export

bjornvolcker:
As a preparation for future use when signing partial and multiple
GOPs using different file export test streams. Also includes
scrubbing in "sign partial GOPs" tests.

The validation_stream no longer consumes the items to be able to
scrub a file.


lusikamalo2 (approved):
OK.

PR #343, opened 2025-02-03
Adds a setter for max_signing_frames

bjornvolcker:
The member name has changed from max_signing_bu because it is
easier for the user to count frames rather than Bitstream
Units.

API tests have been added.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #344, opened 2025-02-03
Adds new member to gop_info_t

bjornvolcker:
The flag triggered_partial_gop is used to communicate if the SEI
was triggered by a partial GOP. It is needed for the validation
side to know that a complete GOP is not expected. If this is not
communicated there will be faulty detection of missing Bitstream
Units.

The general tag TLV has been updated with the flag. No need to
bump up the version of that tag since the default value is
'false' and what is done historically.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #345, opened 2025-02-03
Moves update_linked_hash to hash functions

bjornvolcker:
This will simplify updating the linked hash when signing
partial GOPs.


lusikamalo2 (approved):
OK.

PR #346, opened 2025-02-04
Renames some gop_info_t members

bjornvolcker:
Then naming now reflects partial GOPs rather than (full) GOPs.


lusikamalo2 (approved):
OK.

PR #347, opened 2025-02-04
Remove unused member signature_hash_type

mithydolphin:
The member signature_hash_type was unused in the codebase and
has been removed

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/sv_tlv.c
+++ b/lib/src/sv_tlv.c
@@ -913,7 +913,7 @@ decode_signature(signed_video_t *self, const uint8_t *data, size_t data_size)
   svrc_t status = SV_UNKNOWN_FAILURE;
 
   SV_TRY()
-    SV_THROW_IF(version > 0 && version < 3, SV_OK);
+    SV_THROW_IF(version < 1 || version > 2, SV_INCOMPATIBLE_VERSION);
     SV_THROW_IF(max_signature_size < signature_size, SV_AUTHENTICATION_ERROR);
     if (!*signature_ptr) {
       verify_data->max_signature_size = 0;
--- a/lib/src/sv_tlv.c
+++ b/lib/src/sv_tlv.c
@@ -842,7 +842,7 @@ encode_signature(signed_video_t *self, uint8_t *data)
   gop_info_t *gop_info = self->gop_info;
   sign_or_verify_data_t *sign_data = self->sign_data;
   size_t data_size = 0;
-  const uint8_t version = 1;  // Increment when the change breaks the format
+  const uint8_t version = 2;  // Increment when the change breaks the format
 
   // Value fields:
   //  - version (1 byte)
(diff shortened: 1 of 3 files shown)


lib/src/sv_tlv.c
```diff
@@ -876,7 +870,6 @@ encode_signature(signed_video_t *self, uint8_t *data)
   // Write info field
   write_byte(last_two_bytes, &data_ptr, gop_info->encoding_status, epb);
   // Write hash type
-  write_byte(last_two_bytes, &data_ptr, (uint8_t)gop_info->signature_hash_type, epb);
```
bjornvolcker, line 879: Simply removing data from the TLV will break backward compatibility. Use the `version` to handle both cases.

bjornvolcker (changes requested):

lib/src/sv_tlv.c
```diff
@@ -898,6 +898,7 @@ decode_signature(signed_video_t *self, const uint8_t *data, size_t data_size)
   uint8_t **signature_ptr = &verify_data->signature;
   uint8_t version = *data_ptr++;
   uint8_t encoding_status = *data_ptr++;
+  if(version < 2) (void)(*data_ptr++); //If the version is older than 2 move past the hash_type byte.
```
bjornvolcker, line 901: I think it becomes more clear if you do this after the declarations, like
```
if (version < 2) {
  // Move past the written hash type since it is never used.
  data_ptr++;
}
```
or if you we should be really strict, you should check the hash_type value and make sure it equals 1.

bjornvolcker (changes requested):

lib/src/sv_tlv.c
```diff
@@ -918,7 +914,6 @@ decode_signature(signed_video_t *self, const uint8_t *data, size_t data_size)
 
   SV_TRY()
     SV_THROW_IF(version == 0, SV_INCOMPATIBLE_VERSION);
```
bjornvolcker, line 916: Update this check so that versions 1 and 2 are OK. See `decode_general()`

bjornvolcker (changes requested):

lusikamalo2 (approved):
OK.

lib/src/sv_tlv.c
```diff
@@ -913,7 +913,7 @@ decode_signature(signed_video_t *self, const uint8_t *data, size_t data_size)
   svrc_t status = SV_UNKNOWN_FAILURE;
 
   SV_TRY()
-    SV_THROW_IF(version == 0, SV_INCOMPATIBLE_VERSION);
+    SV_THROW_IF(version > 0 && version < 3, SV_OK);
```
bjornvolcker, line 916: This will not work. This will return `SV_OK` all the time. You should provide an error message, not an everything-is-fine-message

bjornvolcker (changes requested):

bjornvolcker (approved):

lusikamalo2 (approved):
OK.

bjornvolcker (approved):

PR #348, opened 2025-02-04
Enables signing partial GOPs

bjornvolcker:
Adds a counter for number of frames in a partial GOP. This is
needed since the setting of partial GOP size is determined in
frames instead of Bitstream Units for ease of use.

Signing side tests have been updated. Further, tests now also
verifies signing partial GOPs for multi-sliced streams.
Note that the validation side cannot handle signed partial
GOPs yet.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #352, opened 2025-02-05
Changes the helper function that modifies an item

bjornvolcker:
Now adds 128 to the item id. Previously added 1, which meant
that it could become identical with the next item in the list,
affecting validation by reporting missing instead of invalid.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #355, opened 2025-02-05
Updates use_in_gop_hash wrt partial GOPs

bjornvolcker:
When preparing for validation the number of BUs marked now
depends on the number of sent BUs if a partial GOP was triggered.
If validation fails, the number of marked items are extended to
avoid aborting validation too early.
Last, when verifying individual hashes the number of verified
ones will not exceed the number sent.

### DIFF
--- a/lib/src/sv_auth.c
+++ b/lib/src/sv_auth.c
@@ -201,10 +201,11 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, bu_list_item_t 
   assert(self);
 
   // Initialize pointers and variables
+  gop_info_t *gop_info = self->gop_info;
   bu_list_t *bu_list = self->bu_list;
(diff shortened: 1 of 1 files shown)


lib/src/sv_auth.c
```diff
@@ -230,18 +231,23 @@ prepare_for_link_and_gop_hash_verification(signed_video_t *self, bu_list_item_t
       if (num_i_frames > 1) break;  // Break if encountered second I frame.
       // Break at I-frame if NAL Units have been added to GOP hash, since a GOP hash cannot span
```
lusikamalo2, line 232: NAL -> Bitstream

lusikamalo2 (dismissed):
OK.

mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #361, opened 2025-02-12
Statically allocates memory for product info

bjornvolcker:
When the signed_video_t object is created the
signed_video_product_info_t is statically allocated where
all strings have a preset fixed length of 255 bytes. This
makes it more clear to the user that strings longer than
255 characters will be truncated.

In the TLV, the strings are now encoded without null-terminated
characters. The operation is backward compatible.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #362, opened 2025-02-13
Computes correct sei size for memory allocation

bjornvolcker:
A copy-paste error computed a much larger size than necessary
for the SEIs.


lusikamalo2 (approved):
OK.

PR #363, opened 2025-02-13
Moves Axis vendor tag to optional tags

bjornvolcker:
This commit lets the TLV code handle vendor tags. The Axis
vendor tag is now part of the optional tags and if built
without Axis Communication, the encoder will be empty and
by default not be added.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #367, opened 2025-02-17
Add functions to convert between timestamps

mithydolphin:
Add functions to convert between Unix and 1601-based timestamps

Change-Id: I0baf3c9a4b1d34ecde114fcb7344b4b97785ec21

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/sv_common.c
+++ b/lib/src/sv_common.c
@@ -158,6 +158,20 @@ sign_or_verify_data_free(sign_or_verify_data_t *self)
   free(self);
 }
 
+// Convert Unix timestamp (seconds since 1970) to 1601-based timestamp
+void
+convert_unix_to_1601(int64_t *timestamp)
+{
+  *timestamp = (*timestamp + EPOCH_DIFF) * HUNDRED_NANOSECONDS;
+}
+
+// Convert 1601-based timestamp (100-nanosecond intervals since 1601) to Unix timestamp
+void
+convert_1601_to_unix(int64_t *timestamp)
+{
+  *timestamp = (*timestamp / HUNDRED_NANOSECONDS) - EPOCH_DIFF;
+}
+
 /* Reads the version string and puts the Major.Minor.Patch in the first, second and third element of
  * the array, respectively */
 static bool
--- a/lib/src/sv_common.c
+++ b/lib/src/sv_common.c
@@ -158,18 +158,20 @@ sign_or_verify_data_free(sign_or_verify_data_t *self)
   free(self);
 }
 
-// Convert Axis Unix timestamp (microseconds since 1970) to 1601-based timestamp (100-ns intervals)
+// Convert Unix timestamp (microseconds since 1970) to 1601-based timestamp (100-nanosecond
+// intervals)
 int64_t
-convert_unix_to_1601(int64_t timestamp)
+convert_unix_us_to_1601(int64_t timestamp)
 {
-   return (timestamp + (EPOCH_DIFF * 1000000LL)) * MICROSECONDS;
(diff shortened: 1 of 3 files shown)


lib/src/sv_common.c
```diff
@@ -158,6 +158,20 @@ sign_or_verify_data_free(sign_or_verify_data_t *self)
   free(self);
 }
 
+// Convert Unix timestamp (seconds since 1970) to 1601-based timestamp
+void
+convert_unix_to_1601(int64_t *timestamp)
```
bjornvolcker, line 163: I think it is more clear to return the "new" timestamp instead of passing in a pointer. Then you can use the function inside another function like

```
convert_1601_to_unix(convert_unix_to_1601(timestamp));
```
What do you think?

lib/src/sv_common.c
```diff
@@ -158,6 +158,20 @@ sign_or_verify_data_free(sign_or_verify_data_t *self)
...
+{
+  *timestamp = (*timestamp + EPOCH_DIFF) * HUNDRED_NANOSECONDS;
```
bjornvolcker, line 165: The Axis Unix Epoch is in microseconds, not seconds. I see that the information is missing in signed_video_sign.h, but present in signed_video_auth.h

lib/src/sv_common.c
```diff
@@ -158,6 +158,20 @@ sign_or_verify_data_free(sign_or_verify_data_t *self)
...
+}
+
+// Convert 1601-based timestamp (100-nanosecond intervals since 1601) to Unix timestamp
+void
+convert_1601_to_unix(int64_t *timestamp)
```
bjornvolcker, line 170: How about adding a test for this and put it in check_signed_video_common.c

bjornvolcker (changes requested):

lib/src/sv_common.c
```diff
@@ -158,18 +158,18 @@ sign_or_verify_data_free(sign_or_verify_data_t *self)
   free(self);
 }
 
-// Convert Unix timestamp (seconds since 1970) to 1601-based timestamp
-void
-convert_unix_to_1601(int64_t *timestamp)
+// Convert Axis Unix timestamp (microseconds since 1970) to 1601-based timestamp (100-ns intervals)
```
bjornvolcker, line 161: This is not Axis specific. The Unix Epoch time can be interpreted in seconds, ms, us, ns depending on the number of digits. See for example https://www.epochconverter.com/

If you want to build a true converter you should analyze the timestamp to determine in what scale it represents the time.
I think it is fine here to assume microseconds, but maybe name the function `convert_unix_us_to_1601(...)`

lib/src/sv_internal.h
```diff
@@ -69,8 +69,7 @@
 #define HASH_LIST_SIZE (MAX_HASH_SIZE * MAX_GOP_LENGTH)
 
 #define EPOCH_DIFF 11644473600LL  // Difference between 1601 and 1970 in seconds
```
bjornvolcker, line 71: Can't you represent this in microseconds at once?

lib/src/sv_internal.h
```diff
@@ -69,8 +69,7 @@
...
-#define HUNDRED_NANOSECONDS 10000000LL  // 1 second = 10^7 * 100-nanosecond
-
+#define MICROSECONDS 10LL         // 1 microsecond = 10 * 100-nanoseconds
```
bjornvolcker, line 72: This name is misleading. Maybe `MICROSEC_TO_100_NANOSEC` or shorter if possible

bjornvolcker (changes requested):

bjornvolcker (approved):

PR #368, opened 2025-02-18
Add Media Signing Return Code Converter

mithydolphin:
Ensure onfiv directory exists, add safeguards to compile even if
it doesn't, and implement a converter for media signing return codes
to sign video return codes.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/includes/signed_video_common.h
+++ b/lib/src/includes/signed_video_common.h
@@ -21,19 +21,6 @@
 
 #ifndef __SIGNED_VIDEO_COMMON_H__
 #define __SIGNED_VIDEO_COMMON_H__
-#ifndef HAS_ONVIF
-// Define a fallback version if ONVIF is missing
-typedef enum {
-  OMS_OK = 0,
-  OMS_MEMORY = -1,
-  OMS_INVALID_PARAMETER = -10,
-  OMS_NOT_SUPPORTED = -12,
-  OMS_INCOMPATIBLE_VERSION = -15,
-  OMS_EXTERNAL_ERROR = -20,
-  OMS_AUTHENTICATION_ERROR = -30,
-  OMS_UNKNOWN_FAILURE = -100
-} MediaSigningReturnCode;
-#endif
 
 #include <stdbool.h>
 #include <stddef.h>
--- a/lib/src/includes/signed_video_common.h
+++ b/lib/src/includes/signed_video_common.h
@@ -21,6 +21,19 @@
 
 #ifndef __SIGNED_VIDEO_COMMON_H__
 #define __SIGNED_VIDEO_COMMON_H__
+#ifndef HAS_ONVIF
+// Define a fallback version if ONVIF is missing
+typedef enum {
+  OMS_OK = 0,
+  OMS_MEMORY = -1,
+  OMS_INVALID_PARAMETER = -10,
+  OMS_NOT_SUPPORTED = -12,
+  OMS_INCOMPATIBLE_VERSION = -15,
+  OMS_EXTERNAL_ERROR = -20,
+  OMS_AUTHENTICATION_ERROR = -30,
+  OMS_UNKNOWN_FAILURE = -100
+} MediaSigningReturnCode;
+#endif
 
 #include <stdbool.h>
 #include <stddef.h>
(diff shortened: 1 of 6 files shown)


lib/src/includes/signed_video_common.h
```diff
@@ -21,6 +21,19 @@
 
 #ifndef __SIGNED_VIDEO_COMMON_H__
 #define __SIGNED_VIDEO_COMMON_H__
+#ifndef HAS_ONVIF
+// Define a fallback version if ONVIF is missing
+typedef enum {
```
bjornvolcker, line 26: This cannot be in a public header file. Initially it can be placed in sv_internal.h, but eventually we might use a sv_onvif.h file to put all ONVIF related stuff.

lib/src/sv_internal.h
```diff
@@ -32,6 +32,10 @@
 #include "legacy_validation.h"  // legacy_sv_t
 #include "sv_defines.h"  // svrc_t, sv_tlv_tag_t
 
+#ifndef HAS_ONVIF
+// If ONVIF is missing, define it as an alias to signed_video_t
+typedef struct signed_video_t onvif_media_signing_t;
```
bjornvolcker, line 37: It might be better to depend on something more standard like `void` instead of `struct signed_video_t`

lib/src/sv_internal.h
```diff
@@ -337,6 +341,8 @@ struct _signed_video_t {
 
   // Legacy validation
   legacy_sv_t *legacy_sv;
+  // Onviff validation and signing
```
bjornvolcker, line 344: nit: ONVIF

lib/src/sv_internal.h
```diff
@@ -420,6 +426,11 @@ bu_type_to_str(const bu_info_t *bu);
 char
 bu_type_to_char(const bu_info_t *bu);
 
+/**
+ * Converts a MediaSigningReturnCode to a SignedVideoReturnCode. */
+SignedVideoReturnCode
+media_signing_return_code_to_signed_video_return_code(int code);
```
bjornvolcker, line 432: Since this is an internal function this name is a little bit too long. We use `svrc` as a short name for `SignedVideoReturnCode` internally. How about naming this `msrc_to_svrc(MediaSigningReturnCode msrc)`?

Note that you should for clarity use the full name instead of `int`

meson.build
```diff
@@ -38,6 +38,22 @@ elif (signing_plugin == 'threaded_unless_check_dep')
   signing_plugin = 'threaded'
 endif
 
+onvif_subdir = 'onvif_media_signing'
+onvif_path = meson.source_root() / onvif_subdir
+
+onvif_check = run_command('test', '-d', onvif_path, check: false)
+
+if onvif_check.returncode() == 0
```
bjornvolcker, line 46: I don't think what is inside this if-statement works. Have you tried it?

I think you at this point should keep it as simple as possible and skip all the changes in the meson.build. It should compile perfectly anyhow.

lib/src/sv_onvif.h
```diff
@@ -0,0 +1,41 @@
...
+  OMS_MEMORY = -1,
+  OMS_INVALID_PARAMETER = -10,
+  OMS_NOT_SUPPORTED = -12,
+  OMS_INCOMPATIBLE_VERSION = -15,
+  OMS_EXTERNAL_ERROR = -20,
+  OMS_AUTHENTICATION_ERROR = -30,
+  OMS_UNKNOWN_FAILURE = -100
+} MediaSigningReturnCode;
+#endif
+;
```
bjornvolcker, line 39: Remove ";"

lib/src/sv_onvif.h
```diff
@@ -0,0 +1,41 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2021 Axis Communications AB
```
bjornvolcker, line 4: I actually think it should be "2025", but I'm not sure. Let's have a look in the other public repositories.

meson.build
```diff
@@ -38,7 +38,6 @@ elif (signing_plugin == 'threaded_unless_check_dep')
   signing_plugin = 'threaded'
 endif
 
-
```
bjornvolcker, line 41: Add back this empty line to make this file untouched

PR #370, opened 2025-02-18
Ensure compilation when setting private key

mithydolphin:
When ONVIF is missing, onvif_media_signing_set_signing_key_pair() now returns
OMS_NOT_SUPPORTED (-12) to prevent compilation issues.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/sv_onvif.h
+++ b/lib/src/sv_onvif.h
@@ -21,6 +21,10 @@
 #ifndef __SV_ONVIF_H__
 #define __SV_ONVIF_H__
 
+#include <stdbool.h>  // bool
+#include <stdint.h>  // uint8_t
+#include <string.h>  // size_t
+
 #ifndef HAS_ONVIF
 // Define a placeholder for onvif_media_signing_t to avoid compilation errors
 typedef void onvif_media_signing_t;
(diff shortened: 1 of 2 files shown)


lib/src/sv_sign.c
```diff
@@ -781,6 +781,13 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
   if (!self || !private_key || private_key_size == 0) return SV_INVALID_PARAMETER;
 
   svrc_t status = SV_UNKNOWN_FAILURE;
+  // If ONVIF is available, call its function and map the return code
+  if (self->onvif) {
+    MediaSigningReturnCode onvif_status = onvif_media_signing_set_signing_key_pair(
+        self->onvif, private_key, private_key_size, NULL, 0, false);
```
bjornvolcker, line 787: Hmmm, this will not work in real life I guess. Does the ONVIF onvif_media_signing_set_signing_key_pair() API say that you can add the private key and certificate chain separately?

lib/src/sv_onvif.h
```diff
@@ -35,6 +39,26 @@ typedef enum {
...
+static inline MediaSigningReturnCode
+onvif_media_signing_set_signing_key_pair(onvif_media_signing_t *self,
+    const char *private_key,
+    size_t private_key_size,
+    const char *certificate_chain,
+    size_t certificate_chain_size,
+    bool user_provisioned)
+{
+
+  (void)self;  // Prevent unused variable warnings
```
bjornvolcker, line 53: Have a look in sv_internal.h
There is a  define for unused variable that I think is a better option. ATTR_UNUSED
Put that in a good place so it can be included and used here as well.

bjornvolcker (changes requested):

PR #376, opened 2025-02-20
Passes on BU for signing to ONVIF if available

mithydolphin:
The BU data  is passed on to the ONVIF Media Signing API using `onvif_media_signing_add_nalu_for_signing`
if ONVIF Media Signing is available.# Please enter the commit message for your changes.

Change-Id: I0b5c518692d03b494c78e61f62cb20f650df2511

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/sv_sign.c
+++ b/lib/src/sv_sign.c
@@ -466,11 +466,6 @@ prepare_for_signing(signed_video_t *self)
 SignedVideoReturnCode
 signed_video_add_nalu_for_signing(signed_video_t *self, const uint8_t *bu_data, size_t bu_data_size)
 {
-  if (self->onvif) {
-    int64_t timestamp = 0;  // Use 0 or another valid default timestamp
-    return msrc_to_svrc(
-        onvif_media_signing_add_nalu_for_signing(self, bu_data, bu_data_size, timestamp));
-  }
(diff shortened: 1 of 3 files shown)


lib/src/sv_sign.c
```diff
@@ -466,6 +466,11 @@ prepare_for_signing(signed_video_t *self)
 SignedVideoReturnCode
 signed_video_add_nalu_for_signing(signed_video_t *self, const uint8_t *bu_data, size_t bu_data_size)
 {
+  if (self->onvif) {
+    int64_t timestamp = 0;  // Use 0 or another valid default timestamp
```
bjornvolcker, line 470: No we cannot make up timestamps like this. What would a client say in court if they show a video with a timestamp set to year 1601?
ONVIF requires a valid timestamp. That is the reason why there is only one API. We added timestamps at a later stage. That is why there are two APIs in signed-video-framework.
If no timestamp is set, we should use Axis Signed Video or return an error code to alert the user.

bjornvolcker (changes requested):

bjornvolcker (commented):
LGTM. Will approve when merge conflicts have been resolved and tests pass.

bjornvolcker (approved):

PR #395, opened 2025-02-25
Transfer validation to ONVIF if detected

bjornvolcker:
If an ONVIF Media Signing SEI is detected and the library
is build with Axis Communications vendor, an ONVIF Media
Signing session is created. The trusted anchor certificate
is passed in to the new session and all queued Bitstream
Units are transferred to the new session.

Stubs have been added for compilation purpose.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #397, opened 2025-02-25
Add ONVIF SEI as a character type for future usage

mithydolphin:
Introduced 'O' as a character identifier for ONVIF SEI in the SEI type handler.

Change-Id: I5ef7c34deaee910fdb561a32ce488438e4aa04fc

### Describe your changes

- Fixes #(issue)


lib/src/sv_common.c
```diff
@@ -109,6 +109,7 @@ bu_type_to_char(const bu_info_t *bu)
 
   switch (bu->bu_type) {
     case BU_TYPE_SEI:
+      if (bu->uuid_type == UUID_TYPE_ONVIF_MEDIA_SIGNING) return 'O';
```
bjornvolcker, line 112: I don't think we should display to the user if it is a special ONVIF SEI. When validating with ONVIF Media Signing these characters will not be used anyhow.

tests/check/test_stream.c
```diff
@@ -120,6 +120,8 @@ get_type_char(const uint8_t *data, size_t data_size, SignedVideoCodec codec)
         type = 'Z';
       else if (bu.is_golden_sei)
         type = 'G';
+      else if (bu.uuid_type == UUID_TYPE_ONVIF_MEDIA_SIGNING)
+        type = 'O';
```
bjornvolcker, line 124: I'm missing an update of `test_stream_item_create_from_type()`.
If I generate a test stream like "IPPIOPPIOP" I'd expect a correct linked list of items.

bjornvolcker (changes requested):

tests/check/test_stream.c
```diff
@@ -59,6 +60,9 @@ static const uint8_t p_nalu_h264[DUMMY_NALU_SIZE] = {0x01, 0x00, 0xff, 0x00, 0x8
 static const uint8_t pps_nalu_h264[DUMMY_NALU_SIZE] = {0x28, 0x00, 0xff, 0x00, 0x80};
 static const uint8_t sei_nalu_h264[DUMMY_SEI_SIZE] = {0x06, 0x05, 0x12, 0xaa, 0xaa, 0xaa, 0xaa,
     0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0x00, 0x00, 0x80};
+static const uint8_t oms_sei_nalu_h264[DUMMY_OMS_SEI_SIZE] = {
+    0x00, 0x5b, 0xc9, 0x3f, 0x2d, 0x71, 0x5e, 0x95, 0xad, 0xa4, 0x79, 0x6f, 0x90, 0x87, 0x7a, 0x6f};
```
bjornvolcker, line 64: This will not work. It is not a complete NAL Unit. Have you tested it?

tests/check/test_stream.c
```diff
@@ -77,6 +81,8 @@ static const uint8_t p_nalu_h265[DUMMY_NALU_SIZE] = {0x02, 0x01, 0x00, 0x00, 0x8
 static const uint8_t pps_nalu_h265[DUMMY_NALU_SIZE] = {0x44, 0x01, 0x00, 0x00, 0x80};
 static const uint8_t sei_nalu_h265[DUMMY_SEI_SIZE] = {0x4e, 0x01, 0x05, 0x11, 0xaa, 0xaa, 0xaa,
     0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0x00, 0x80};
+static const uint8_t oms_sei_nalu_h265[DUMMY_OMS_SEI_SIZE] = {
+    0x00, 0x5b, 0xc9, 0x3f, 0x2d, 0x71, 0x5e, 0x95, 0xad, 0xa4, 0x79, 0x6f, 0x90, 0x87, 0x7a, 0x6f};
```
bjornvolcker, line 85: This is not working either.

tests/check/test_stream.c
```diff
@@ -192,6 +200,10 @@ test_stream_item_create_from_type(char type, uint8_t id, SignedVideoCodec codec)
       bu_data = codec == SV_CODEC_H264 ? p_nalu_h264
                                        : (codec == SV_CODEC_H265 ? p_nalu_h265 : invalid_av1);
       break;
+    case 'O':
+      bu_data = codec == SV_CODEC_H264 ? oms_sei_nalu_h264
+                                       : (codec == SV_CODEC_H265 ? oms_sei_nalu_h265 : sei_av1);
```
bjornvolcker, line 205: Right now you use a dummy SEI for AV1, which is the same as `Z`. Is that the best option? Or should it be NULL or something else?

bjornvolcker (changes requested):

tests/check/test_stream.c
```diff
@@ -30,6 +30,7 @@
 #define START_CODE_SIZE 4
 #define DUMMY_NALU_SIZE 5
 #define DUMMY_SEI_SIZE 22
+#define DUMMY_OMS_SEI_SIZE 16
```
bjornvolcker, line 33: This is not used anywhere. Remove

tests/check/test_stream.c
```diff
@@ -59,6 +60,9 @@ static const uint8_t p_nalu_h264[DUMMY_NALU_SIZE] = {0x01, 0x00, 0xff, 0x00, 0x8
 static const uint8_t pps_nalu_h264[DUMMY_NALU_SIZE] = {0x28, 0x00, 0xff, 0x00, 0x80};
 static const uint8_t sei_nalu_h264[DUMMY_SEI_SIZE] = {0x06, 0x05, 0x12, 0xaa, 0xaa, 0xaa, 0xaa,
     0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0x00, 0x00, 0x80};
+static const uint8_t oms_sei_nalu_h264[DUMMY_SEI_SIZE] = {0x00, 0x5b, 0xc9, 0x3f, 0x2d, 0x71, 0x5e,
```
bjornvolcker, line 63: This does still not follow the H.264 standard. See for example https://www.itu.int/rec/T-REC-H.264

bjornvolcker (changes requested):

tests/check/test_stream.c
```diff
@@ -116,10 +121,12 @@ get_type_char(const uint8_t *data, size_t data_size, SignedVideoCodec codec)
       type = 'V';
       break;
     case BU_TYPE_SEI: {
-      if (!bu.is_sv_sei)
+      if (!bu.is_sv_sei && !(bu.uuid_type == UUID_TYPE_ONVIF_MEDIA_SIGNING))
         type = 'Z';
       else if (bu.is_golden_sei)
         type = 'G';
+      else if (bu.uuid_type == UUID_TYPE_ONVIF_MEDIA_SIGNING)
```
bjornvolcker, line 128: If you check this first you don't need the non-check on line 124

tests/check/test_stream.c
```diff
@@ -192,6 +199,11 @@ test_stream_item_create_from_type(char type, uint8_t id, SignedVideoCodec codec)
       bu_data = codec == SV_CODEC_H264 ? p_nalu_h264
                                        : (codec == SV_CODEC_H265 ? p_nalu_h265 : invalid_av1);
       break;
+    case 'O':
+      bu_data = codec == SV_CODEC_H264 ? oms_sei_nalu_h264
+                                       : (codec == SV_CODEC_H265 ? oms_sei_nalu_h265 : NULL);
+      bu_data_size = (codec != SV_CODEC_AV1) ? DUMMY_SEI_SIZE : DUMMY_NALU_SIZE;
```
bjornvolcker, line 205: Isn't the size 0 if AV1 is selected?

bjornvolcker (changes requested):

tests/check/test_stream.c
```diff
@@ -192,6 +199,11 @@ test_stream_item_create_from_type(char type, uint8_t id, SignedVideoCodec codec)
       bu_data = codec == SV_CODEC_H264 ? p_nalu_h264
                                        : (codec == SV_CODEC_H265 ? p_nalu_h265 : invalid_av1);
       break;
+    case 'O':
+      bu_data = codec == SV_CODEC_H264 ? oms_sei_nalu_h264
+                                       : (codec == SV_CODEC_H265 ? oms_sei_nalu_h265 : NULL);
+      bu_data_size = (codec != SV_CODEC_AV1) ? DUMMY_SEI_SIZE : 0;
```
bjornvolcker, line 205: This will still create an item further down, but without valid content. Also, even if you pass in zero as size in memcpy I think the source parameter needs to be valid, that is, not NULL.

I suggest that you handle this such that this function returns no item if `bu_data = NULL`.

Look at `test_stream_create()` and how items are treated.

bjornvolcker (changes requested):

tests/check/check_signed_video_auth.c
```diff
@@ -1431,6 +1431,40 @@ START_TEST(file_export_with_two_useless_seis)
 }
 END_TEST
 
+/* Test description
+ * Verify that we do not get any authentication if the stream has onvif SEIs.
+ */
+START_TEST(onvif_seis)
+{
+  // ONVIF SEI is not supported for AV1, so skip the test in that case
```
bjornvolcker, line 1439: I think you should use the real name "ONVIF Media Signing is not supported for AV1"

tests/check/check_signed_video_auth.c
```diff
@@ -1452,6 +1452,44 @@ START_TEST(no_signature)
...
+  test_stream_item_t *item = list->first_item;
+  int count = 0;
+  SignedVideoReturnCode sv_rc;
+
+  // Iterate through the first 9 elements in the list
+  while (item && count < 9) {
+    // If the current item's type corresponds to 'O', expect SV_EXTERNAL_ERROR
+    if (item->type == 'O') {
+      sv_rc = signed_video_add_nalu_and_authenticate(sv, item->data, item->data_size, NULL);
+      ck_assert_int_eq(sv_rc, SV_EXTERNAL_ERROR);
```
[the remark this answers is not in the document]
bjornvolcker: What I mean is that the if-statement can be replaced with
```
ck_assert_int_eq(sv_rc, item->type == 'O' ? SV_EXTERNAL_ERROR : SV_OK);
```

but remember to keep the comment/description

tests/check/test_stream.c
```diff
@@ -192,6 +199,11 @@ test_stream_item_create_from_type(char type, uint8_t id, SignedVideoCodec codec)
       bu_data = codec == SV_CODEC_H264 ? p_nalu_h264
                                        : (codec == SV_CODEC_H265 ? p_nalu_h265 : invalid_av1);
       break;
+    case 'O':
+      bu_data = codec == SV_CODEC_H264 ? oms_sei_nalu_h264
+                                       : (codec == SV_CODEC_H265 ? oms_sei_nalu_h265 : invalid_av1);
```
bjornvolcker, line 204: But now you're adding an invalid OBU.

Think of this as a camera where the user has turned on ONVIF Media Signing for an AV1 recording. Since AV1 is not supported by Media Signing, the stream will contain no SEIs at all. It will definitely not add some invalid OBUs (invalid_av1) or arbitrary SEI (sei_av1).

If I ask for "IPIOPIOP" like in the added test the AV1 stream would then look like "IPIPIP", right.
Is it possible to make sure you get that behavior?

bjornvolcker (changes requested):

bjornvolcker (approved):
Much better!

lusikamalo2 (approved):
OK.

PR #399, opened 2025-02-27
Restructures meson

bjornvolcker:
The shared library is now created from root meson.
The lib/src/ directory is now added as include directory
to avoid absolute paths from root.

Tests have been updated with the changes.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #401, opened 2025-02-27
Renames internal files

bjornvolcker:
When bringing in ONVIF Media Signing as a submodule
some functions share the same name. This will cause
multiple definitions if building the library with a
populated submodule, like FilePlayer will.

This commit prepends internal functions with "sv_"
to make names unique. This only applies to function
names common to ONVIF Media Signing.

### DIFF
--- a/lib/plugins/threaded-signing/plugin.c
+++ b/lib/plugins/threaded-signing/plugin.c
@@ -20,7 +20,7 @@
  */
 
 /**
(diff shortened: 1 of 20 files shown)


mithydolphin (dismissed):

lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #402, opened 2025-02-27
Updated signed_video_set_private_key

mithydolphin:
Updated signed_video_set_private_key to work with ONVIF. Added getter
functions for the private key and certificate chain. Included a test
to verify the signing functionality.

### Describe your changes

- Fixes #(issue)

### DIFF
--- a/lib/src/sv_common.c
+++ b/lib/src/sv_common.c
@@ -35,13 +35,13 @@
 #include "includes/signed_video_helpers.h"  // onvif_media_signing_parse_sei()
 #include "includes/signed_video_openssl.h"  // pem_pkey_t, sign_or_verify_data_t
 #include "includes/signed_video_signing_plugin.h"
-#include "sv_authenticity.h"  // latest_validation_init()
+#include "sv_authenticity.h"  // sv_latest_validation_init()
 #include "sv_bu_list.h"  // bu_list_create(), bu_list_free()
 #include "sv_codec_internal.h"  // parse_h264_nalu_header(), parse_av1_obu_header()
 #include "sv_defines.h"  // svrc_t
 #include "sv_internal.h"  // gop_info_t, validation_flags_t, MAX_HASH_SIZE, DEFAULT_HASH_SIZE
 #include "sv_openssl_internal.h"
-#include "sv_tlv.h"  // read_32bits()
+#include "sv_tlv.h"  // sv_read_32bits()
 
 #define USER_DATA_UNREGISTERED 5
 #define H264_NALU_HEADER_LEN 1  // length of forbidden_zero_bit, nal_ref_idc and nal_unit_type
@@ -198,7 +198,7 @@ sign_or_verify_data_free(sign_or_verify_data_t *self)
 {
   if (!self) return;
 
-  openssl_free_key(self->key);
+  sv_openssl_free_key(self->key);
   free(self->hash);
   free(self->signature);
   free(self);
@@ -234,7 +234,7 @@ version_str_to_bytes(int *arr, const char *str)
 
 /* Puts Major, Minor and Patch from a version array to a version string */
 void
-bytes_to_version_str(const int *arr, char *str)
+sv_bytes_to_version_str(const int *arr, char *str)
 {
   if (!arr || !str) return;
   sprintf(str, "v%d.%d.%d", arr[0], arr[1], arr[2]);
@@ -400,7 +400,7 @@ remove_epb_from_sei_payload(bu_info_t *bu)
     // prevention bytes removed.
     const uint8_t *hashable_data_ptr = bu->hashable_data;
     for (size_t i = 0; i < data_size; i++) {
-      bu->nalu_data_wo_epb[i] = read_byte(&last_two_bytes, &hashable_data_ptr, true);
+      bu->nalu_data_wo_epb[i] = sv_read_byte(&last_two_bytes, &hashable_data_ptr, true);
     }
     // Point |tlv_data| to the first byte of the TLV part in |nalu_data_wo_epb|.
     bu->tlv_data = &bu->nalu_data_wo_epb[data_size - bu->payload_size + UUID_LEN];
@@ -455,7 +455,7 @@ parse_bu_info(const uint8_t *bu_data,
 
   if (codec != SV_CODEC_AV1) {
     // There is no start code for AV1.
-    read_bytes = read_32bits(bu_data, &start_code);
+    read_bytes = sv_read_32bits(bu_data, &start_code);
     if (start_code != kStartCode) {
       // Check if this is a 3 byte Start Code.
       read_bytes = 3;
@@ -675,7 +675,7 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
   // If this is the start of the GOP, initialize |crypto_handle| to enable
   // updating the hash with each received BU.
   if (*list_idx == 0) {
-    openssl_init_hash(self->crypto_handle, true);
+    sv_openssl_init_hash(self->crypto_handle, true);
   }
   // If the upcoming hash doesn't fit in the hash list buffer, set *list_idx to -1
   // to indicate that the hash list is full, and the hash list is no longer accessible.
@@ -688,7 +688,7 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
     memcpy(&hash_list[*list_idx], hash, hash_size);
     *list_idx += (int)hash_size;
   }
-  openssl_update_hash(self->crypto_handle, hash, hash_size, true);
+  sv_openssl_update_hash(self->crypto_handle, hash, hash_size, true);
 }
 
 /*
@@ -697,7 +697,7 @@ check_and_copy_hash_to_hash_list(signed_video_t *self, const uint8_t *hash, size
  * previous hash moved to the first slot.
  */
 svrc_t
-update_linked_hash(signed_video_t *self, uint8_t *hash, size_t hash_size)
+sv_update_linked_hash(signed_video_t *self, uint8_t *hash, size_t hash_size)
 {
   if (!self || !hash) return SV_INVALID_PARAMETER;
   if (self->authentication_started) {
@@ -755,7 +755,7 @@ update_hash(signed_video_t *self,
   const uint8_t *hashable_data = bu->hashable_data;
   size_t hashable_data_size = bu->hashable_data_size;
 
-  return openssl_update_hash(self->crypto_handle, hashable_data, hashable_data_size, false);
+  return sv_openssl_update_hash(self->crypto_handle, hashable_data, hashable_data_size, false);
 }
 
 /* simply_hash()
@@ -772,12 +772,12 @@ simply_hash(signed_video_t *self, const bu_info_t *bu, uint8_t *hash, size_t has
 
   if (bu->is_first_bu_part) {
     // Entire BU can be hashed in one part.
-    return openssl_hash_data(self->crypto_handle, hashable_data, hashable_data_size, hash);
+    return sv_openssl_hash_data(self->crypto_handle, hashable_data, hashable_data_size, hash);
   } else {
     svrc_t status = update_hash(self, bu, hash, hash_size);
     if (status == SV_OK) {
       // Finalize the ongoing hash of BU parts.
-      status = openssl_finalize_hash(self->crypto_handle, hash, false);
+      status = sv_openssl_finalize_hash(self->crypto_handle, hash, false);
     }
     return status;
   }
@@ -805,7 +805,7 @@ hash_and_copy_to_ref(signed_video_t *self, const bu_info_t *bu, uint8_t *hash, s
     memcpy(reference_hash, hash, hash_size);
     // Update |linked_hash| with |reference_hash| if applied on the signing side.
     if (!self->authentication_started) {
-      update_linked_hash(self, reference_hash, hash_size);
+      sv_update_linked_hash(self, reference_hash, hash_size);
     }
   SV_CATCH()
   SV_DONE(status)
@@ -840,12 +840,12 @@ hash_with_reference(signed_video_t *self,
     // Hash BU data and store as |bu_hash|.
     SV_THROW(simply_hash(self, bu, bu_hash, hash_size));
     // Hash reference hash together with the |bu_hash| and store in |buddy_hash|.
-    SV_THROW(
-        openssl_hash_data(self->crypto_handle, gop_info->hash_buddies, hash_size * 2, buddy_hash));
+    SV_THROW(sv_openssl_hash_data(
+        self->crypto_handle, gop_info->hash_buddies, hash_size * 2, buddy_hash));
     // Copy |buddy_hash| to |linked_hash| queue if signing is triggered. Only applies on
     // the signing side.
     if (gop_info->triggered_partial_gop && !self->authentication_started) {
-      update_linked_hash(self, buddy_hash, hash_size);
+      sv_update_linked_hash(self, buddy_hash, hash_size);
     }
   SV_CATCH()
   SV_DONE(status)
@@ -854,7 +854,7 @@ hash_with_reference(signed_video_t *self,
 }
 
 svrc_t
-hash_and_add(signed_video_t *self, const bu_info_t *bu)
+sv_hash_and_add(signed_video_t *self, const bu_info_t *bu)
 {
   if (!self || !bu) return SV_INVALID_PARAMETER;
 
@@ -873,7 +873,7 @@ hash_and_add(signed_video_t *self, const bu_info_t *bu)
     if (bu->is_first_bu_part && !bu->is_last_bu_part) {
       // If this is the first part of a non-complete BU, initialize the |crypto_handle| to
       // enable sequentially updating the hash with more parts.
-      SV_THROW(openssl_init_hash(self->crypto_handle, false));
+      SV_THROW(sv_openssl_init_hash(self->crypto_handle, false));
     }
     // Select hash function, hash the BU and store as 'latest hash'
     hash_wrapper_t hash_wrapper = get_hash_wrapper(self, bu);
@@ -980,7 +980,7 @@ signed_video_create(SignedVideoCodec codec)
     self->codec = codec;
 
     // Setup crypto handle.
-    self->crypto_handle = openssl_create_handle();
+    self->crypto_handle = sv_openssl_create_handle();
     SV_THROW_IF(!self->crypto_handle, SV_EXTERNAL_ERROR);
 
     self->gop_info = gop_info_create();
@@ -1001,7 +1001,7 @@ signed_video_create(SignedVideoCodec codec)
     self->sei_epb = codec != SV_CODEC_AV1;
     self->signing_started = false;
     self->sign_data = sign_or_verify_data_create();
-    self->sign_data->hash_size = openssl_get_hash_size(self->crypto_handle);
+    self->sign_data->hash_size = sv_openssl_get_hash_size(self->crypto_handle);
     // Make sure the hash size matches the default hash size.
     SV_THROW_IF(self->sign_data->hash_size != DEFAULT_HASH_SIZE, SV_EXTERNAL_ERROR);
 
@@ -1025,7 +1025,7 @@ signed_video_create(SignedVideoCodec codec)
     self->has_public_key = false;
 
     self->verify_data = sign_or_verify_data_create();
-    self->verify_data->hash_size = openssl_get_hash_size(self->crypto_handle);
+    self->verify_data->hash_size = sv_openssl_get_hash_size(self->crypto_handle);
   SV_CATCH()
   {
     signed_video_free(self);
@@ -1058,15 +1058,15 @@ signed_video_reset(signed_video_t *self)
     gop_info_reset(self->gop_info);
 
     validation_flags_init(&(self->validation_flags));
-    latest_validation_init(self->latest_validation);
-    accumulated_validation_init(self->accumulated_validation);
+    sv_latest_validation_init(self->latest_validation);
+    sv_accumulated_validation_init(self->accumulated_validation);
     // Empty the |bu_list|.
     bu_list_free_items(self->bu_list);
 
     memset(self->gop_info->linked_hashes, 0, sizeof(self->gop_info->linked_hashes));
     memset(self->last_bu, 0, sizeof(bu_info_t));
     self->last_bu->is_last_bu_part = true;
-    SV_THROW(openssl_init_hash(self->crypto_handle, false));
+    SV_THROW(sv_openssl_init_hash(self->crypto_handle, false));
 
     self->gop_info->num_in_partial_gop = 0;
   SV_CATCH()
@@ -1085,8 +1085,6 @@ signed_video_free(signed_video_t *self)
   onvif_media_signing_free(self->onvif);
   // Free the legacy validation if present.
   legacy_sv_free(self->legacy_sv);
-  // Free the onvif object if present.
-  onvif_media_signing_free(self->onvif);
 
   // Teardown the plugin before closing.
   sv_signing_plugin_session_teardown(self->plugin_handle);
@@ -1095,7 +1093,7 @@ signed_video_free(signed_video_t *self)
   sv_vendor_axis_communications_teardown(self->vendor_handle);
 #endif
   // Teardown the crypto handle.
-  openssl_free_handle(self->crypto_handle);
+  sv_openssl_free_handle(self->crypto_handle);
 
   // Free any pending SEIs
   free_sei_data_buffer(self->sei_data_buffer);
@@ -1131,7 +1129,7 @@ signed_video_compare_versions(const char *version1, const char *version2)
 
   int result = 0;
   int j = 0;
-  while (result == 0 && j < SV_VERSION_BYTES) {
+  while (result == 0 && j < SV_VERSION_BYTES - 1) {  // Skip patch number
     result = arr1[j] - arr2[j];
     j++;
   }
@@ -1174,7 +1172,7 @@ signed_video_parse_sei(uint8_t *bu, size_t bu_size, SignedVideoCodec codec)
     }
     printf("\n");
     signed_video_t *self = signed_video_create(codec);
-    tlv_decode(self, bu_info.tlv_data, bu_info.tlv_size);
+    sv_tlv_decode(self, bu_info.tlv_data, bu_info.tlv_size);
     signed_video_free(self);
   }
 
--- a/lib/src/sv_common.c
+++ b/lib/src/sv_common.c
@@ -563,6 +563,13 @@ parse_bu_info(const uint8_t *bu_data,
   return bu;
 }
 
+char *
+get_private_key_from_sv(signed_video_t *self)
+{
+  if (!self) return NULL;
+  return get_private_key_from_sign_data(self->sign_data);
+}
+
 /**
  * @brief Copy a Bitstream Unit Information struct (bu_info_t)
  *
--- a/lib/src/sv_common.c
+++ b/lib/src/sv_common.c
@@ -933,6 +933,33 @@ hash_and_add_for_auth(signed_video_t *self, bu_list_item_t *item)
   return status;
 }
 
+/*
+ * This function initializes ONVIF settings by porting all the settings,
+ * retrieving the private key and certificate chain from the signed video object,
+ * and setting the ONVIF signing key pair.
+ */
+svrc_t
+initialize_onvif(signed_video_t *self)
+{
+  // Convert port settings to ONVIF settings
+  svrc_t status = port_settings_to_onvif(self);
+  if (status != SV_OK) return SV_NOT_SUPPORTED;  // Cleanup on failure
+
+  // Retrieve the private key
+  const char *private_key = get_private_key_from_sv(self);
+  assert(private_key);
+  size_t private_key_size = strlen(private_key);
+  // Retrieve the certificate chain
+  const char *certificate_chain = get_axis_communications_certificate_chain(self->vendor_handle);
+  assert(certificate_chain);
+  size_t certificate_chain_size = strlen(certificate_chain);
+  // Set the signing key pair for ONVIF media signing
+  status = msrc_to_svrc(onvif_media_signing_set_signing_key_pair(
+      self->onvif, private_key, private_key_size, certificate_chain, certificate_chain_size, true));
+
+  return status;
+}
+
 /* Public signed_video_common.h APIs */
 signed_video_t *
 signed_video_create(SignedVideoCodec codec)
--- a/lib/src/sv_common.c
+++ b/lib/src/sv_common.c
@@ -28,6 +28,18 @@
 #include <stdlib.h>  // free, calloc, malloc
 #include <string.h>  // size_t
 
+// Include ONVIF Media Signing
+#if defined(NO_ONVIF_MEDIA_SIGNING)
+#include "sv_onvif.h"  // Stubs for ONVIF APIs and structs
+#elif defined(ONVIF_MEDIA_SIGNING_INSTALLED)
+// ONVIF Media Signing is installed separately; Camera
+#include <media-signing-framework/onvif_media_signing_common.h>
+#include <media-signing-framework/onvif_media_signing_signer.h>
+#else
+// ONVIF Media Signing is dragged in as a submodule; FilePlayer
+#include "includes/onvif_media_signing_common.h"
+#include "includes/onvif_media_signing_signer.h"
+#endif
 #ifdef SV_VENDOR_AXIS_COMMUNICATIONS
 #include "axis-communications/sv_vendor_axis_communications_internal.h"
 #endif
@@ -563,13 +575,6 @@ parse_bu_info(const uint8_t *bu_data,
   return bu;
 }
 
-char *
-get_private_key_from_sv(signed_video_t *self)
-{
-  if (!self) return NULL;
-  return get_private_key_from_sign_data(self->sign_data);
-}
-
 /**
  * @brief Copy a Bitstream Unit Information struct (bu_info_t)
  *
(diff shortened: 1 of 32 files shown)


lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -817,6 +818,23 @@ get_axis_communications_supplemental_authenticity(void *handle,
 
   return status;
 }
+/**
+ * Retrieves the certificate chain from the vendor handle. */
+const char *
+get_certificate_chain(signed_video_t *sv)
```
bjornvolcker, line 824: The only member in `sv` that is used is `vendor_handle`. Then I think the input should be the handle. See for example `sv_vendor_axis_communications_teardown()`

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -872,7 +890,17 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
       sv->onvif = onvif_media_signing_create(sv_codec_to_ms_codec(sv->codec));
     }
   }
-
+  if (sv->sign_data) {
+    if (sv->onvif) {
```
bjornvolcker, line 894: It looks like these two if-statements can be combined into one

lib/vendors/axis-communications/sv_vendor_axis_communications_internal.h
```diff
@@ -150,4 +150,13 @@ get_axis_communications_supplemental_authenticity(void *handle,
...
 get_axis_communications_trusted_certificate(void);
 
+/**
+ * @brief Retrieves the certificate chain from the vendor handle.
+ *
+ * @param sv The signed_video_t structure containing vendor_handle.
+ * @return The certificate chain as a null-terminated string, or NULL on failure.
+ */
+const char *
+get_certificate_chain(signed_video_t *sv);
```
bjornvolcker, line 160: The name of this function should be `get_axis_communications_certificate_chain` to follow the name convention of the other function names.

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -872,7 +890,17 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
...
+      // Set the signing key pair if sign_data is not available
```
bjornvolcker, line 895: I think you should explain why it is possible to set the signing key pair if `sign_data` is available.
Also, there is a typo in the comment. I think the "not" should be removed.

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -872,7 +890,17 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
...
+      char *private_key = get_private_key_from_sv(sv);
```
bjornvolcker, line 896: The `private_key` might be NULL. Is it fine to proceed if that is the case or should you handle it defferently?

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -872,7 +890,17 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
...
+      size_t private_key_size = strlen(private_key);
+      MediaSigningReturnCode onvif_status = onvif_media_signing_set_signing_key_pair(sv->onvif,
```
bjornvolcker, line 898: I suggest that you avoid `MediaSigningReturnCode` and use only `SignedVideoReturnCode`/`svrc_t`

Also, what if ONVIF returns an error. Is it fine to return that error or are there other options?

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -872,7 +890,17 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
...
+          private_key, private_key_size, certificate_chain, certificate_chain_size, true);
+      free(private_key);
+      return msrc_to_svrc(onvif_status);
```
bjornvolcker, line 901: You need to transfer all Signed Video settings before returning.

lib/src/sv_sign.c
```diff
@@ -860,6 +862,16 @@ SignedVideoReturnCode
 signed_video_set_private_key(signed_video_t *self, const char *private_key, size_t private_key_size)
 {
   if (!self || !private_key || private_key_size == 0) return SV_INVALID_PARAMETER;
+#if defined(SIGNED_VIDEO_DEBUG)
+#endif
+  // If ONVIF is available, call its function and map the return code
+  if (self->onvif) {
+    const char *certificate_chain = get_certificate_chain(self);
```
bjornvolcker, line 869: Since this is inside the if, `certificate_chain` cannot be NULL, but it is probably a good idea to check that with an assert.

lib/src/sv_sign.c
```diff
@@ -860,6 +862,16 @@ SignedVideoReturnCode
...
+    size_t certificate_chain_size = strlen(certificate_chain);
+    MediaSigningReturnCode onvif_status = onvif_media_signing_set_signing_key_pair(self->onvif,
+        private_key, private_key_size, certificate_chain, certificate_chain_size, true);
+    return msrc_to_svrc(onvif_status);
```
bjornvolcker, line 873: You need to transfer the Signed Video settings before returning.

lib/src/sv_sign.c
```diff
@@ -860,6 +862,16 @@ SignedVideoReturnCode
 signed_video_set_private_key(signed_video_t *self, const char *private_key, size_t private_key_size)
 {
   if (!self || !private_key || private_key_size == 0) return SV_INVALID_PARAMETER;
+#if defined(SIGNED_VIDEO_DEBUG)
+#endif
+  // If ONVIF is available, call its function and map the return code
+  if (self->onvif) {
```
bjornvolcker, line 868: Note that the same action inside this if-statement is also done in `sv_vendor_axis_communications_set_attestation_report(...)`. So, isn't it possible to create a function that does this and call it from both places?

tests/check/check_signed_video_sign.c
```diff
@@ -498,6 +498,33 @@ START_TEST(correct_signed_stream_with_eos)
 END_TEST
 #endif
 
+START_TEST(check_the_priv_key)
```
bjornvolcker, line 501: It is a great idea to have this kind of test since it verifies that we can reconstruct the private key and set it in ONVIF Media Signing. But, the placement of the test in the code is off. Please, put it in the same order as it is called, that is, as a last test.

tests/check/check_signed_video_sign.c
```diff
@@ -498,6 +498,33 @@ START_TEST(correct_signed_stream_with_eos)
...
+{
+  SignedVideoCodec codec = settings[_i].codec;
+
```
bjornvolcker, line 504: Add comments in the test so it becomes easier what is done.

tests/check/check_signed_video_sign.c
```diff
@@ -498,6 +498,33 @@ START_TEST(correct_signed_stream_with_eos)
...
+  signed_video_t *sv = signed_video_create(codec);
+  ck_assert(sv);
+  char *private_key = NULL;
+  size_t private_key_size = 0;
+
+  ck_assert(read_test_private_key(settings[_i].ec_key, &private_key, &private_key_size, false));
+  signed_video_set_private_key(sv, private_key, private_key_size);
+  const char *private_key_2 = get_private_key_from_sv(sv);
```
bjornvolcker, line 512: Rename to `reconstructed_private_key` so it becomes more clear what it is.

tests/check/check_signed_video_sign.c
```diff
@@ -498,6 +498,33 @@ START_TEST(correct_signed_stream_with_eos)
...
+  private_key_size = strlen(private_key_2);
```
bjornvolcker, line 513: Can we expect the size to be unchanged? If so, I would instead check that with
`ck_assert_int_eq(strlen(reconstructed_private_key), private_key_size);`

tests/check/check_signed_video_sign.c
```diff
@@ -498,6 +498,33 @@ START_TEST(correct_signed_stream_with_eos)
...
+{
+  SignedVideoCodec codec = settings[_i].codec;
+
+  signed_video_t *sv = signed_video_create(codec);
+  ck_assert(sv);
+  char *private_key = NULL;
+  size_t private_key_size = 0;
+
+  ck_assert(read_test_private_key(settings[_i].ec_key, &private_key, &private_key_size, false));
+  signed_video_set_private_key(sv, private_key, private_key_size);
```
bjornvolcker, line 511: I think you should free `private_key` here. Then we can detect if the library stores a copy and pass it on, which it should not do.

tests/check/check_signed_video_sign.c
```diff
@@ -942,6 +969,7 @@ signed_video_suite(void)
   tcase_add_loop_test(tc, limited_sei_payload_size, s, e);
   tcase_add_loop_test(tc, signing_partial_gops, s, e);
   tcase_add_loop_test(tc, signing_mulitslice_stream_partial_gops, s, e);
+  tcase_add_loop_test(tc, check_the_priv_key, s, e);
```
bjornvolcker, line 972: Can you rename the test to something more descriptive?

bjornvolcker (changes requested):
Some initial comments

tests/check/check_signed_video_sign.c
```diff
@@ -934,6 +907,49 @@ START_TEST(signing_mulitslice_stream_partial_gops)
...
+  // Create a signed video
+  signed_video_t *sv = signed_video_create(codec);
+  ck_assert(sv);
+
+  char *private_key = NULL;
+  size_t private_key_size = 0;
+
+  // Read and set the private key
+  ck_assert(read_test_private_key(settings[_i].ec_key, &private_key, &private_key_size, false));
+  signed_video_set_private_key(sv, private_key, private_key_size);
```
bjornvolcker, line 929: Is it possible to reconstruct the private key without create->set->free of signed_video_t?

Also, and maybe more important, have you verified that it works on a device? If it does not you have a lot of dead code we need to rewrite.

bjornvolcker (changes requested):
There are still a bunch of unanswered review questions.

Also, have you verified that it works to use the reconstructed key on a device?

lib/src/sv_sign.c
```diff
@@ -869,10 +871,13 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
     assert(certificate_chain);
     size_t certificate_chain_size = strlen(certificate_chain);
     status = port_settings_to_onvif(self);
-    if (status != SV_OK) return status;
-
-    return msrc_to_svrc(onvif_media_signing_set_signing_key_pair(self->onvif, private_key,
-        private_key_size, certificate_chain, certificate_chain_size, true));
+    if (status == SV_OK) {
+      return msrc_to_svrc(onvif_media_signing_set_signing_key_pair(self->onvif, private_key,
```
bjornvolcker, line 875: What if this call fails, shouldn't you free `self->onvif` then as well? Otherwise you get different behaviors for different type of failures.

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -888,17 +885,17 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
...
+    if (status != SV_OK) goto catch_error;  // Cleanup on failure
     char *private_key = get_private_key_from_sv(sv);
-    if (!private_key) goto catch_error;
+    if (!private_key) goto catch_error;  // Ensure cleanup
     size_t private_key_size = strlen(private_key);
-    SignedVideoReturnCode status = msrc_to_svrc(onvif_media_signing_set_signing_key_pair(
+    status = msrc_to_svrc(onvif_media_signing_set_signing_key_pair(
         sv->onvif, private_key, private_key_size, certificate_chain, certificate_chain_size, true));
     free(private_key);
+    if (status != SV_OK) goto catch_error;  // Cleanup on failure
```
bjornvolcker, line 898: Now all failures within this if-statement behaves the same, but it behaves differently from the failures in the function `sv_sign.c`. You need to have a unified behavior for ONVIF Media Signing.

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -888,17 +885,17 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
       sv->onvif = onvif_media_signing_create(sv_codec_to_ms_codec(sv->codec));
     }
   }
+  // Set the signing key pair if sign_data is available
```
bjornvolcker, line 888: I still think this comment is unclear. If you ask any other person in our team. Do you think it can be easily understood? Can you describe it in common words rather in terms of variables, that is, without `sign_data`?

lib/vendors/axis-communications/sv_vendor_axis_communications_internal.h
```diff
@@ -153,7 +153,7 @@ get_axis_communications_trusted_certificate(void);
 /**
  * @brief Retrieves the certificate chain from the vendor handle.
  *
- * @param sv The signed_video_t structure containing vendor_handle.
+ * @param handle The handle containing certificate chain.
```
bjornvolcker, line 156: I think you should use a description like on line 133. By design, the user already knows that the certificate chain is "inside" the handle, but what handle should be passed in. That is the kind of information to provide the user.

tests/check/check_signed_video_sign.c
```diff
@@ -907,6 +907,49 @@ START_TEST(signing_mulitslice_stream_partial_gops)
...
+  free(private_key);
+  // Retrieve the private key from the sv
+  const char *reconstructed_private_key = get_private_key_from_sv(sv);
+  // Verify the key size matches the original
+  ck_assert_int_eq(strlen(reconstructed_private_key), private_key_size);
+  signed_video_free(sv);
+
+  // Create a new signed video object  and reuse the reconstructed private key
```
bjornvolcker, line 937: nit: There is a double white space after "object"

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -29,7 +29,7 @@
 #include <stdbool.h>
 #include <stdlib.h>  // malloc, memcpy, calloc, free
 
-#include "sv_internal.h"  // signed_video_t
+#include "sv_internal.h"  // signed_video_t, onvif_media_signing_set_signing_key_pair()
```
bjornvolcker, line 32: AFAIK, `onvif_media_signing_set_signing_key_pair()` is not declared in `sv_internal.h`. Include the right header file.

Also, you use the function `port_settings_to_onvif()`. From where is that included?

bjornvolcker (changes requested):
Still a few old review comments to address. And a few new ones.

lib/src/sv_common.c
```diff
@@ -933,6 +933,33 @@ hash_and_add_for_auth(signed_video_t *self, bu_list_item_t *item)
...
+/*
+ * This function initializes ONVIF settings by porting all the settings,
+ * retrieving the private key and certificate chain from the signed video object,
+ * and setting the ONVIF signing key pair.
+ */
+svrc_t
+initialize_onvif(signed_video_t *self)
+{
+  // Convert port settings to ONVIF settings
+  svrc_t status = port_settings_to_onvif(self);
```
bjornvolcker, line 945: May I suggest to use the SV_TRY, SV_CATCH macros. Then this function would become something like below.

You also leaking memory since you do not free the key.
I also thought that the function should clean up the ONVIF session upon failure and if then by design falling back to normal Signed Video.

```
char *private_key = NULL;
const char *certificate_chain = NULL;
SV_TRY()
  SV_THROW(port_settings_to_onvif(self));
  private_key = get_private_key_from_sv(self);
  SV_THROW_IF(!private_key, SV_MEMORY);
  certificate_chain = get_axis_communications_certificate_chain(self->vendor_handle);
  SV_THROW_IF(!certificate_chain, SV_MEMORY);
  SV_THROW(msrc_to_svrc(onvif_media_signing_set_signing_key_pair(self->onvif, private_key, private_key_size, certificate_chain, certificate_chain_size, false)));
SV_CATCH()
{
  onvif_media_signing_free(self->onvif);
  self->onvif = NULL;
}
SV_DONE(status)

free(private_key);

return status;
```

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -29,7 +29,7 @@
 #include <stdbool.h>
 #include <stdlib.h>  // malloc, memcpy, calloc, free
 
-#include "sv_internal.h"  // signed_video_t
+#include "sv_internal.h"  // signed_video_t, port_settings_to_onvif()
```
bjornvolcker, line 32: You are not using `port_settings_to_onvif()` anywhere in this file. Remove.

lib/src/sv_sign.c
```diff
@@ -883,10 +886,10 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
     SV_THROW_IF(!self->plugin_handle, SV_EXTERNAL_ERROR);
   SV_CATCH()
   SV_DONE(status)
-
-  // Free the EVP_PKEY since it is no longer needed. It is handled by the signing plugin.
-  sv_openssl_free_key(self->sign_data->key);
-  self->sign_data->key = NULL;
+  // If ONVIF is available, call initialize Onvif.
+  if (self->onvif) {
+    status = initialize_onvif(self);
```
bjornvolcker, line 891: This is not correct. If `status` is an error before this line and `initialize_onvif` succeeds, you will replace `status` with `SV_OK` and return it when it actually failed further up.

tests/check/check_signed_video_sign.c
```diff
@@ -912,6 +912,47 @@ START_TEST(signing_mulitslice_stream_partial_gops)
...
+{
+  SignedVideoCodec codec = settings[_i].codec;
+  SignedVideoReturnCode sv_rc;
+  // Create a sign_data
+  sign_or_verify_data_t sign_data = {0};
+
+  char *private_key = NULL;
+  size_t private_key_size = 0;
+
+  // Read and set the private key
```
bjornvolcker, line 931: grammar: remove "and set"

tests/check/check_signed_video_sign.c
```diff
@@ -912,6 +912,47 @@ START_TEST(signing_mulitslice_stream_partial_gops)
...
+  ck_assert(read_test_private_key(settings[_i].ec_key, &private_key, &private_key_size, false));
+  sv_rc = openssl_private_key_malloc(&sign_data, private_key, private_key_size);
+  ck_assert_int_eq(sv_rc, SV_OK);
+  free(private_key);
+  // Retrieve the private key from the sign_data
+  const char *reconstructed_private_key = get_private_key_from_sign_data(&sign_data);
+  // Verify the key size matches the original
+  ck_assert_int_eq(strlen(reconstructed_private_key), private_key_size);
+
+  // Create a new signed video object and reuse the reconstructed private key
```
bjornvolcker, line 941: grammar: remove "new" and "reuse" -> "set"

lib/vendors/axis-communications/sv_vendor_axis_communications_internal.h
```diff
@@ -150,4 +150,13 @@ get_axis_communications_supplemental_authenticity(void *handle,
 const char *
 get_axis_communications_trusted_certificate(void);
 
+/**
+ * @brief Retrieves the certificate chain from the vendor handle.
+ *
```
bjornvolcker, line 155: You should mention that the ownership is NOT transferred.

lib/src/sv_openssl.c
```diff
@@ -147,6 +147,51 @@ openssl_private_key_malloc(sign_or_verify_data_t *sign_data,
   return status;
 }
 
+/**
+ * Extracts the private key from the signing data structure.
+ * This function retrieves the private key stored within the `sign_data` structure
+ * and returns it as a dynamically allocated null-terminated string in PEM format.
+ */
+char *
+get_private_key_from_sign_data(sign_or_verify_data_t *sign_data)
```
bjornvolcker, line 156: The only thing this function uses from `sign_data` is the member `key`. May I suggest that this function takes that key as input instead.

Then rename the function accordingly.

And the function name should begin with `openssl_`

And also, the declaration should be in sv_openssl_internal.h and not in sv_internal.h

lib/src/sv_openssl.c
```diff
@@ -147,6 +147,51 @@ openssl_private_key_malloc(sign_or_verify_data_t *sign_data,
...
+{
+  if (!sign_data || !sign_data->key) return NULL;
```
bjornvolcker, line 158: May I suggest that you make this function use the SV_TRY macros

lib/src/sv_internal.h
```diff
@@ -454,6 +457,21 @@ sv_codec_to_ms_codec(SignedVideoCodec sv_codec);
...
+ * This function initializes ONVIF settings by porting all the settings,
+ * retrieving the private key and certificate chain from the signed video object,
+ * and setting the ONVIF signing key pair.
+ */
+svrc_t
+initialize_onvif(signed_video_t *self);
+
+/**
+ * This function retrieves the private key stored within the `sign_data` structure
+ * and returns it as a dynamically allocated null-terminated string in PEM format.
```
bjornvolcker, line 470: You should add that the memory ownership is transferred.

lib/src/sv_common.c
```diff
@@ -563,6 +563,13 @@ parse_bu_info(const uint8_t *bu_data,
   return bu;
 }
 
+char *
+get_private_key_from_sv(signed_video_t *self)
```
bjornvolcker, line 567: This function is only a wrapper of `get_private_key_from_sign_data()` so I think you should use that one directly in all the places.

lib/src/sv_common.c
```diff
@@ -926,6 +933,33 @@ hash_and_add_for_auth(signed_video_t *self, bu_list_item_t *item)
...
+  const char *private_key = get_private_key_from_sv(self);
+  assert(private_key);
+  size_t private_key_size = strlen(private_key);
+  // Retrieve the certificate chain
+  const char *certificate_chain = get_axis_communications_certificate_chain(self->vendor_handle);
+  assert(certificate_chain);
+  size_t certificate_chain_size = strlen(certificate_chain);
+  // Set the signing key pair for ONVIF media signing
+  status = msrc_to_svrc(onvif_media_signing_set_signing_key_pair(
+      self->onvif, private_key, private_key_size, certificate_chain, certificate_chain_size, true));
```
bjornvolcker, line 958: You need to change `true` to `false` because the key is not user provisioned. It is factory provisioned.

bjornvolcker (changes requested):

bjornvolcker:
This PR is quite large. Is it possible to split it into a series of 5-10 separate PRs. It will make reviewing much easier.

lusikamalo2 (commented):
No extra comments from my side so far.

lib/src/sv_sign.c
```diff
@@ -887,7 +925,7 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
   SV_CATCH()
   SV_DONE(status)
   // If ONVIF is available, call initialize Onvif.
-  if (self->onvif) {
+  if (self->onvif && status == SV_OK) {
```
bjornvolcker, line 928: Isn't it better to put this inside the TRY/CATCH if you anyhow need to check `status`?

lib/src/sv_sign.c
```diff
@@ -771,6 +771,44 @@ signed_video_get_sei(signed_video_t *self,
...
+    // Convert port settings to ONVIF settings
+    SV_THROW(port_settings_to_onvif(self));
+    // Retrieve the private key
+    private_key = openssl_extract_private_key(self->sign_data);
+    SV_THROW_IF(!private_key, SV_MEMORY);
+    size_t private_key_size = strlen(private_key);
+    // Retrieve the certificate chain
+    certificate_chain = get_axis_communications_certificate_chain(self->vendor_handle);
+    SV_THROW_IF(!certificate_chain, SV_MEMORY);
+    size_t certificate_chain_size = strlen(certificate_chain);
```
bjornvolcker, line 795: I think it is fine to use `strlen(...)` directly in the call below.

lib/src/sv_openssl.c
```diff
@@ -148,48 +148,61 @@ openssl_private_key_malloc(sign_or_verify_data_t *sign_data,
...
+  // Free BIO and return success
   BIO_free(bio);
   return private_key;
+
+cleanup:
+  // Free allocated resources on failure
+  if (bio) {
+    BIO_free(bio);
+  }
+  if (private_key) {
```
bjornvolcker, line 202: No need to put check for NULL. `free()` can handle NULL pointers.

lib/src/sv_openssl.c
```diff
@@ -148,48 +148,61 @@ openssl_private_key_malloc(sign_or_verify_data_t *sign_data,
...
+  private_key = (char *)malloc(buf_mem->length + 1);
   if (!private_key) {
-    BIO_free(bio);
-    return NULL;
+    goto cleanup;
   }
-
+  // Copy the private key and null-terminate it
   memcpy(private_key, buf_mem->data, buf_mem->length);
   private_key[buf_mem->length] = '\0';
```
bjornvolcker, line 191: If you use `calloc` instead of `malloc` you do not need to explicitly set the null-terminating character.

lib/src/sv_common.c
```diff
@@ -28,6 +28,18 @@
...
+#include "sv_onvif.h"  // Stubs for ONVIF APIs and structs
+#elif defined(ONVIF_MEDIA_SIGNING_INSTALLED)
+// ONVIF Media Signing is installed separately; Camera
+#include <media-signing-framework/onvif_media_signing_common.h>
+#include <media-signing-framework/onvif_media_signing_signer.h>
+#else
+// ONVIF Media Signing is dragged in as a submodule; FilePlayer
+#include "includes/onvif_media_signing_common.h"
+#include "includes/onvif_media_signing_signer.h"
+#endif
```
bjornvolcker, line 42: I cannot see where in `sv_common.c` these onvif includes are needed.

bjornvolcker (changes requested):

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -886,7 +886,7 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
     }
   }
   // Initialize Onvif if private key is set before.
-  if (sv->sign_data && sv->onvif) {
+  if (sv->sign_data->key && sv->onvif) {
```
bjornvolcker, line 889: You need to check `sv->sign_data` for not being a NULL pointer as well, otherwise you get a segmentation fault here.

bjornvolcker (changes requested):

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -886,12 +886,11 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
     }
   }
   // Initialize Onvif if private key is set before.
-  if (sv->sign_data->key && sv->onvif) {
+  if (sv->sign_data && sv->onvif) {
```
bjornvolcker, line 889: But now you've removed the `key` check. You need both

lib/src/sv_sign.c
```diff
@@ -918,12 +924,11 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
...
     SV_THROW(openssl_read_pubkey_from_private_key(self->sign_data, &self->pem_public_key));
 
-    self->plugin_handle = sv_signing_plugin_session_setup(private_key, private_key_size);
-    SV_THROW_IF(!self->plugin_handle, SV_EXTERNAL_ERROR);
-    // If ONVIF is available, call initialize Onvif.
-    if (self->onvif) {
-      SV_THROW(initialize_onvif(self));
-    }
+    // Fallback to signing plugin if ONVIF is unavailable or ONVIF cant be initialized.
+    if(!self->onvif || !(initialize_onvif(self) == SV_OK)) {
```
bjornvolcker, line 928: This is a really awkward check and construction. Have you tested this for both orders of "set_private_key" and "set_attestation"?
Have you checked how many signing plugins run when signing has started?

lib/src/sv_internal.h
```diff
@@ -457,20 +457,12 @@ port_settings_to_onvif(signed_video_t *self);
 /*
  * This function initializes ONVIF settings by porting all the settings,
  * retrieving the private key and certificate chain from the signed video object,
- * and setting the ONVIF signing key pair.
+ * and setting the ONVIF signing key pair. The function will free the ONVIF object
+ * on failure, allowing the framework to continue with the signed video.
```
bjornvolcker, line 461: Yes, this is the desired behavior, but is that really what happens with your implementation?
Have you tested/verified that?

bjornvolcker (changes requested):
It still doesn't work. How are you testing the various cases?

- order of operations
- number of signing plugins
- fallback to signed video upon failure

lib/src/sv_sign.c
```diff
@@ -881,13 +920,13 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
 
     self->plugin_handle = sv_signing_plugin_session_setup(private_key, private_key_size);
```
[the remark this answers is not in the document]
bjornvolcker: ONVIF signing will still be used through this plugin, hence should be kept. I noticed that after adding this comment.

lib/src/sv_sign.c
```diff
@@ -923,12 +918,15 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
...
     SV_THROW(openssl_private_key_malloc(self->sign_data, private_key, private_key_size));
     SV_THROW(openssl_read_pubkey_from_private_key(self->sign_data, &self->pem_public_key));
+    self->private_key = private_key;
 
-    // Fallback to signing plugin if ONVIF is unavailable or ONVIF cant be initialized.
-    if(!self->onvif || !(initialize_onvif(self) == SV_OK)) {
+    // Try to initialize ONVIF; if it is not available or cannot be initialized,
+    // fallback to Signed Video signing plugin.
+    initialize_onvif(self);
+    if (!self->onvif) {
```
bjornvolcker, line 926: Now when ONVIF signing plugin is called through the Axis signing plugin (see other PR) you need to remove the if-statement

lib/src/sv_sign.c
```diff
@@ -808,9 +804,8 @@ initialize_onvif(signed_video_t *self)
   }
   SV_DONE(status);
   // Free allocated memory
-  free(private_key);
-
-  return status;
+  self->private_key = NULL;
+  if (status != SV_OK && self->onvif != NULL) assert(true);
```
bjornvolcker, line 808: Well, you should also check the opposite. I only gave you an example. And also shouldn't it be false in assert?
```
assert((status == SV_OK && !self->onvif) || (status != SV_OK && self->onvif));
```

lib/src/sv_sign.c
```diff
@@ -775,31 +775,27 @@ signed_video_get_sei(signed_video_t *self,
...
-svrc_t
+void
 initialize_onvif(signed_video_t *self)
 {
-  char *private_key = NULL;
   const char *certificate_chain = NULL;
   // Sanity check ONVIF object.
-  if(!self->onvif) {
-    return SV_INVALID_PARAMETER;
+  if (!self->onvif) {
```
bjornvolcker, line 783: You should also check if `private_key` exists

lib/src/sv_sign.c
```diff
@@ -923,12 +918,15 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
     // Temporally turn the PEM |private_key| into an EVP_PKEY and allocate memory for signatures.
     SV_THROW(openssl_private_key_malloc(self->sign_data, private_key, private_key_size));
     SV_THROW(openssl_read_pubkey_from_private_key(self->sign_data, &self->pem_public_key));
+    self->private_key = private_key;
```
bjornvolcker, line 921: You forgot to set `private_key_size`

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -886,11 +886,8 @@ sv_vendor_axis_communications_set_attestation_report(signed_video_t *sv,
     }
   }
   // Initialize Onvif if private key is set before.
-  if (sv->sign_data && sv->onvif) {
-    SignedVideoReturnCode status = initialize_onvif(sv);
-    if (status == SV_OK) {
-      return status;
-    }
+  if (sv->sign_data && sv->sign_data->key && sv->onvif) {
```
bjornvolcker, line 889: No need for if-statement. The check should be inside `initilaize_onvif()`

lib/src/sv_internal.h
```diff
@@ -304,6 +304,8 @@ struct _signed_video_t {
   // For signing plugin
   void *plugin_handle;
   sign_or_verify_data_t *sign_data;  // Pointer to all necessary information to sign in a plugin.
+  const char *private_key;
```
bjornvolcker, line 307: Add comment what this is used for. A reader might assume that this is the signing key used during the session, but it is not.

lib/src/sv_sign.c
```diff
@@ -878,16 +918,18 @@ signed_video_set_private_key(signed_video_t *self, const char *private_key, size
     // Temporally turn the PEM |private_key| into an EVP_PKEY and allocate memory for signatures.
     SV_THROW(openssl_private_key_malloc(self->sign_data, private_key, private_key_size));
     SV_THROW(openssl_read_pubkey_from_private_key(self->sign_data, &self->pem_public_key));
-
-    self->plugin_handle = sv_signing_plugin_session_setup(private_key, private_key_size);
-    SV_THROW_IF(!self->plugin_handle, SV_EXTERNAL_ERROR);
+    self->private_key = private_key;
+
+    // Try to initialize ONVIF; if it is not available or cannot be initialized,
+    // fallback to Signed Video signing plugin.
```
bjornvolcker, line 924: The comment about "fallback" should be inside `initialize_onvif` since you do not take any actions regarding that out here.

lib/vendors/axis-communications/sv_vendor_axis_communications.c
```diff
@@ -818,6 +818,19 @@ get_axis_communications_supplemental_authenticity(void *handle,
...
+/**
+ * Retrieves the certificate chain from the vendor handle. */
+const char *
+get_axis_communications_certificate_chain(void *handle)
+{
+  if (!handle) return NULL;
+
+  sv_vendor_axis_communications_t *self = (sv_vendor_axis_communications_t *)handle;
+
+  // Return the certificate chain
```
bjornvolcker, line 830: Redundant comment. I think the reader is clever enough to understand that from the line below.

bjornvolcker (changes requested):

PR #404, opened 2025-03-03
Fix compile error if warning treated as error

bjornvolcker:
This solves the warning -Werror=stringop-truncation.
The structures are identical in length, so it is fine to simply
copy all the content even though there are null-terminated
characters.


lusikamalo2 (approved):
OK.

PR #405, opened 2025-03-03
Includes ONVIF Media Signing if present

bjornvolcker:
The meson build structure has been updated to support ONVIF
Media Signing, either through a submodule or if installed
separately.
Includes have been updated and tests can now run if ONVIF
Media Signing code is present.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #406, opened 2025-03-04
Changes behavior when comparing versions

bjornvolcker:
There is a check for version mismatch between signing and
validation. If the stream has been signed with a newer version
than what is used when validating the validation result cannot
be guaranteed. The reason is that it is not possible to be
forward compatible.
However, if the change is minor, like when the patch number
only has changed, it should still be possible to trust the
result.

This commit checks only major and minor for version mismatch.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #407, opened 2025-03-04
Fixes tests to pass if Media Signing code is present

bjornvolcker:
Removes an assert, since hash algo will never be known if the
stream was signed with ONVIF Media Signing.

Removes a double free of the ONVIF Media Signing session.

Updates a validation test w.r.t. presence of ONVIF Media Signing
code.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #420, opened 2025-03-19
Fixes bug if lib is not built with a vendor

bjornvolcker:
If the library was built without any vendor (Axis) the
authentication side tried to create an ONVIF Media Signing
session when it should not. This commit fixes that.

In addition a minor spell change of GLib has been made.

The version has been changed to v2.1.1.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #423, opened 2025-03-28
Instead of setting the same timestamp all the time

bjornvolcker:
this commit increments the timestamp for primary slices.
Tests now verify that the end_timestamp is larger than the
start_timestamp.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #424, opened 2025-03-31
Merges generate_sei and add_payload_to_buffer

bjornvolcker:
These two functions are always called right after each other.
They have been combined into one function
generate_sei_and_add_to_buffer(...).

Also, the variable 'payload' has been changed to 'sei' in this
function since it actually generates a complete SEI and not
just the payload of the SEI.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #427, opened 2025-04-01
Allows for unsigned SEIs

bjornvolcker:
To be able to support signing multiple GOPs the library
has to manage unsigned SEIs. This means that the queue
of SEIs can have some already completed SEIs as well as
SEIs waiting for the signature to be completed.

This commit changes the behavior slightly when getting
SEIs.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #429, opened 2025-04-01
Implements support for signing multiple GOPs

bjornvolcker:
Through the API signed_video_set_signing_frequency(...) the user
can specify how often GOPs should be signed. Unsigned GOPs will
still produce SEIs, but without signatures. These SEIs are
hashed like normal frames and eventually signed with the final
SEI.

A test has been added for signing multiple GOPs.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #430, opened 2025-04-02
Adds tests for signing multiple GOPs

bjornvolcker:
Currently all tests are disabled since validation has not yet
been implemented and the code ends up in a dead lock.

A new test function to pop complete GOPs instead of counting
items has been added.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #435, opened 2025-04-04
Decodes SEIs at registration

bjornvolcker:
This commit change the behavior to decode the optional tags in
a SEI when being registered. After that, the signature is
verified.

Further, to avoid the risk of reading the same Public key twice
the Public key is not updated if it has changed. This is a
security precaution since change of key should never be allowed.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #438, opened 2025-04-04
Get next SEI only once

bjornvolcker:
The next available SEI is searched for in several places. It is
only necessary to do it once, when preparing for validation.

This commit passes that SEI on for further validation.


lusikamalo2 (approved):
OK.

PR #440, opened 2025-04-07
Starts using associated_sei instead of used_in_gop_hash

bjornvolcker:
When dealing with unsigned SEIs it will become important to which
SEI a frame belongs, since the final validation can only be made
once the signed SEI has been received. Therefore, a boolean is
not enough to handle signing multiple GOPs.

This commits replaces the boolean flag in associated places.
When for example removing items or adding missing items the code
still relies on the flag. These will be changed in later commits.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #445, opened 2025-04-08
Sets validation status of SEI when preparing for validation

bjornvolcker:
If no SEIs are detected as lost the status is set. If the SEI is
signed the validation_status is set at once. Otherwise
validation_status_if_sei_ok is set. This member is used when
validating a GOP that is associated with a SEI that is not signed.
The content of the SEI will be determined when the final signed
SEI is used in validation.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #447, opened 2025-04-08
Resets validation_status_if_sei_ok

bjornvolcker:
when removing sei association and/or validating without sei.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #448, opened 2025-04-08
Removes unused parts in verify_hashes_with_hash_list

bjornvolcker:
Since items are already associated with a sei there is no longer
a need to check for end of GOPs etc.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #450, opened 2025-04-09
Removes the used_for_linked_hash flag

bjornvolcker:
There is a more efficient way of detecting the BU hash that
also represents a linked hash. The first pending item should
by design be the linked hash.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #453, opened 2025-04-10
Move computing pending BUs into while loop

bjornvolcker:
The computation should be done when there is a signed SEI. If
the first validation fails due to an outdated SEI pending BUs
have to be re-computed.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #455, opened 2025-04-11
Simplifies Golden SEI validation

bjornvolcker:
Since SEIs are (partially) decoded and hashed upon registration
there is no longer a need to decode and hash Golden SEIs
separately.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #456, opened 2025-04-22
Validates intact stream signed with multiple GOPs

bjornvolcker:
The signing side has been updated to send correct number of hashed
BUs. Also, SEI association on the validation side is removed if
a lost SEI was detected, or if the first validation failed.


lusikamalo2 (approved):
OK.

PR #459, opened 2025-04-22
Makes latest_validated_gop int64_t

bjornvolcker:
This helps in handling an arbitrary amount of wraparound, which
will be implemented.

The commit also has minor changes wher a helper member gop_info
is added.


lusikamalo2 (approved):
OK.

PR #462, opened 2025-04-22
Determines wraparounds in partial GOP counter

bjornvolcker:
Some variable renaming and changes to use int64_t.


lusikamalo2 (approved):
OK.

PR #463, opened 2025-04-23
Can now handle negative num_lost_seis

bjornvolcker:
This can happen if SEIs are re-ordered, session reset on camera,
or if there are SEIs in the beginning of a session that belongs to
a previous part of the stream.

Adds a new sei_in_sync validation flag and a function to check if
a hash is empty.


lusikamalo2 (approved):
OK.

PR #464, opened 2025-04-23
Completes some struct resets

bjornvolcker:
Affects accumulated auth report, gop_info and validation flags.
The authenticity report "Signature present" is now only shown
once, which affects for example file scrubbing. Tests have been
updated accordingly.


lusikamalo2 (approved):
OK.

PR #465, opened 2025-04-23
Code cleanup of prepare_for_validation

bjornvolcker:
Now uses validation flag num_lost_seis and only detects lost SEIs
once. Also renames compute gop hash function since it now only
computes it. Also renames update link hash function.


lusikamalo2 (approved):
OK.

PR #466, opened 2025-04-23
Removes redundant code

bjornvolcker:
Two members have been removed since they are redundant. These
are has_lost_sei and partial_gop_is_synched.

This change breaks file scrubbing forward. Therefore, these tests
have been disabled until solved.


lusikamalo2 (approved):
OK.

PR #468, opened 2025-04-23
Fixes forward scrubbing

bjornvolcker:
The root cause was among the validation flags, where is_first_sei
was not updated correctly.


lusikamalo2 (approved):
OK.

PR #469, opened 2025-04-23
Refactors verify_hashes_with_sei slightly

bjornvolcker:
Moves out some of the logic that should be handled generically.


lusikamalo2 (approved):
OK.

PR #470, opened 2025-04-24
Handles num_invalid across unsigned GOPs

bjornvolcker:
and can therefore move the code of final stats prints into the
while loop.


lusikamalo2 (approved):
OK.

PR #472, opened 2025-04-24
Removes lost_start_of_gop

bjornvolcker:
It is not needed since there is a separate mechanism
to synchronize the first SEI.


lusikamalo2 (approved):
OK.

PR #473, opened 2025-04-24
Simplifies verify_hashes_with_hash_list

bjornvolcker:
Removes stats update since it is done outside this function.
Also, marking only first item if link is incorrect is also done
outside this function.


lusikamalo2 (approved):
OK.

PR #479, opened 2025-04-28
Adds a nullptr check

bjornvolcker:
This pleases Coverity. It should not be possible to become
a nullptr, but since the validation side code has become
more complex it is not a bad thing to add.

Bumps version to v2.2.1.


lusikamalo2 (approved):
OK.

PR #484, opened 2025-05-08
Make current_partial_gop int64_t

bjornvolcker:
This makes it easier to track wraparounds etc. Note that the
value is still stored as an uint32_t in the TLV.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #485, opened 2025-05-09
Adds next_partial_gop counter

bjornvolcker:
The decoded partial GOP counter is stored for later use.
Currently current_partial_gop is updated right after decode,
but later it will be updated after validation based on the
result. There could for example be missing SEIs or the SEIs
might not be synchronized.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #486, opened 2025-05-09
Removes latest_validated_gop

bjornvolcker:
When validating and decoding the procedure has now
changed to look forward rather than backward. The
member current_partial_gop now reflects the GOP
number that currently has been validated. It is
updated after a validation has been performed,
either with next_partial_gop or incremented if
a GOP has been lost.

The member latest_validated_gop is no longer
used, and therefore removed.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #487, opened 2025-05-09
Increment of current_partial_gop done outside TLV

bjornvolcker:
On the signing side, the current_partial_gop was updated
when encoding the TLVs. The TLV encoding process should
not take any actions. It should solely write the current
state into the SEI.

The update is now done outside encoding. Further, the
counter is initialized to -1 instead of 0. This makes
it more clear on the validation side when the first SEI
is to be used.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #488, opened 2025-05-09
Forces verify_without_sei to validate at least one BU

bjornvolcker:
When trying to synchronize SEIs situations when validating
without a SEI to close a gap can occur. In such a case, it
should select and validate at least one BU. Otherwise, an
endless loop can occur.

This commit affects one test,
fast_forward_stream_without_reset, which temporarily has been
changed until SEI synchronization has been separated from
first validation.


mithydolphin (approved):

mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #490, opened 2025-05-12
Removes detect_lost_seis

bjornvolcker:
When repeating validation with the same SEI lost SEIs should
not be re-detected based on the SEI. The num_lost_sei counter
should be updated separately and independent on the SEI.

Two new tests temporarily fails due to this change. The have
been disabled for now.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #493, opened 2025-05-12
Computes SEI synchronization seperately

bjornvolcker:
The sei_in_sync is only updated in one place and only if it
was not in sync before.

This solves two previous disabled tests, but fails a new one,
which has been disabled for now.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #495, opened 2025-05-12
Handles reset on camera

bjornvolcker:
A reset made on the camera will rewind the GOP counter
which on the validation side will be detected as an
old SEI. If an old SEI is detected, but the GOP hash
is correct return success.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

PR #496, opened 2025-05-12
The first_validation is updated to rely on sei_in_sync

bjornvolcker:
Special operations on the first validation is taken if the
SEI is not in sync. All validation has to be rewinded if so.
So far this decision has been made based on the validation
results. Now it is based on wether the SEI is in sync or
not.


lusikamalo2 (approved):
OK.

PR #497, opened 2025-05-12
Makes a second (final) validation stats update

bjornvolcker:
If multiple GOPs are validated and the validation started
being out of sync, but sync was established, the first GOP
should no longer be ignored. Therefore, perform a final
stats update after having looped through the GOPs.


lusikamalo2 (approved):
OK.

mithydolphin (approved):

PR #500, opened 2025-06-16
Adds sanity check on TLV length

bjornvolcker:
If the read TLV length parameter is too large, i.e., will make
reading outside memory, when finding a tag no tag is returned.
Further, when finding the signature tag a second time fails
the SEI is put in an error state with member is_valid.


mithydolphin (approved):
LGTM

PR #504, opened 2025-08-11
Test dummy data now supports FH + TG for AV1

bjornvolcker:
A test stream can now be created with FH and following TG. The
TG is represented by 't'. That is, a GOP with frame header is
written as "ItPtPtPt".


lusikamalo2 (approved):
OK.

PR #505, opened 2025-08-11
Adds new AV1 tests in a separate file

bjornvolcker:
One test has been added and that is "normal" signing with FH.
A new input parameter to test stream creation has been added.
This parameter tells the system to use OBU_FRAME or FH+TG.


lusikamalo2 (approved):
OK.

PR #507, opened 2025-08-12
Hashes FH+TGs together

bjornvolcker:
For efficiency when transmitting the hash list hashing both the
FH and all TGs becomes quite costly. Therefore, one single hash
for the entire frame (FH+TGs) is generated.
The logic is similar to hashing when a NALU is split into
several parts. There is one major difference though, and that is
that for FH+TGs it is not known when to complete the ongoing
hash until the next frame has been detected.

Therefore, a helper function sv_add_ongoing_hash() that handles
frame transitions has been added.

The commit updates both the signing and the validation side
which keeps the test intact.


mithydolphin (approved):

lusikamalo2 (approved):
OK.

